PR #17, opened 2014-06-24
Implement Block-test

BigCrunsh:
The B-test is a test for the two-sample problem for high-dimensions.

This is an implementation of:
- Zaremba, W., Gretton, A., Blaschko, M. B-tests: Low Variance Kernel
  Two-Sample Tests (NIPS)


ararslan:
@BigCrunsh Out of curiosity, was this based on the reference Matlab code that accompanies the paper?

PR #33, opened 2015-02-24
Power Divergence Test

bdeonovic:
Implements power divergence test, which is a family of tests for multinomial proportions that includes the chi-squared test. Also includes several popular multinomial confidence intervals (e.g. sison-glaz, bootstrap)

Since this is the first hypothesis test for a vector of parameters I wasn't quite sure how you guys would want the output summary presented. Let me know what you think!


src/power_divergence.jl
```diff
@@ -0,0 +1,349 @@
+export PowerDivergenceTest, ChisqTest, MultinomialLRT
+
+boundproportion(x) = max(min(x,1),0)
+
+immutable MultinomialData
+  observed::Vector
```
johnmyleswhite: These things should probably be parametric types. As is, they're black holes for the type inference system.

johnmyleswhite:
Generally, this looks good to me.

bdeonovic:
I tidied up the code, squashed commits together. I think this is ready to be pulled now.

simonbyrne:
Great, thanks!

PR #38, opened 2015-08-12
Make PowerDivergenceTest 0.4 compatible

bdeonovic:
Fix Tuple compatibility
Fix formula display issues and grammar
Add multinomial CI documentation


simonster:
I'm not sure why this isn't actually throwing errors on 0.3, just returning an `Array{(Any...,),1}`, but you would need to use  `@compat` before the calls to `Tuple`. But a better approach would be to define `boundproportion` as `max(min(convert(Float64, x),1.0),0.0)`, and then the arrays should not need to be typed.

PR #40, opened 2015-09-25
Fixed spelling of 'Kolmorov' to 'Kolmogorov'.

janmtl:
Just noticed this spelling error in the testname variable. Commit only affects three hard-coded strings.


simonster:
Thanks!

PR #44, opened 2016-01-07
Avoid overflow in PowerDivergenceTest

garborg:
See 0ef5d1b for change.

### DIFF
--- a/src/power_divergence.jl
+++ b/src/power_divergence.jl
@@ -3,39 +3,39 @@ export PowerDivergenceTest, ChisqTest, MultinomialLRT
 typealias Levels{T} @compat(Tuple{UnitRange{T},UnitRange{T}})
 
 function boundproportion{T<:Real}(x::T)
-  max(min(convert(Float64,x),1.0),0.0)
+    max(min(convert(Float64,x),1.0),0.0)
 end
 
 
 immutable PowerDivergenceTest <: HypothesisTest
-  lambda::Float64
-  theta0::Vector{Float64}
-  stat::Float64
-  df::Int64
-  observed::Matrix{Int64}
-  n::Int64
-  thetahat::Vector{Float64}
+    lambda::Float64
+    theta0::Vector{Float64}
+    stat::Float64
+    df::Int64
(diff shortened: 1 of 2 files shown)


garborg:
Make that 86536b6 (I forgot about 0.3 compatibility).

bdeonovic:
I'm not quite sure why I put that in there. It was quite awhile ago now. I think I was following some R code for the chisq test. 

Yes, the R code in the chisq.test function is 

```R
   if (is.na(nr) || is.na(nc) || is.na(nr * nc)) 
      stop("invalid nrow(x) or ncol(x)", domain = NA)
```

I guess I wasn't thinking when I was converting the code. It can probably be removed.

garborg:
@bdeonovic Shoot, sounds like the code in 'src/power_divergence.jl' was adapted from the (GPL-licensed) stats package distributed with R?

If so, I don't think it can stay in here (unless this package was relicensed. I don't think that's desirable, but that would probably be @simonster's call).

bdeonovic:
I'm not sure I'd go as far as to say it was adapted from that code. I used it as a guide to make sure I wasn't missing anything obvious since the Power Divergence Test is a generalization of the Chi Square test.

garborg:
Okay, I'm no expert here -- maybe @simonster will have more specific questions or will ping someone who has more experience with the ins and outs of licensing.

garborg:
Not sure, but I think JuliaStats has avoided integrating implementations where the author has even looked though non-MIT-compatible code implementing the same. The tests implemented in #33 (or at least `PowerDivergenceTest` & `ChisqTest`) pulled in some steps from the corresponding R code.

@simonbyrne @simonster -- It looks like you two are the most involved here atm -- could one of you either, follow up with @bdeonovic and either explicitly approve or remove #33, or ping someone you're comfortable delegating that to?

slundberg:
This bug fix seems separate from the licensing questions. Could the PR get merged separate from that discussion?

andreasnoack:
This was fixed in https://github.com/JuliaStats/HypothesisTests.jl/pull/91

PR #46, opened 2016-02-05
Don't remove -Inf when computing the Anderson-Darling test statistic.

### LINKED ISSUE #45: Is Anderson–Darling working fine?
In the next example, AD tests is failing to reject H0 while KS is getting a P value << 0.05:

```julia
julia> using Distributions

julia> using HypothesisTests
WARNING: Method definition call(Polynomials.Poly, Any) in module Polynomials at /home/diego/.julia/v0.4/Polynomials/src/Polynomials.jl:302 overwritten in module CurveFit at /home/diego/.julia/v0.4/CurveFit/src/linfit.jl:124.

julia> ExactOneSampleKSTest(vcat(rand(Normal(),500), rand(Beta(2,2),500)), Beta(2,2))
Exact one sample Kolmogorov-Smirnov test
----------------------------------------
Population details:
    parameter of interest:   Supremum of CDF differences
    value under h_0:         0.0
    point estimate:          0.23369801954886604

Test summary:
    outcome with 95% confidence: reject h_0
    two-sided p-value:           1.6298124488993274e-48 (extremely significant)

Details:
    number of observations:   1000

julia> OneSampleADTest(vcat(rand(Normal(),500), rand(Beta(2,2),500)), Beta(2,2))
One sample Anderson-Darling test
--------------------------------
Population details:
    parameter of interest:   not implemented yet
    value under h_0:         NaN
    point estimate:          NaN

Test summary:
    two-sided p-value:           1.0 (not significant)

Details:
    number of observations:   1000
    sample mean:              0.22686908598687877
    sample SD:                0.7924243264634008
    A² statistic:             -52.401631357070755

```

Similar example in R... AD test reject H0 (alpha 0.05):

```R
> ad.test(c(rnorm(500), rbeta(500, shape1=2, shape2=2)), function(x){pbeta(x, shape1=2, shape2=2)})

	Anderson-Darling GoF Test

data:  c(rnorm(500), rbeta(500, shape1 = 2, shape2 = 2))  and  function(x) {c(rnorm(500), rbeta(500, shape1 = 2, shape2 = 2))  and      pbeta(x, shape1 = 2, shape2 = 2)c(rnorm(500), rbeta(500, shape1 = 2, shape2 = 2))  and  }
AD = Inf, p-value = 6e-07
alternative hypothesis: NA
```

@wildart, maybe will be easy for you to find the bug.

andreasnoack:
These values occur when observations are outside the support of the
distribution for which we test against and is clearly against the null.
Therefore they should't be set to zero which is strongly in favor of
the null.

Also cut off the polynomial for very large values when computing the
p-values.

Avoid many Infs by using `logcdf` and `logccdf` instead of `log(cdf)` and
`log(1 - cdf)`.

Fixes #45 

cc: @wildart

### DIFF
--- a/src/anderson_darling.jl
+++ b/src/anderson_darling.jl
@@ -12,11 +12,14 @@ function adstats{T<:Real}(x::AbstractVector{T}, d::UnivariateDistribution)
     y = sort(x)
     μ = mean(y)
     σ = std(y)
-    z = cdf(d, (y-μ)/σ)
-    i = 1:n
-    S = ((2*i-1.0)/n) .* (log(z[i])+log(1-z[n+1-i]))
-    S[isinf(S)] = 0. # remove infinity
-    A² = -n-sum(S)
+    A² = convert(typeof(μ), -n)
+    for i = 1:n
+        zi = (y[i] - μ)/σ
+        zni1 = (y[n - i + 1] - μ)/σ
+        lcdfz = logcdf(d, zi)
+        lccdfz = logccdf(d, zni1)
+        A² -= (2*i - 1)/n * (lcdfz + lccdfz)
+    end
     (n, μ, σ, A²)
 end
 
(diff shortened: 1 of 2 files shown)


wildart:
:+1:

PR #48, opened 2016-02-13
add confidence interval calculation for Signed Rank test

rawls238:
@simonster

### DIFF
--- a/src/wilcoxon.jl
+++ b/src/wilcoxon.jl
@@ -31,9 +31,9 @@ function SignedRankTest{T<:Real}(x::AbstractVector{T})
     (W, ranks, signs, tie_adjustment, n, median) = signedrankstats(x)
     n_nonzero = length(ranks)
     if n_nonzero <= 15 || (n_nonzero <= 50 && tie_adjustment == 0)
-        ExactSignedRankTest(W, ranks, signs, tie_adjustment, n, median)
+        ExactSignedRankTest(x, W, ranks, signs, tie_adjustment, n, median)
     else
-        ApproximateSignedRankTest(W, ranks, signs, tie_adjustment, n, median)
+        ApproximateSignedRankTest(x, W, ranks, signs, tie_adjustment, n, median)
     end
 end
 SignedRankTest{T<:Real,S<:Real}(x::AbstractVector{T}, y::AbstractVector{S}) = SignedRankTest(x - y)
@@ -51,10 +51,10 @@ function signedrankstats{S<:Real}(x::AbstractVector{S})
    (W, ranks, nonzero_x .> 0, tieadj, length(x), median(x))
 end
 
-
 ## EXACT WILCOXON SIGNED RANK TEST
 
 immutable ExactSignedRankTest{T<:Real} <: HypothesisTest
+    vals::Vector            # original values
     W::Float64              # test statistic: Wilcoxon rank-sum statistic
     ranks::Vector{T}        # ranks without ties (zero values)
     signs::BitArray{1}      # signs of input of ranks
@@ -63,7 +63,7 @@ immutable ExactSignedRankTest{T<:Real} <: HypothesisTest
     median::Float64         # sample median
 end
 ExactSignedRankTest{T<:Real}(x::AbstractVector{T}) =
-    ExactSignedRankTest(signedrankstats(x)...)
+    ExactSignedRankTest(x, signedrankstats(x)...)
 ExactSignedRankTest{S<:Real,T<:Real}(x::AbstractVector{S}, y::AbstractVector{T}) =
     ExactSignedRankTest(x - y)
 
@@ -100,7 +100,7 @@ function signedrankenumerate(x::ExactSignedRankTest)
     (le/tot, gr/tot)
 end
 
-function pvalue(x::ExactSignedRankTest; tail=:both) 
+function pvalue(x::ExactSignedRankTest; tail=:both)
     n = length(x.ranks)
     if n == 0
         1
@@ -129,10 +129,13 @@ function pvalue(x::ExactSignedRankTest; tail=:both)
     end
 end
 
+ci(x::ExactSignedRankTest, alpha::Float64=0.05; tail=:both) = calculate_ci(x.vals, alpha; tail=tail)
+
 
 ## APPROXIMATE SIGNED RANK TEST
 
 immutable ApproximateSignedRankTest{T<:Real} <: HypothesisTest
+    vals::Vector            # original values
     W::Float64              # test statistic: Wilcoxon rank-sum statistic
     ranks::Vector{T}        # ranks without ties (zero values)
     signs::BitArray{1}      # signs of input of ranks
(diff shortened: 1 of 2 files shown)


src/wilcoxon.jl
```diff
@@ -129,10 +129,13 @@ function pvalue(x::ExactSignedRankTest; tail=:both)
     end
 end
 
+ci(x::ExactSignedRankTest, alpha::Float64=0.05; tail=:both) = calculate_ci(x.vals, alpha; tail=tail)
+
 
 ## APPROXIMATE SIGNED RANK TEST
 
 immutable ApproximateSignedRankTest{T<:Real} <: HypothesisTest
+    vals::Vector{T}         # original values
```
rawls238: let me know if anybody has any objections to this but I need the original data to compute the confidence intervals (let me know if you know of any alternative ways to compute the CIs without having to have access to the original data)

src/wilcoxon.jl
```diff
@@ -179,6 +182,43 @@ function pvalue(x::ApproximateSignedRankTest; tail=:both)
     end
 end
 
+ci(x::ApproximateSignedRankTest, alpha::Float64=0.05; tail=:both) = calculate_ci(x.vals, alpha; tail=tail)
+
+function calculate_ci(x::AbstractVector, alpha::Float64=0.05; tail=:both)
```
rawls238: hm, now that I think more about this, I'm not actually sure that this method will work for `ApproximateSignedRankTest`, but going to need to test it out more to see

src/wilcoxon.jl
```diff
@@ -51,10 +51,10 @@ function signedrankstats{S<:Real}(x::AbstractVector{S})
    (W, ranks, nonzero_x .> 0, tieadj, length(x), median(x))
 end
 
-
 ## EXACT WILCOXON SIGNED RANK TEST
 
 immutable ExactSignedRankTest{T<:Real} <: HypothesisTest
+    vals::Vector            # original values
```
rawls238: one possible alternative to this is for me to just compute what I need for CI on initialization and just add it as a field here instead of retaining it as a field on the type.

rawls238:
bump :)

rawls238:
@andreasnoack @simonbyrne could you guys take a look over this?

andreasnoack:
@rawls238 Could you explain the method?

rawls238:
@andreasnoack it's adapted from this set of lecture notes (section 2.3): http://www.stat.umn.edu/geyer/old03/5102/notes/rank.pdf

rawls238:
Bump on this...

src/wilcoxon.jl
```diff
@@ -179,6 +182,43 @@ function pvalue(x::ApproximateSignedRankTest; tail=:both)
...
+function calculate_ci(x::AbstractVector, alpha::Float64=0.05; tail=:both)
+    check_alpha(alpha)
+
+    if tail == :both
+        c = 1 - alpha
+    else
+        c = 1 - 2 * alpha
+    end
+    n = length(x)
+    m = convert(Int, n * (n + 1) / 2)
```
andreasnoack: No need to promote back and forth to `Float64`. You can write `div(n * (n - 1), 2)`.

src/wilcoxon.jl
```diff
@@ -179,6 +182,43 @@ function pvalue(x::ApproximateSignedRankTest; tail=:both)
...
+    k_range = 1:(m/2)
```
andreasnoack: ...and `div(m,2)` here.

src/wilcoxon.jl
andreasnoack: Not necessary to restrict to `Flaot64`. `Real` should be sufficient.

andreasnoack:
I've made a few code comments. I'm not that familiar with this calculation but I've read the note you linked to and I believe your implementation is correct. Please provide a link to the note as a comment in the code. With that and rebase I think it can be merged. Thanks for submitting this.

src/wilcoxon.jl
KristofferC: This is not a concrete type but maybe it doesn't matter for performance.

andreasnoack: Good catch. Changing it to `Vector{T}` should be an easy fix.

src/wilcoxon.jl
```diff
@@ -129,10 +129,13 @@ function pvalue(x::ExactSignedRankTest; tail=:both)
     end
 end
 
+ci(x::ExactSignedRankTest, alpha::Float64=0.05; tail=:both) = calculate_ci(x.vals, alpha; tail=tail)
+
 
 ## APPROXIMATE SIGNED RANK TEST
 
 immutable ApproximateSignedRankTest{T<:Real} <: HypothesisTest
+    vals::Vector            # original values
```
KristofferC: same here

src/wilcoxon.jl
rawls238: good catch will change that

src/wilcoxon.jl
rawls238: annoyingly changing this `Vector{T}` causes issues with the call from https://github.com/JuliaStats/HypothesisTests.jl/blob/master/src/wilcoxon.jl#L155

going to dig into why this is the case tomorrow but if anyone has any ideas let me know

rawls238:
@andreasnoack this should be good to go, thanks for reviewing

src/wilcoxon.jl
```diff
@@ -129,12 +129,15 @@ function pvalue(x::ExactSignedRankTest; tail=:both)
...
+ci(x::ExactSignedRankTest, alpha::Real=0.05; tail=:both) = calculate_ci(x.vals, alpha; tail=tail)
+
 
 ## APPROXIMATE SIGNED RANK TEST
 
 immutable ApproximateSignedRankTest{T<:Real} <: HypothesisTest
+    vals::AbstractVector{T} # original values
     W::Float64              # test statistic: Wilcoxon rank-sum statistic
-    ranks::Vector{T}        # ranks without ties (zero values)
+    ranks::Vector{Float64} # ranks without ties (zero values)
```
rawls238: this always just returns `Float64` anyways so to prevent the weird coverage errors I ended up just explicitly defining this type, let me know if anyone has any issues with this

src/wilcoxon.jl
```diff
@@ -179,6 +182,44 @@ function pvalue(x::ApproximateSignedRankTest; tail=:both)
...
+        c = 1 - alpha
+    else
+        c = 1 - 2 * alpha
+    end
+    n = length(x)
+    m = div(n * (n + 1), 2)
+    k_range = 1:div(m, 2)
+    l = [1 - 2 * psignrank(i, n, true) for i in k_range]
+    k = indmin(abs(l-c))
+    vals = Float64[]
```
KristofferC: Couldn't the function be parameterized on the element type of x 
```jl
function calculate_ci{T <: Real}(x::AbstractVector{T}, alpha::Real=0.05; tail=:both)
```
and then have `vals = T[]` so that vals is the same type as the input data?

rawls238: What would the benefit of that be? It seems like it would occasionally lead to incorrect behavior since the elements of the vals array won't always be the same type as the input array (e.g. If the inputs are all ints, say [1, 2], then we would still want to have the vals array be Float64 not Int)

KristofferC: Yes, possibly with `T <: AbstractFloat` then but I see that other types just use `Vector{Float64}` so it should be fine as is.

andreasnoack: Eventually, it would be great to remove `Float64` assumptions when possible but I think it is fine to do that another time.

src/wilcoxon.jl
andreasnoack: This is okay. We can try to remove the `Float64` assumption later on.

src/wilcoxon.jl
```diff
@@ -129,12 +129,15 @@ function pvalue(x::ExactSignedRankTest; tail=:both)
     end
 end
 
+ci(x::ExactSignedRankTest, alpha::Real=0.05; tail=:both) = calculate_ci(x.vals, alpha; tail=tail)
+
 
 ## APPROXIMATE SIGNED RANK TEST
 
 immutable ApproximateSignedRankTest{T<:Real} <: HypothesisTest
+    vals::AbstractVector{T} # original values
```
andreasnoack: Oops. This is unfortunate because the type is abstract. I guess you changed away from `Vector` for a reason so could you share the reason here?

rawls238: no this was an oversight I should just move this back to `Vector` for now

rawls238:
@andreasnoack any more issues here?

andreasnoack:
No. I think it is ready. Thanks for the contribution (and the patience.)

PR #52, opened 2016-05-04
WIP: fix computation of g2 in PowerDivergenceTest

wehlutyk:
The [May & Johnson (2000) SAS macro](https://www.researchgate.net/profile/William_Johnson9/publication/5142788_Constructing_Two-Sided_Simultaneous_Confidence_Intervals_for_Multinomial_Proportions_for_Small_Counts_in_a_Large_Number_of_Cells/links/0deec52a60bb3cdf50000000.pdf) (see page 14 of the pdf) from which this code is inspired defines `g2 = s4/(s2^2)` (with a division), not `g2 = s4*(s2^2)` (with a multiplication), as is also described in [Levin (1981)](http://projecteuclid.org/download/pdf_1/euclid.aos/1176345593) (see page 3 of the pdf). (The [MultinomialCI](https://cran.r-project.org/web/packages/MultinomialCI/index.html) package in R also has this right.)

I don't have access to Agresti (2007) from which the test code is inspired (according to the comments), but with this correction the tests fail on line 19 of `test/power_divergence.jl` (the confidence intervals). How were those tests constructed? Do the values come directly from Agresti (2007)?

### DIFF
--- a/src/power_divergence.jl
+++ b/src/power_divergence.jl
@@ -149,7 +149,7 @@ function ci_sison_glaz(x::PowerDivergenceTest, alpha::Float64; skew_correct::Boo
     s1,s2,s3,s4 = mapslices(sum,m,1)
     z  = (x.n-s1)/sqrt(s2)
     g1 = s3/(s2^(3/2))
-    g2 = s4*(s2^2)
+    g2 = s4/(s2^2)
     
     poly = 1 + g1 * (z^3 - 3*z)/6 + g2*(z^4-6*z^2+3)/24 + g1^2 * (z^6 - 15*z^4 + 45*z^2 -15)/72
     f = poly*exp(-z^2/2)/(sqrt(2*pi))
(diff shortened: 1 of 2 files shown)


nalimilan:
Agresti's book only reports the standardized residuals, so other values here may well be incorrect. One should check against R, though. No idea what's the name of the function the test file refers to.

johnmyleswhite:
We should really consider testing everything in this package using simulations. This kind of error is something that most simulations would catch, hopefully with very little computational cost if the error is as severe as it seems to me.

wehlutyk:
I'm implementing the confidence-interval part of this for statsmodels, and running into what looks a lot like floating-point arithmetic errors getting blown up in the computation of gamma_1, gamma_2, and the hermite polynomials (see statsmodels/statsmodels#2953 and the comments in the `results_multinomial_proportions.py` file, where test values compared to R MultinomialCI values can't be more precise than 4 decimals).

Fixing the test values here (using values computed with R MultinomialCI instead) seems to run into the same problem (can't figure out any other reason). Testing with simulations would indeed be a good solution to all this :)

andreasnoack:
Superseded by #101

PR #55, opened 2016-06-12
Updated docstring for binomial confidence intervals

juliangehring:
The existing docstring described the available confidence interval methods for power divergence test, not for the binomial test.

The new docstring now describes the binomial CI methods that are used.


nalimilan:
Thanks, and sorry for the delay!

PR #57, opened 2016-06-20
Correct P value default

diegozea:
This fixes a small bug with  χ² test... `show` calls `pvalue(test)` and expect `tail` to be `:both`

image


diegozea:
Since the test p value has sense for `tail=:right`, maybe is better to close this and make explicit `tail=:both` in `show`...

diegozea:
I close this in favor of https://github.com/JuliaStats/HypothesisTests.jl/pull/58

PR #62, opened 2016-07-07
WIP: Get things working on 0.5

ararslan:
A release candidate for 0.5 is quickly approaching, so we should make sure this no longer fails the nightlies. I've also bumped minimum Julia version to 0.4 and allowed testing on both OS X and Linux rather than just Linux.

I've also done a tad of general cleanup (e.g. removing tabs).


ararslan:
Okay, so there are some *very* disconcerting test failures on 0.5. In particular, there's a test that passes on 0.4 but on 0.5 the computed p-value is 0.9617 and it's supposed to be 0.2413 according to the test. 😕

ararslan:
This, coupled with #65, should get things working on 0.5. (Fingers crossed)

ararslan:
https://github.com/JuliaLang/julia/commit/02aeb44299d090d50d2c58e004f58a8b8d4f3da6 is what's breaking the printing for Fisher's exact test on 0.5 since the size arguments in `print_matrix` were removed.

ararslan:
*sigh* There seem to be a number of precision-related test failures.

test/runtests.jl
```diff
@@ -1,3 +1,15 @@
...
+# approximate array equality is incorrect when either limit is
+# infinite.
+macro test_ci_approx(x::Expr, y::Expr)
+    @eval begin
+        Base.Test.@test typeof($x) <: Tuple{Real,Real}
+        Base.Test.@test typeof($x) == typeof($y)
+        Base.Test.@test all(map(isapprox, $x, $y))
+    end
+end
+
```
KristofferC: Shouldn't this be something like:

```jl
macro test_ci_approx(x, y)
     return quote
        Base.Test.@test typeof($(esc(x))) <: Tuple{Real,Real}
        Base.Test.@test typeof($(esc(x))) == typeof($(esc(y)))
        Base.Test.@test all(map(isapprox, $(esc(x)), $(esc(y))))
        end
 end
```

ararslan: I dunno, it works as-is. But I'm not a macro expert so if you say it should be that then I'll change it to that.

KristofferC: A good way to think about macros is that they take an expression and return another expression.  Evalings stuff inside a macro is almost always the wrong thing to do.

ararslan: Oh wow, that's really helpful, thank you! I'll definitely keep that in mind.

ararslan:
Travis is passing on 0.4 and 0.5!! 🎉

src/HypothesisTests.jl
```diff
@@ -27,7 +27,7 @@ VERSION >= v"0.4.0-dev+6521" && __precompile__()
 module HypothesisTests
 
 using Distributions, Roots, StatsBase, Compat, Combinatorics
-using Rmath: pwilcox
+using Rmath: pwilcox, psignrank
```
tkelman: so these were right with the rather old rmath bindings from that package? statsfuns didn't expose something more recent?

ararslan: Correct

Issue #63, opened 2016-07-14
Dependency broken on v0.5


ranjanan:
The dependency `CurveFit` in the `REQUIRE` file is broken on v0.5. (https://github.com/pjabardo/CurveFit.jl/issues/6)

There is a PR to fix this. (https://github.com/pjabardo/CurveFit.jl/pull/7) Is there anyone here who can review and merge please?

ararslan:
#62 here removes the CurveFit dependency entirely, but there's still work to be done on that PR.

Since CurveFit is on a user's account rather than an organization's, it would require a collaborator to merge. Looks like the only contributor to that repo so far has been the original author, so I'd be surprised if anyone here had the permissions necessary to merge there.

ararslan:
Actually, it appears that CurveFit may have been effectively superseded by JuliaOpt/LsqFit.jl.

ranjanan:
@ararslan I was aware of that, but currently for users of this package on v0.5, this package would break because of the dependency. This was to address specifically that. 

>Actually, it appears that CurveFit may have been effectively superseded by JuliaOpt/LsqFit.jl.

Do you think it makes sense to remove CurveFit from the REQUIRE and add LsqFit instead?

ararslan:
Maybe, but that wouldn't fix all of the issues this package currently has on 0.5. If you look at the [0.5 test log](http://pkg.julialang.org/logs/HypothesisTests_0.5.log) from the package evaluator, it's having the p-value issue with the block test that I mentioned in #62. I'm not familiar with that particular test so I'm not entirely sure how to proceed with fixing that. 😕

ranjanan:
Actually the function `ccdf` from `Distributions.jl` is behaving differently. That's why you're seeing this error. On v0.4, it gives 0.24 and on v0.5 it gives 0.96. Digging into it....

ararslan:
Great catch, @ranjanan! Thanks for looking into it!

ranjanan:
Oh wait, sorry. What actually is that the inputs to that function are changing when you call the `BTest` constructor.

ranjanan:
Hey, I managed to trace the issue. This is due difference in behaviour of `randperm` from v0.4 to v0.5 even with the same seed. Consider the following code. 

```julia
function thing(uniform = false)
    srand(1)

[...]

PR #65, opened 2016-07-22
RFC: Remove block test entirely

ararslan:
Ref https://github.com/JuliaStats/HypothesisTests.jl/issues/63#issuecomment-234658572. Assuming this is the way we want to go (rather than, say, quarantining the files), once the block test things are figured out we can revert this PR.

cc @simonbyrne @ranjanan


ararslan:
The failures are unrelated and the coverage drop is expected (I would think, given that tests are being removed).

ararslan:
No one has spoken up so I'm inclined to go ahead and merge this. Will do so a little later barring any objections.

PR #76, opened 2016-09-13
added permutation test

bjarthur:
if you like it, i'll work on adding docs.


src/permutation.jl
```diff
@@ -0,0 +1,65 @@
...
+end
+
+"""
+    PermutationTest(x::Vector, y::Vector, f::Function [, n::Int])
+
+Perform a permutation test (a.k.a. randomization test) of the null hypothesis
+that the probability of `f(x)` is equal to `f(y)`.
+
+If `n` is *not* supplied, all possible permutations are sampled.  Otherwise,
+only `n` of the `factorial(length(vec(x,y)))` permutations are sampled.
```
ararslan: `vec(x, y)` doesn't work, at least in Julia 0.5. Perhaps clearer anyway as `factorial(length(x) + length(y))`.

src/permutation.jl
```diff
@@ -0,0 +1,65 @@
...
+
+function ptstats(x,y,f)
+    xy = vcat(x,y)
+    rx = 1 : length(x)
+    ry = length(xy)-length(y)+1 : length(xy)
+    T = typeof(f(x))  # is there a better way?  e.g., does a funtype() exist, like eltype()??
+    (xy, rx, ry, T)
+end
+
+immutable PermutationTest <: HypothesisTests.HypothesisTest
```
ararslan: Shouldn't need to qualify `HypothesisTest` here

src/permutation.jl
```diff
@@ -0,0 +1,65 @@
+export PermutationTest
+
+function ptstats(x,y,f)
+    xy = vcat(x,y)
+    rx = 1 : length(x)
+    ry = length(xy)-length(y)+1 : length(xy)
+    T = typeof(f(x))  # is there a better way?  e.g., does a funtype() exist, like eltype()??
```
ararslan: What you're doing here is setting `T` to be the type of the result after applying `f` to `x`. So if `f(x)` returns 1, this will be `Int`, for example. Is this what you're after? I'm not sure what you mean regarding `eltype`.

src/permutation.jl
```diff
@@ -0,0 +1,65 @@
...
+that the probability of `f(x)` is equal to `f(y)`.
+
+If `n` is *not* supplied, all possible permutations are sampled.  Otherwise,
+only `n` of the `factorial(length(vec(x,y)))` permutations are sampled.
+"""
+function PermutationTest{R<:Real,S<:Real}(x::AbstractVector{R}, y::AbstractVector{S}, f::Function)
+    xy, rx, ry, T = ptstats(x,y,f)
+    P = permutations(xy)
+    xsamples = Array(T,length(P))
+    ysamples = Array(T,length(P))
```
ararslan: You're allowing the element types of `x` and `y` to differ, but you're specifying the vectors of samples to have a common type, namely the return type of `f` as applied to `x`. What if `f` has a different return type when applied to `y`?

ararslan: Also the preferred syntax for this would be `Array{T}(length(P))`, but since you know this will only be 1-dimensional, you can specify it as `Vector{T}(length(P))`.

src/permutation.jl
```diff
@@ -0,0 +1,65 @@
...
+    for (i,p) in enumerate(P)
+        xsamples[i] = f(p[rx])
+        ysamples[i] = f(p[ry])
+    end
```
ararslan: If you know that `rx`, `ry`, and `i` are all valid indices for the respective arrays, you should be able to do `@inbounds for (i, p) ...`, which can be faster.

src/permutation.jl
```diff
@@ -0,0 +1,65 @@
+export PermutationTest
+
+function ptstats(x,y,f)
+    xy = vcat(x,y)
+    rx = 1 : length(x)
+    ry = length(xy)-length(y)+1 : length(xy)
```
ararslan: Looks like you're constructing the concatenation of `x` and `y`, but only using it to compute the length. This will be inefficient when `x` and/or `y` are large.

`length(vcat(x, y))` should be `length(x) + length(y)`, right? Then `length(vcat(x, y)) - length(y) + 1 == length(y) + length(x) - length(y) + 1`, which simplifies to `length(x) + 1`. So you can write this as
```julia
nx = length(x)
rx = 1:nx
ry = nx+1:nx+length(y)
```

src/permutation.jl
```diff
@@ -0,0 +1,65 @@
...
+        shuffle!(xy)
+        xsamples[i] = f(xy[rx])
+        ysamples[i] = f(xy[ry])
+    end
+    PermutationTest(f(x) - f(y), xsamples - ysamples)
+end
+
+function pvalue(apt::PermutationTest; tail=:both)
+    if tail == :both
+        count = sum(abs(apt.observation) .>= abs.(apt.samples))
```
ararslan: The dot vectorization syntax was introduced in Julia 0.5, so the use here is causing the tests to fail on 0.4. To fix that you can simply do `@compat abs.(apt.samples)`.

src/permutation.jl
simonster: `@inbounds` won't make a difference since `rx` and `ry` are ranges and `p[rx]` and `p[ry]` are allocating new vectors. This would ideally be written in a way that does not allocate memory, although that's somewhat difficult because `permutations` itself allocates new vectors.

ararslan: Ah good catch, forgot that `rx` and `ry` were ranges

src/permutation.jl
```diff
@@ -0,0 +1,65 @@
...
+
+function PermutationTest{R<:Real,S<:Real}(x::AbstractVector{R}, y::AbstractVector{S}, f::Function, n::Int)
+    xy, rx, ry, T = ptstats(x,y,f)
+    xsamples = Array(T,n)
+    ysamples = Array(T,n)
+    for i = 1:n
+        shuffle!(xy)
+        xsamples[i] = f(xy[rx])
+        ysamples[i] = f(xy[ry])
+    end
```
simonster: If you create views into `xy` outside the loop, then you don't need to do any indexing in the loop here, which will avoid allocation.

src/permutation.jl
bjarthur: `xy` is returned from `ptstats` and used by it's callers.

src/permutation.jl
bjarthur: oops, meant to say `vcat` not `vec`.

src/permutation.jl
bjarthur: `T` is now `typeof(f(xy))`.

also, why is `Vector{T}(length(P))` preferred to `Array(T,length(P))`?  is there a performance advantage, or is it just stylistic?

src/permutation.jl
bjarthur: is there a function/macro that intputs another function and returns it's return type?  i would think that at least in some cases the type of returned variables could by inferred from the AST.  here i just called the function because i'm unware that this ability currently exists in julia.

src/permutation.jl
bjarthur: i changed `xy[rx]` to `view(xy,rx)`.  is that what you suggested?  not sure how to do this "outside the loop".

src/permutation.jl
ararslan: The syntax `Array(T, n)` is deprecated in favor of `Array{T}(n)`. `Vector{T}(n)` is the same as `Array{T,1}(n)`, just clearer.

bjarthur:
okay, so i think i've addressed everyone's comments.  let me know if i missed anything or you have other suggestions.

re. documentation.  what's the long term plan for this package?  any consideration of migrating to docstrings and Documenter.jl?  if so, i'd be inclined to just make that change now, instead of manually duplicating permutation test's docstrings.

bjarthur:
rebased

bjarthur:
tests pass.  can someone please review?  thanks.

src/permutation.jl
```diff
@@ -0,0 +1,65 @@
+export PermutationTest
+
+function ptstats(x,y,f)
+    xy = vcat(x,y)
+    rx = 1 : length(x)
+    ry = length(xy)-length(y)+1 : length(xy)
+    T = typeof(f(xy))  # is there a better way?  e.g., does a funtype() exist, like eltype()??
```
nalimilan: Yes, a better way would be to use an array comprehension in the constructor below, so that the element type is computed automatically.

src/permutation.jl
```diff
@@ -0,0 +1,65 @@
...
+    (xy, rx, ry, T)
+end
+
+immutable PermutationTest <: HypothesisTest
+    observation
```
nalimilan: Should add a type parameter so that the fields are concretely typed.

src/permutation.jl
```diff
@@ -0,0 +1,65 @@
+export PermutationTest
+
+function ptstats(x,y,f)
+    xy = vcat(x,y)
+    rx = 1 : length(x)
```
nalimilan: No spaces around `:` here and below. Use parentheses where needed to make the precedence rules more immediately visible.

src/permutation.jl
```diff
@@ -0,0 +1,65 @@
...
+        shuffle!(xy)
+        xsamples[i] = f(view(xy,rx))
+        ysamples[i] = f(view(xy,ry))
+    end
+    PermutationTest(f(x) - f(y), xsamples - ysamples)
+end
+
+function pvalue(apt::PermutationTest; tail=:both)
+    if tail == :both
+        count = sum(abs(apt.observation) .>= @compat abs.(apt.samples))
```
nalimilan: Better use a generator to avoid allocating a temporary `Vector{Bool}`.

test/permutation.jl
```diff
@@ -0,0 +1,12 @@
+using HypothesisTests, Base.Test
+
+@test pvalue(PermutationTest([1,2,3],[4,5,6],mean), tail=:both) - 1.0 <= 1e-4
```
nalimilan: You can use `isapprox` here.

src/permutation.jl
```diff
@@ -0,0 +1,65 @@
...
+end
+
+"""
+    PermutationTest(x::Vector, y::Vector, f::Function [, n::Int])
+
+Perform a permutation test (a.k.a. randomization test) of the null hypothesis
+that the probability of `f(x)` is equal to `f(y)`.
+
+If `n` is *not* supplied, all possible permutations are sampled.  Otherwise,
+only `n` of the `factorial(length(x)+length(y))` permutations are sampled.
```
nalimilan: Should mention that they are chosen at random.

src/permutation.jl
```diff
@@ -0,0 +1,65 @@
...
+immutable PermutationTest <: HypothesisTest
+    observation
+    samples
+end
+
+"""
+    PermutationTest(x::Vector, y::Vector, f::Function [, n::Int])
+
+Perform a permutation test (a.k.a. randomization test) of the null hypothesis
+that the probability of `f(x)` is equal to `f(y)`.
```
nalimilan: Is is really the "probability"? IIUC that could be any value, right?

src/permutation.jl
```diff
@@ -0,0 +1,65 @@
...
+
+function ptstats(x,y,f)
+    xy = vcat(x,y)
+    rx = 1 : length(x)
+    ry = length(xy)-length(y)+1 : length(xy)
+    T = typeof(f(xy))  # is there a better way?  e.g., does a funtype() exist, like eltype()??
+    (xy, rx, ry, T)
+end
+
+immutable PermutationTest <: HypothesisTest
```
ararslan: `struct`

src/permutation.jl
```diff
@@ -0,0 +1,65 @@
...
+"""
+    PermutationTest(x::Vector, y::Vector, f::Function [, n::Int])
+
+Perform a permutation test (a.k.a. randomization test) of the null hypothesis
+that the probability of `f(x)` is equal to `f(y)`.
+
+If `n` is *not* supplied, all possible permutations are sampled.  Otherwise,
+only `n` of the `factorial(length(x)+length(y))` permutations are sampled.
+"""
+function PermutationTest{R<:Real,S<:Real}(x::AbstractVector{R}, y::AbstractVector{S}, f::Function)
```
ararslan: Should use `where` now

src/permutation.jl
```diff
@@ -0,0 +1,65 @@
...
+    xsamples = Vector{T}(length(P))
+    ysamples = Vector{T}(length(P))
+    for (i,p) in enumerate(P)
+        xsamples[i] = f(view(p,rx))
+        ysamples[i] = f(view(p,ry))
+    end
+    PermutationTest(f(x) - f(y), xsamples - ysamples)
+end
+
+function PermutationTest{R<:Real,S<:Real}(x::AbstractVector{R}, y::AbstractVector{S}, f::Function, n::Int)
```
ararslan: `where` here too

src/permutation.jl
ararslan: No longer necessary to `@compat` this, since support for 0.4 has been dropped (and 0.5 as well)

bjarthur: you mean like this instead?  `count = sum(abs(apt.observation) .>= x for x in abs.(apt.samples))`.  i'm not sure that's faster.  for example:

```
julia> @time sum(rand(1000) .> 0.5)
  0.029506 seconds (6.24 k allocations: 352.697 KiB)
517

julia> @time sum(x .> 0.5 for x in rand(1000))
  0.052041 seconds (15.20 k allocations: 852.639 KiB)
482
```

nalimilan: Don't benchmark in global scope.

src/permutation.jl
bjarthur: when pvalue is called on a PermutationTest it returns a value between 0 and 1.  what would you prefer to call it other than a probability?

src/permutation.jl
bjarthur: i'm not sure what you mean.  can you be more specific?

src/permutation.jl
nalimilan: If you refer to the p-value, the wording is wrong. Shouldn't it be something like "of the null hypothesis that `f(x)` is equal to `f(y)`"? Then the p-value is something different which can be obtained via `pvalue` and which can be mentioned later, but is separate from the definition of the tested hypothesis. See examples in other tests.

src/permutation.jl
```diff
@@ -0,0 +1,71 @@
...
+end
+
+"""
+    ExactPermutationTest(x::Vector, y::Vector, f::Function)
+
+Perform a permutation test (a.k.a. randomization test) of the null hypothesis
+that the probability of `f(x)` is equal to `f(y)`.  All possible permutations are sampled.
+"""
+function ExactPermutationTest(x::AbstractVector{R}, y::AbstractVector{S},
+            f::Function) where {R<:Real,S<:Real}
```
nalimilan: Incorrect indentation here and below.

src/permutation.jl
nalimilan: I mean drop the `for (i,p) in enumerate(P)` loop, and instead do `xsamples = [f(view(p, rx)) for p in P]`.

bjarthur: ahah.  but...  is omitting a call to `f` really preferable to looping over `P` twice?  i'm also not sure how to similarly change the `shuffle` call in `ApproximatePermutationTest`.

nalimilan: The advantage is not omitting a call to `f`, it's relying on the compiler to do inference for us. For example, if `f` returned elements of heterogeneous types (which admittedly isn't typical, but still), using the type of the first element wouldn't be correct. I'm not sure how costly going over `P` is, though.

For `ApproximatePermutationTest`, you could use a single comprehension and store an array of two-element tuples.

bjarthur:
tests pass on 0.6.  i think i've addressed all comments.

src/permutation.jl
```diff
@@ -0,0 +1,59 @@
...
+function ApproximatePermutationTest(x::AbstractVector{R}, y::AbstractVector{S},
+                                    f::Function, n::Int) where {R<:Real,S<:Real}
+    xy, rx, ry = ptstats(x,y,f)
+    samples = [(shuffle!(xy); f(view(xy,rx)) - f(view(xy,ry))) for i = 1:n]
+    PermutationTest(f(x) - f(y), samples)
+end
+
+function pvalue(apt::PermutationTest; tail=:both)
+    if tail == :both
+        count = sum(abs(apt.observation) .>= x for x in abs.(apt.samples))
```
nalimilan: Should use `>=`. Same below. Also better call `abs` inside the loop.

src/permutation.jl
```diff
@@ -0,0 +1,59 @@
+export ExactPermutationTest, ApproximatePermutationTest
+
+function ptstats(x,y,f)
+    xy = vcat(x,y)
+    rx = 1:length(x)
+    ry = length(xy)-length(y)+1 : length(xy)
```
nalimilan: Using parentheses instead of spacing to make precedence more obvious would be more standard IMHO.

bjarthur:
done.  and still passing tests.

andreasnoack:
We should probably allow failures on nightly in the `.travis.yml`.

yakir12:
Sorry for necro-posting. I have two reasons to:

1) It is not clear to me why `ptstats` needs the `f` function as an argument ([see here](https://github.com/JuliaStats/HypothesisTests.jl/blob/750598dc069d2b5da071aad8b987edf4c05bbac5/src/permutation.jl#L3)). It doesn't use it for anything, and it doesn't seem to be used for typing? Or is it?
2) how complicated would it be to allow for these permutation tests to accept more than 2 groups for testing (i.e. a table with a variable and a grouping column, or just a variable vector and a label vector)? I can't seem to find clear directions on how that would even look like. 

Thanks!

nalimilan:
>     1. It is not clear to me why `ptstats` needs the `f` function as an argument ([see here](https://github.com/JuliaStats/HypothesisTests.jl/blob/750598dc069d2b5da071aad8b987edf4c05bbac5/src/permutation.jl#L3)). It doesn't use it for anything, and it doesn't seem to be used for typing? Or is it?

Good catch. Sounds quite trivial to fix. Would you make a PR?

> how complicated would it be to allow for these permutation tests to accept more than 2 groups for testing (i.e. a table with a variable and a grouping column, or just a variable vector and a label vector)? I can't seem to find clear directions on how that would even look like.

No idea. Does that exist in the literature?

yakir12:
done: https://github.com/JuliaStats/HypothesisTests.jl/pull/197

yakir12:
Apparently, it is supposed to be possible. I need to try it out, Dave F Kleinschmidt has been answering a lot of my questions, but we'll see if I manage to build a MWE.

Issue #59, opened 2016-06-20
doc strings missing for `HypothesisTest`s


diegozea:
Would be great to have documentation on the REPL for `HypothesisTest` types. The documentation on the page is fine, it could be moved into doc strings. 
Also, It would be great to add an _assumptions_ section for each test. 

Best,

BenjaminBorn:
The reverse problem also exists. I have added a couple of tests that have docstrings but do not appear in the online documentation.

Would it make sense to move the complete documentation to Documenter.jl? What would be the first step? Transfer the information in the `.rst` files into docstrings?

ararslan:
@BenjaminBorn Yes, we definitely want to set up Documenter here. I've opened an issue for it: https://github.com/JuliaStats/HypothesisTests.jl/issues/103.

PR #82, opened 2016-10-30
added example usage of ChiSqTest and MultinomialLRT, fixed typos

cjprybol:
I didn't see any example usage to accompany the text description, so I thought I'd offer some.

I reviewed these edits using https://livesphinx.herokuapp.com/, and they look good there. I'm having trouble getting Sphinx to run locally so that web app is currently the only test I've run to confirm these changes will render properly

### DIFF
--- a/doc/parametric/power_divergence_test.rst
+++ b/doc/parametric/power_divergence_test.rst
@@ -10,8 +10,8 @@ Power Divergence Test
     those in ``theta0``, or are all equal if ``theta0`` is not given.
 
     If ``x`` is a matrix with at least two rows and columns, it is taken as a two-dimensional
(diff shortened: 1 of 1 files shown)


cjprybol:
Or maybe it would be better to write these as doc strings, as mentioned in https://github.com/JuliaStats/HypothesisTests.jl/issues/59?

Either way, I'd like to write up some examples for the various functions of the package if others think that would be useful.

cjprybol:
Closing this on account that I think I answered my own question. Found helpful discussions here https://github.com/JuliaStats/DataFrames.jl/issues/775 and here https://github.com/JuliaStats/Distributions.jl/issues/550 on the topic of writing documentation. I'll reopen a new pull request after I've finished reading the docs on making docs

PR #86, opened 2017-01-28
Adding the arcsine binomial proportion confidence interval

DCLukas:
Adding another method for CI calculation & making use of Julia Unicode support

### DIFF
--- a/src/binomial.jl
+++ b/src/binomial.jl
@@ -85,6 +85,8 @@ function StatsBase.confint(x::BinomialTest, alpha::Float64=0.05; tail=:both, met
             ci_jeffrey(x, alpha)
         elseif method == :agresti_coull
             ci_agresti_coull(x, alpha)
+        elseif method == :arcsine
+            ci_arcsine(x, alpha)
         else
             throw(ArgumentError("method=$(method) is not implemented yet"))
         end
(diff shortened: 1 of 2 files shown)


ararslan:
Adding the confidence interval method is fine, but I think the second commit should be removed as it's unnecessary and non-user-facing code churn that just adds noise to the PR diff.

DCLukas:
No problem!

ararslan:
Oh, the PR isn't merged so this shouldn't be closed yet; I only submitted an approval. I figured I'd leave it open for a bit to let others have the chance to comment before it gets merged.

src/binomial.jl
```diff
@@ -131,6 +133,13 @@ function ci_wilson(x::BinomialTest, alpha::Float64=0.05)
     (μ-q*σ, μ+q*σ)
 end
 
+# Arcsine transformation interval
+function ci_arcsine(x::BinomialTest, alpha::Float64=0.05)
+    q = quantile(Normal(), 1-alpha/2)
+    p = x.x / x.n
+    σ = q/(2*sqrt(x.n))
```
johnmyleswhite: If you feel up to it, I'd call this `z` since it's neither a standard deviation nor a standard error. Matching the notation from Wikipedia (and maybe even linking to the article in a comment) would help people understand this code if they read it at a future date and want to understand how it works: https://en.wikipedia.org/wiki/Binomial_proportion_confidence_interval#Arcsine_transformation

ararslan:
Actually, I just realized: Could you add a test or two for this method?

src/binomial.jl
DCLukas: Fair point

DCLukas:
Working on mobile has proven to be rather difficult for the tests. Will fix later when on workstation.

DCLukas:
There we are.

ararslan:
Looking good. I'll merge this within the next few days barring further comments from others.

cjprybol:
Looks good! Bumping @ararslan on the merge

ararslan:
Sorry @DCLukas, this fell off my radar. Thanks for the reminder, @cjprybol.

Issue #16, opened 2014-06-15
List of tests common to time series analysis


milktrader:
- [x] Augmented Dickey-Fuller (unit root)
- [ ] Phillips-Perron (unit root)
- [ ] KPSS (unit root)
- [ ] Shapiro-Wilks (normality)
- [x] Jaques-Berra (3rd and 4th moments for normality) (TimeModels has this implemented)
- [x] Ljung-Box (auto correlation)
- [x] Durbin-Watson (auto correlation)
- [x] Breusch-Godfrey (serial correlation)

rawls238:
I'm going to work on implementing these - has any work been done on this?

tomaskrehlik:
Not really.

milktrader:
@rawls238 cool, go for it!

BenjaminBorn:
I have started implementing autocorrelation tests in #87.

BenjaminBorn:
Ok, ADF test is here #92. Please feel free to comment.

@milktrader Should we move the Jarque-Berra test to HypothesisTests as well? It seems a bit lost in TimeModels if all other tests are here.

milktrader:
@BenjaminBorn yes, I think so. TimeModels has not been developed for some time now.

ararslan:
> TimeModels has not been developed for some time now.

In particular, I don't think it even works on Julia versions later than 0.4. If it won't be maintained going forward, perhaps it should be deprecated and the relevant pieces moved to other packages as appropriate? @BenjaminBorn has already been doing excellent work implementing the tests in HypothesisTests; it would just be a matter of finding appropriate homes for the other things.

milktrader:
I agree @ararslan. I could move the package back to my repos and let it sit there until I have resources to develop it.

BenjaminBorn:
I have added the Jarque-Bera test to HypothesisTests (see #95).

Concerning TimeModels, it seems it consists of two main parts. ARIMA modelling based on a state-space approach and some GARCH functionally (transferred from [GARCH.jl](https://github.com/AndreyKolev/GARCH.jl)). Would it make sense to split it again? There are also ARIMA.jl and RARIMA.jl (an R wrapper) but they also seem to be dormant.

milktrader:
Yes, I think a big reason for dormant activity is because it's become too unwieldy.

BenjaminBorn:
Durbin-Watson test added in #102.

azev77:
[...]

PR #87, opened 2017-03-14
Add Box-Pierce and Ljung-Box Autocorrelation Tests

BenjaminBorn:
This is a start to address the lack of time series tests in Julia (see list in https://github.com/JuliaStats/HypothesisTests.jl/issues/16).

I started with easy tests to learn the syntax of `HypothesisTests` (and github in general) but plan to eventually also implement e.g. unit root tests.

Please let me know how I can improve this pull request. I will also add tests and documentation once I have figured out how to do it.


src/box_test.jl
```diff
@@ -0,0 +1,87 @@
...
+    Q::Float64          # test statistic
+end
+
+function BoxPierceTest{T<:Real}(y::AbstractVector{T}, lag::Int, dof::Int=0)
+    if dof>=lag
+        throw(ArgumentError("Number of lags has to be larger than d-o-f correction"))
+    end
+	n = size(y)[1]
+	Q = n * mapreduce(k -> autocor(y,k:k).^2, +, 1:lag)[1]
+	BoxPierceTest(y,n,lag,dof,Q)
```
ararslan: These should be 4-space indents rather than tabs. (Same applies further down)

src/box_test.jl
```diff
@@ -0,0 +1,87 @@
...
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+# LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+# OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+# WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
```
ararslan: Ultimately I'd like to move the individual licenses out of the files in favor of the single primary MIT license, copyright the contributors collectively. Not something that you have to deal with in this PR, I'm just thinking aloud.

src/box_test.jl
```diff
@@ -0,0 +1,87 @@
...
+    y::Vector{Float64}  # time series vector
+    n::Int              # number of observations
+    lag::Int            # number of lags in test statistic
+    dof::Int            # degrees-of-freedom correction (if y is residual)
+    Q::Float64          # test statistic
+end
+
+function BoxPierceTest{T<:Real}(y::AbstractVector{T}, lag::Int, dof::Int=0)
+    if dof>=lag
+        throw(ArgumentError("Number of lags has to be larger than d-o-f correction"))
```
ararslan: I would spell out "degrees of freedom" here

src/box_test.jl
```diff
@@ -0,0 +1,87 @@
...
+    end
+	n = size(y)[1]
```
ararslan: `size(y, 1)`

src/box_test.jl
```diff
@@ -0,0 +1,87 @@
...
+	Q = n * mapreduce(k -> autocor(y,k:k).^2, +, 1:lag)[1]
```
ararslan: `n * first(sum(k -> autocor(y, k:k).^2, 1:lag))`

`first` is equivalent to `[1]` but I usually prefer it for its generality. Also, the `mapreduce` can just be `sum` with a function argument, since `sum(f, x)` just calls `mapreduce(f, +, x)`.

ararslan:
Tests and docs should be pretty simple: tests go in a file in the test directory in the same manner as the other tests, and docs go inline in docstrings. I can answer any more specific questions should you have them.

src/box_test.jl
ararslan: and elsewhere

src/box_test.jl
```diff
@@ -0,0 +1,87 @@
...
+    if dof>=lag
+        throw(ArgumentError("Number of lags has to be larger than d-o-f correction"))
+    end
+	n = size(y)[1]
+	Q = n*(n+2)* mapreduce(k -> autocor(y,k:k).^2/(n-k), +, 1:lag)[1]
+	LjungBoxTest(y,n,lag,dof,Q)
+end
+
+testname(::LjungBoxTest) = "Ljung-Box autocorrelation test"
+population_param_of_interest(x::LjungBoxTest) = ("not applicable", NaN, NaN)
```
ararslan: Why are these "not applicable" and `NaN`s? Assuming I'm remembering my time series courses correctly, the parameter of interest is the *Q* statistic, which is some measure of serial correlation, the value under H<sub>0</sub> is 0, and the observed value is the computed value of *Q*. Right?

BenjaminBorn: Maybe it's semantics but for me Q is the test statistic with a certain distribution under the null, not a parameter. I have never seen the null stated as Q=0 vs. Q>0.

But from a practical point of view I can certainly change it.

ararslan: Hm, that makes sense. So maybe what you have here now is better. I'd be interest to hear from @milktrader, supreme overlord of JuliaStats time series analysis.

BenjaminBorn:
Thanks for the super quick and helpful review. I'll fix the open issues and see that I add some tests and documentation.

ararslan:
Sounds great. Thanks for your help here!

src/box_test.jl
```diff
@@ -36,21 +36,21 @@ end
 
 function BoxPierceTest{T<:Real}(y::AbstractVector{T}, lag::Int, dof::Int=0)
     if dof>=lag
-        throw(ArgumentError("Number of lags has to be larger than d-o-f correction"))
+        throw(ArgumentError("Number of lags has to be larger than degrees-of-freedom correction"))
```
ararslan: No hyphens between "degrees of freedom," please

BenjaminBorn:
I have now added docstrings for the two tests. As there was no general template for what should go into the docstrings for tests in this package, I just wrote what I think is important information. Please let me know if you have a certain format in mind.

ararslan:
I really like those docstrings; they're simple yet informative enough to be immediately useful. That combination is tough to get right sometimes, so nice work. 🙂 

This is looking good to me, save for the .gitignore. The macOS-specific .DS_Store files should be in a global .gitignore rather than being specified in every repository, and you seem to be ignoring a legitimate file. I think just removing those changes should be sufficient.

src/box_test.jl
ararslan: Also cc @andreasnoack for this

BenjaminBorn:
Ok, I made a big mess trying to get rid of the .gitignore. I have now repackaged everything in one commit. Hope it is ok this way.

ararslan:
Yep, that's totally fine. I figured I'd squash the commits on merge anyway, so no worries.

src/box_test.jl
BenjaminBorn: It might be helpful to think of a regression where you want to test a number of joint hypotheses using an F- (or Wald) statistic. You would not write H_0: F=0.

ararslan: Right, but in that case you'd still have a parameter you're trying to estimate, even if it's a linear combination of treatment means, for example. I still think providing more meaningful values from this function would be useful in this case, I'm just not sure how best to specify them.

BenjaminBorn: Could we call it "statistic of interest" instead of "parameter of interest"?

BenjaminBorn: This problem is not unique to these tests so it might be worthwhile to allow this option.

ararslan: Mmm, maybe, though saying "parameter" is more informative for most other tests.

BenjaminBorn: Oh I agree. I was thinking that there could be the option of printing one or the other.

BenjaminBorn: I had a look at the Kruskal Wallis test and there it says
`population_param_of_interest(x::KruskalWallisTest) = ("Location parameters", "all equal", NaN)`

Maybe a compromise here could be
`population_param_of_interest(x::BoxPierceTest) = ("Autocorrelations up to lag k", "all zero", NaN)`

The value of the Q statistic is then printed in the details a few lines down.

ararslan: That seems good to me 👍

BenjaminBorn: Done :-)

src/box_test.jl
```diff
@@ -60,7 +60,8 @@ function BoxPierceTest{T<:Real}(y::AbstractVector{T}, lag::Int, dof::Int=0)
 end
 
 testname(::BoxPierceTest) = "Box-Pierce autocorrelation test"
-population_param_of_interest(x::BoxPierceTest) = ("not applicable", NaN, NaN)
+population_param_of_interest(x::BoxPierceTest) = ("autocorrelations up to lag k",
+ "all zero", NaN)
```
ararslan: Super pedantic, but it'd be nice to indent this by 4 spaces rather than 1

BenjaminBorn: No problem. Better?

src/box_test.jl
```diff
@@ -0,0 +1,121 @@
...
+model, set `dof=p+q`.
+
+External links
+
+* [Box-Pierce test on Wikipedia](https://en.wikipedia.org/wiki/Ljung–Box_test#Box-Pierce_test)
+"""
+function BoxPierceTest{T<:Real}(y::AbstractVector{T}, lag::Int, dof::Int=0)
+    if dof>=lag
+        throw(ArgumentError("Number of lags has to be larger than degrees of
+        freedom correction"))
```
ararslan: The extra leading whitespace here isn't stripped out, so what I would do is:
```julia
if dof >= lag
    throw(ArgumentError("Number of lags has to be larger than degrees of" *
                        "freedom correction"))
end
```

test/box_test.jl
```diff
@@ -0,0 +1,41 @@
...
...;-0.0667679865065892;1.22246184846806;0.567695459738444;0.763498928323387;0.378598879853350;-0.645597056636597;-0.664646244391031;-1.80302632421089;-1.28904760384974;-0.335270079518118;0.0704676013611580;0.341793767379823;1.73516566605438;1.29991596431019;0.206364081402617;-1.00885995553760;-0.850055670315398;1.12940943588433;-1.34853345649183;1.24029090042906;-0.0556470932073243;0.759390213945460;-0.0302668866467625;-0.361082667756094;-2.00730320253256;-0.568859127346638;-1.14962752439700;1.88062217463119;-1.47062454138738;-0.535210970165732;-0.963543959825501;-1.38510806767591;0.134300237561270;-0.616116666008819;-1.71998935631974;0.320769279987411;-1.44737289788151;-0.492271188069566;-0.413087383983183;1.00192851245595;1.17468132597511;-1.32697034506205;-1.59321486804407;-0.438709975758797;1.22579749799787;1.48909027806961;2.15943318710085;-0.909479018037503;1.14782030061250;-0.505072196522373;-0.267067035218130;1.49908239393073;0.797304160175390;-0.171640443307097;-0.469079609930096;0.217624021208600;0.359146407602350;0.320024715708058;0.259216328055686;0.459695985041628;0.209986360042143;1.52317539117726;1.25979460361220;-0.165283101114261;0.156194494887636;-0.814315294376699]
```
ararslan: I just noticed how incredibly long this line is. I'd probably break and align after every fourth or fifth semicolon (which could equivalently be commas, FYI--they construct the same thing). Typically we like to stick to the code style used in Base Julia, which sets a cap on line lengths at 92 characters.

A common style in this scenario is to do
```julia
sim_data_h0 = [
    0.297287984535462, 0.382395967790608, -0.597634476728231, -0.0104452446373756,
    -0.839026854388764, 0.311111338498334, 2.29508782383731, -2.26708634880053,
    ...
]
```

src/box_test.jl
ararslan: Yes indeed, thanks!

ararslan:
Nice, this looks good to me now! Out of curiosity, did you get the values used for the test from R? I might be good to check against a value computed by a third party as a sanity check, if it isn't already what you've done here.

BenjaminBorn:
I cross-checked the Ljung-Box test with Matlab's lbqtest routine. Matlab gives exactly the same values. Box-Pierce is not implemented in Matlab but it's a trivial simplification of Ljung-Box so I thought I'm safe. But I could cross-check with R.

PR #90, opened 2017-03-29
No qualitative significance output in pretty-print

alexmorley:
Just personal but wondering whether anyone else might agree that pushing the whole dividing p-values into discrete bands of significance is not ideal.


ararslan:
Thanks. I really thought we shouldn't be doing this but I never pursued it enough to submit a PR. Significance should depend on the problem at hand.

PR #92, opened 2017-04-08
Add Augmented Dickey-Fuller unit root test

BenjaminBorn:
Please have a look at line 68. I need to ensure that `xt` is a two dimensional array in order to multiply it in line 94. For the case of no constant and no lags, `xt` is just a vector so I transpose it twice. Is there a more elegant way?

The test statistic is for all cases exactly the same as in R (urca package) and GRETL. Critical values and p-values are slightly different due to different approaches. One could refine the computation of p-values but the algorithms used by, e.g., R are GPL so one would have to implement them from the underlying research paper (see references in code). Given that they are all approximations, I propose we go with the method used here for now.


src/augmented_dickey_fuller.jl
```diff
@@ -0,0 +1,219 @@
...
+  * [Augmented Dickey-Fuller test on Wikipedia](https://en.wikipedia.org/wiki/Augmented_Dickey–Fuller_test)
+"""
+function ADFTest{T<:Real}(y::AbstractVector{T}, deterministic::AbstractString, lag::Int)
+
+    nobs = length(y)
+    Δy = diff(y)
+
+    if deterministic == "none"
+        Δyt   = Δy[lag+1:end]
+        xt    = (y[lag+1:end-1]')' # to make sure that xt is always two dimensional Array
```
andreasnoack: This won't make it a 2d array on 0.6. Use `y[lag+1:end-1,1:1]`.

src/augmented_dickey_fuller.jl
```diff
@@ -0,0 +1,219 @@
...
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+# LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+# OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+# WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+export ADFTest
+
+immutable ADFTest <: HypothesisTest
+    n               ::Int               # number of observations
+    deterministic   ::AbstractString    # deterministic terms included in regression
```
andreasnoack: Use `String` to abstract types as fields.

src/augmented_dickey_fuller.jl
```diff
@@ -0,0 +1,219 @@
...
+        Δylag = hcat([Δy[lag-i+1:end-i] for i=1:lag]...)
+        xt = [ xt Δylag ]
+    end
+
+    adf_coef = xt \ Δyt
+    adf_res  = Δyt - xt * adf_coef
+    n        = length(adf_res)
+    k        = lag + 1 + dfexo
+    sigma2   = dot(adf_res, adf_res) / (n - k)
+    cov_coef = sigma2 * inv(xt' * xt)
```
andreasnoack: You only use element `1,1` so you can avoid a temporary array by postponing the multiplication of `sigma2` to next line.

src/augmented_dickey_fuller.jl
ararslan: I'd actually use `Symbol` in this case. I think that's more common for situations like this.

BenjaminBorn: I have changed it to `String` for now but I'm happy to change it to `Symbol`. 

Tbh, even after reading Stefan's explanation [here](http://stackoverflow.com/a/23482257) I'm not sure I totally understand symbols.

ararslan: It would be great if you could use symbols, I think that's the better choice here. Stefan's explanation is good but deeper than one really needs to know to use them for everyday tasks. `Symbol` is an interesting type in that it plays a central role in Julia's AST, which is made up of `Expr`s and `Symbol`s, but `Symbol`s can be used more generally; for example, we use them in this package to specify the types for confidence intervals and in DataFrames we use them for column names. You can think of them sort of like strings that cannot be decomposed further. (Strings can be decomposed into `Char`s.)

BenjaminBorn: Thanks for the explanation and your patience. I learn something with every commit ;-) The code now uses symbols.

ararslan: Great, thanks!

BenjaminBorn:
Bump.

PR #93, opened 2017-04-10
Allow for one-sided testing in printing of pvalue

BenjaminBorn:
Prints "one-sided" or "two-sided" instead of the hard coded "two-sided" depending on the test.

### DIFF
--- a/src/HypothesisTests.jl
+++ b/src/HypothesisTests.jl
@@ -40,27 +40,30 @@ check_same_length(x::AbstractVector, y::AbstractVector) = if length(x) != length
 end
 
 # Basic function for finding a p-value given a distribution and tail
-pvalue(dist::ContinuousUnivariateDistribution, x::Number; tail=:both) =
+function pvalue(dist::ContinuousUnivariateDistribution, x::Number; tail=:both)
     if tail == :both
-        min(2 * min(cdf(dist, x), ccdf(dist, x)), 1.0)
+        p = min(2 * min(cdf(dist, x), ccdf(dist, x)), 1.0)
     elseif tail == :left
-        cdf(dist, x)
+        p = cdf(dist, x)
     elseif tail == :right
-        ccdf(dist, x)
+        p = ccdf(dist, x)
     else
         throw(ArgumentError("tail=$(tail) is invalid"))
     end
-
-pvalue(dist::DiscreteUnivariateDistribution, x::Number; tail=:both) =
+    return p, tail
+end
+function pvalue(dist::DiscreteUnivariateDistribution, x::Number; tail=:both)
     if tail == :both
-        min(2 * min(ccdf(dist, x-1), cdf(dist, x)), 1.0)
+        p = min(2 * min(ccdf(dist, x-1), cdf(dist, x)), 1.0)
     elseif tail == :left
-        cdf(dist, x)
+        p = cdf(dist, x)
     elseif tail == :right
-        ccdf(dist, x-1)
+        p = ccdf(dist, x-1)
     else
         throw(ArgumentError("tail=$(tail) is invalid"))
     end
+    return p, tail
+end
 
 function check_alpha(alpha::Float64)
     if alpha <= 0 || alpha >= 0.5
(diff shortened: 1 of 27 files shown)


ararslan:
This quite a large breaking change, since `pvalue` returns a tuple rather than a scalar. I don't think that's what we want to do. Perhaps we should make an internal function `_pvalue` that returns the tuple, then define `pvalue(...) = first(_pvalue(...))`. Then `_pvalue` can be used for printing methods to get the tail-specific behavior, but no user code breaks.

simonster:
Or have a method that returns the default sidedness of the p-value for each test, much like `population_param_of_interest`.

ararslan:
@simonster That's nice idea, though it wouldn't capture the value of `tail` passed to `pvalue`, right?

simonster:
The `show` method here applies to a `HypothesisTest` object, so there's no possibility to specify the tail. The only way to get a p-value with a non-default tail is to call `pvalue(; tail=:whatever)`, and that just returns the number.

BenjaminBorn:
Thanks a lot for the input. Would it also be ok to just add a keyword to `pvalue(...; printtail=0)` so that by default it only returns the value but if I set the keyword to one, it returns the tuple? Or is it a problem that the return of the function changes depending on the keyword?

ararslan:
> The show method here applies to a HypothesisTest object, so there's no possibility to specify the tail.

Ah, you're right. I didn't think about that.

>  Or is it a problem that the return of the function changes depending on the keyword?

Yeah, unfortunately that's not a good way to go, since it introduces type instability in the output.

Simon's default sidedness seems reasonable to me given the limitations imposed by `show`.

BenjaminBorn:
Ok, I have now implemented Simon's approach and updated all existing tests. Thanks again for the feedback.

src/HypothesisTests.jl
```diff
@@ -89,14 +86,17 @@ function Base.show{T<:HypothesisTest}(io::IO, test::T)
...
+    tail = used_tail_of_distribution(test)
     println(io, "Test summary:")
     println(io, "    outcome with 95% confidence: $outcome h_0")
     if tail == :both
         println(io, "    two-sided p-value:           $p")
-    else
+    elseif tail == :left || tail == :right
         println(io, "    one-sided p-value:           $p")
+    else
+        println(io, "    ??-sided p-value:            $p")
```
ararslan: I think in this case I would just say `p-value`

BenjaminBorn: I have changed it to just say `p-value` if a test does not set `tail_of_distribution()`. Btw, feel free to propose a different name for that function.

All tests pass locally. Let met know if I should do anything else. Thanks!

src/HypothesisTests.jl
```diff
@@ -101,6 +108,9 @@ end
 # parameter of interest: name, value under h0, point estimate
 population_param_of_interest{T<:HypothesisTest}(test::T) = ("not implemented yet", NaN, NaN)
 
+# is the test one- or two-sided
+tail_of_distribution{T<:HypothesisTest}(test::T) = :undefined
```
ararslan: It doesn't really matter, but just as a general heads up, in cases like this the typevar isn't necessary; this could just be `tail_of_distribution(test::HypothesisTest)`. Not necessary to change it, just an FYI.

ararslan:
Maybe instead of `tail_of_distribution` I'd call it `default_tail`. It would also be good to have some tests for this functionality. Thanks again for your hard work!

src/HypothesisTests.jl
BenjaminBorn: That's a good point. I just followed the definition of `population_param_of_interest`. Is it because there is only one function argument?

I will change the definition together with the function name.

BenjaminBorn:
I have changed the function name and added some tests. I had to export `default_tail` in order to use it for the unit tests. Is that ok?

ararslan:
Since we don't export things like `population_param_of_interest` I'd say we don't need to export `default_tail`. To use it in the unit tests without qualifying the name, you can just do
```julia
using HypothesisTests: default_tail
```
separately from the primary `using HypothesisTests`.

BenjaminBorn:
That's what I was looking for ;-) Agree, that we don't want to export it. Changed the code accordingly.

ararslan:
Thanks, looks good to me. What do you think, @simonster?

PR #95, opened 2017-04-29
Add Jarque-Bera normality test

BenjaminBorn:
Moves the Jarque-Bera test for normality from TimeModels as discussed in #16. Results were verified with R's tseries package.


src/jarque_bera.jl
```diff
@@ -0,0 +1,69 @@
...
+# OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+# WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+export JarqueBeraTest
+
+immutable JarqueBeraTest <: HypothesisTest
+    n       ::Int           # number of observations
+    JB      ::Float64       # test statistic
+    m3      ::Float64       # skewness
+    m4      ::Float64       # excess kurtosis
```
ararslan: Kind of an oddly large amount of space here between the variable names and the type specifications. The more typical style is to omit any space, so each is `x::T`.

src/jarque_bera.jl
```diff
@@ -0,0 +1,69 @@
...
+
+Compute the Jarque-Bera statistic to test the null hypothesis that a series `y` comes from a normal distribution.
+
+Note that the p-values might be distorted in small samples.
+
+External links
+
+* [Jarque-Bera test on Wikipedia](https://en.wikipedia.org/wiki/Jarque–Bera_test)
+"""
+
```
ararslan: No line break between doc strings and function declarations

src/jarque_bera.jl
```diff
@@ -0,0 +1,69 @@
...
+end
+
+"""
+    JarqueBeraTest(y)
+
+Compute the Jarque-Bera statistic to test the null hypothesis that a series `y` comes from a normal distribution.
+
+Note that the p-values might be distorted in small samples.
+
+External links
```
ararslan: Should have a markdown header hash here: `# External links`. Also I'd say "is normally distributed" rather than "comes from a normal distribution," I'd add the type specification to the signature (`JarqueBeraTest(y::AbstractVector)`), and I'd say "a real-valued vector" rather than "a series" (since there are no series objects in this package).

Also, can you be more specific about the p-value being distorted? Is it inflated? Too small?

src/jarque_bera.jl
```diff
@@ -0,0 +1,69 @@
...
+
+* [Jarque-Bera test on Wikipedia](https://en.wikipedia.org/wiki/Jarque–Bera_test)
+"""
+
+function JarqueBeraTest{T<:Real}(y::AbstractVector{T})
+    n = length(y)
+    m1 = sum(y)/n
+    m2 = sum((y - m1).^2)/n
+    m3 = sum((y - m1).^3)/n
+    m4 = sum((y - m1).^4)/n
```
ararslan: You can simplify this to make a single pass over the input:
```julia
n = length(y)
S = Base.promote_op(/, T, typeof(n))
m1 = m2 = m3 = m4 = zero(S)
@inbounds for yi in y
    m1 += yi / n
    m2 += yi^2 / n
    m3 += yi^3 / n
    m4 += yi^4 / n
end
```
This is 4n divisions, which is not great, but it's better than possibly overflowing if we wait until the end to divide. It should still be better than multiple passes over the input, at any rate.

src/jarque_bera.jl
```diff
@@ -0,0 +1,69 @@
...
+    b1 = (m3/m2^(3/2))^2
+    b2 = (m4/m2^2)
```
ararslan:
```julia
b1 = m3^2 / m2^3
b2 = m4 / m2^2
```
is equivalent but easier to read

test/jarque_bera.jl
```diff
@@ -0,0 +1,58 @@
+using HypothesisTests, Base.Test
```
ararslan: This file will needed to be `include`d in test/runtests.jl. It actually appears that a number of files are missing from that, so we have a bunch of stuff that isn't being tested. 😑

src/jarque_bera.jl
BenjaminBorn: I took the vertical alignment advice from here: [https://github.com/JuliaPraxis/Spacing/blob/master/guides/VerticalSpacing.md](https://github.com/JuliaPraxis/Spacing/blob/master/guides/VerticalSpacing.md) but agree that it was excessive.

src/jarque_bera.jl
BenjaminBorn: That's a good idea. One complication is that we need the central moments (i.e. in deviations from the mean). I now transform raw moments to central moments after the loop.

src/jarque_bera.jl
BenjaminBorn: Changed it slightly differently so that I have skewness and kurtosis as variables that I can use in the test output.

test/jarque_bera.jl
BenjaminBorn: Sorry, that was my mistake. I have now added all the missing tests to the file. Tests still pass :-)

ararslan: Awesome, thank you!

src/jarque_bera.jl
ararslan: The JuliaPraxis guidelines aren't official and aren't used in the JuliaStats packages. We tend to follow the style used in Base, wherein there's no space at all between the variable, `::`, and type name.

BenjaminBorn: Ok, changed it.

ararslan: Cool, thank you!

src/jarque_bera.jl
```diff
@@ -0,0 +1,87 @@
...
+end
+
+"""
+    JarqueBeraTest(y::AbstractVector)
+
+Compute the Jarque-Bera statistic to test the null hypothesis that a real-valued vector `y`
+is normally distributed.
+
+Note that the approximation by the Chi-squared distribution does not work well and the
+speed of convergence is slow. In small samples, the tests tends to be over-sized for
```
ararslan: "tests" -> "test"

BenjaminBorn: Fixed. Thank you so much for your help.

ararslan: And thank you for your consistently fantastic contributions!

PR #99, opened 2017-06-02
Subtypes of HypothesisTest can have optional fields :tail and :alpha

mirkobunse:
This enables the specification of tail (left/right/both) and alpha value in the constructors of HypothesisTest subtypes:

```julia
using HypothesisTests
EqualVarianceTTest(x, y, alpha=0.0025, tail=:right)
```

Since the additional arguments of the constructors are named arguments, nobody using this package has to change his constructor calls. Defaults are used so that the package works just as before when the optional arguments are not provided.

Currently, the new optional functionality of this pull request is only implemented for the t-tests. Other subtypes of HypothesisTest can easily follow the adaptations of `src/t.jl`.

I am looking forward to your comments!
Mirko

PR #100, opened 2017-06-04
Subtypes of HypothesisTest can have optional fields :tail and :alpha

mirkobunse:
This enables the specification of tail (left/right/both) and alpha value in the constructors of HypothesisTest subtypes:

```julia
using HypothesisTests
EqualVarianceTTest(x, y, alpha=0.0025, tail=:right)
```

Since the additional arguments of the constructors are named arguments, nobody using this package has to change his constructor calls. Defaults are used so that the package works just as before when the optional arguments are not provided.

Currently, the new optional functionality of this pull request is only implemented for the t-tests. Other subtypes of HypothesisTest can easily follow the adaptations of `src/t.jl`.

I am looking forward to your comments!
Mirko

---

Note that this pull request was already present as [pull request #99](https://github.com/JuliaStats/HypothesisTests.jl/pull/99). Since there was a problem with the Travis build routine, I reopen that pull request here.

### DIFF
--- a/src/HypothesisTests.jl
+++ b/src/HypothesisTests.jl
@@ -68,21 +68,16 @@ function check_alpha(alpha::Float64)
     end
 end
 
-function get_tail(test::HypothesisTest)
-    if :tail in fieldnames(test)
-        getfield(test, :tail)
+# Utility to get an optional field (e.g., :tail and :alpha)
+getfield(value::HypothesisTest, name::Symbol, default::Any) =
+    if name in fieldnames(value)
+        Base.getfield(value, name)
     else
-        default_tail(test)
+        default
     end
-end
 
-function get_alpha(test::HypothesisTest)
-    if :alpha in fieldnames(test)
-        getfield(test, :alpha)
-    else
-        0.05
-    end
-end
+get_tail(test::HypothesisTest) = getfield(test, :tail, default_tail(test))
+get_alpha(test::HypothesisTest) = getfield(test, :alpha, 0.05)
 
 # Utility for pretty-printing: Append white space so that length(with_trailing_whitespace(s)) = max(len, length(s))
 with_trailing_whitespace(s::String, len::Int) = s * join(repeat([" "], outer=max(len - length(s), 0)), "")
--- a/src/HypothesisTests.jl
+++ b/src/HypothesisTests.jl
@@ -76,10 +76,20 @@ function get_tail(test::HypothesisTest)
     end
 end
 
+function get_alpha(test::HypothesisTest)
+    if :alpha in fieldnames(test)
+        getfield(test, :alpha)
+    else
+        0.05
+    end
+end
+
 # Pretty-print
 function Base.show{T<:HypothesisTest}(io::IO, test::T)
     println(io, testname(test))
     println(io, repeat("-", length(testname(test))))
+    
+    conf_string = string((1 - get_alpha(test)) * 100) * "%"
 
     # population details
     has_ci = applicable(StatsBase.confint, test)
@@ -89,16 +99,16 @@ function Base.show{T<:HypothesisTest}(io::IO, test::T)
     println(io, "    value under h_0:         $param_under_h0")
     println(io, "    point estimate:          $param_estimate")
     if has_ci
-        println(io, "    95% confidence interval: $(StatsBase.confint(test))")
+        println(io, "    $conf_string confidence interval: $(StatsBase.confint(test))") # TODO indent all details like this one
     end
     println(io)
 
     # test summary
     tail = get_tail(test)
     p = pvalue(test, tail=tail)
-    outcome = if p > 0.05 "fail to reject" else "reject" end
+    outcome = if p > get_alpha(test) "fail to reject" else "reject" end
     println(io, "Test summary:")
-    println(io, "    outcome with 95% confidence: $outcome h_0")
+    println(io, "    outcome with $conf_string confidence: $outcome h_0") # TODO indent all details like this one
     if tail == :both
         println(io, "    two-sided p-value:           $p")
     elseif tail == :left || tail == :right
--- a/src/HypothesisTests.jl
+++ b/src/HypothesisTests.jl
@@ -68,6 +68,14 @@ function check_alpha(alpha::Float64)
     end
 end
 
+function get_tail(test::HypothesisTest)
+    if :tail in fieldnames(test)
+        getfield(test, :tail)
+    else
+        default_tail(test)
+    end
+end
+
 # Pretty-print
 function Base.show{T<:HypothesisTest}(io::IO, test::T)
     println(io, testname(test))
@@ -86,9 +94,9 @@ function Base.show{T<:HypothesisTest}(io::IO, test::T)
     println(io)
 
     # test summary
-    p = pvalue(test)
+    tail = get_tail(test)
+    p = pvalue(test, tail=tail)
     outcome = if p > 0.05 "fail to reject" else "reject" end
-    tail = default_tail(test)
     println(io, "Test summary:")
     println(io, "    outcome with 95% confidence: $outcome h_0")
     if tail == :both
--- a/src/HypothesisTests.jl
+++ b/src/HypothesisTests.jl
@@ -84,6 +84,9 @@ function get_alpha(test::HypothesisTest)
     end
 end
 
+# Utility for pretty-printing: Append white space so that length(with_trailing_whitespace(s)) = max(len, length(s))
+with_trailing_whitespace(s::String, len::Int) = s * join(repeat([" "], outer=max(len - length(s), 0)), "")
+
 # Pretty-print
 function Base.show{T<:HypothesisTest}(io::IO, test::T)
     println(io, testname(test))
@@ -93,13 +96,14 @@ function Base.show{T<:HypothesisTest}(io::IO, test::T)
 
     # population details
     has_ci = applicable(StatsBase.confint, test)
+    label_len = has_ci ? length(conf_string) + 21 : 22 # 21 is length of ' confidence interval:', 22 that of 'parameter of interest:'
     (param_name, param_under_h0, param_estimate) = population_param_of_interest(test)
     println(io, "Population details:")
-    println(io, "    parameter of interest:   $param_name")
-    println(io, "    value under h_0:         $param_under_h0")
-    println(io, "    point estimate:          $param_estimate")
+    println(io, "    $(with_trailing_whitespace("parameter of interest:", label_len)) $param_name")
+    println(io, "    $(with_trailing_whitespace("value under h_0", label_len)) $param_under_h0")
+    println(io, "    $(with_trailing_whitespace("point estimate", label_len)) $param_estimate")
     if has_ci
-        println(io, "    $conf_string confidence interval: $(StatsBase.confint(test))") # TODO indent all details like this one
+        println(io, "    $conf_string confidence interval: $(StatsBase.confint(test))")
     end
     println(io)
 
@@ -108,13 +112,14 @@ function Base.show{T<:HypothesisTest}(io::IO, test::T)
     p = pvalue(test, tail=tail)
     outcome = if p > get_alpha(test) "fail to reject" else "reject" end
     println(io, "Test summary:")
-    println(io, "    outcome with $conf_string confidence: $outcome h_0") # TODO indent all details like this one
+    label_len = length(conf_string) + 25 # 25 is length of 'outcome with  confidence:'
+    println(io, "    outcome with $conf_string confidence: $outcome h_0")
     if tail == :both
-        println(io, "    two-sided p-value:           $p")
+        println(io, "    $(with_trailing_whitespace("two-sided p-value:", label_len)) $p")
     elseif tail == :left || tail == :right
-        println(io, "    one-sided p-value:           $p")
+        println(io, "    $(with_trailing_whitespace("one-sided p-value:", label_len)) $p")
     else
-        println(io, "    p-value:                     $p")
+        println(io, "    $(with_trailing_whitespace("p-value:", label_len)) $p")
     end
     println(io)
 
--- a/src/augmented_dickey_fuller.jl
+++ b/src/augmented_dickey_fuller.jl
@@ -207,6 +207,7 @@ end
 testname(::ADFTest) = "Augmented Dickey-Fuller unit root test"
 population_param_of_interest(x::ADFTest) =
     ("coefficient on lagged non-differenced variable", 0, x.coef)
+default_tail(test::ADFTest) = :left
 
 function show_params(io::IO, x::ADFTest, ident)
     println(io, ident, "sample size in regression:          ", x.n)
@@ -215,4 +216,4 @@ function show_params(io::IO, x::ADFTest, ident)
     println(io, ident, "Critical values at 1%, 5%, and 10%: ", x.cv')
 end
 
-pvalue(x::ADFTest) = HypothesisTests.pvalue(Normal(0, 1), adf_pv_aux(x.stat, x.deterministic); tail=:left)
+pvalue(x::ADFTest; tail=default_tail(x)) = HypothesisTests.pvalue(Normal(0, 1), adf_pv_aux(x.stat, x.deterministic); tail=tail)
--- a/src/circular.jl
+++ b/src/circular.jl
@@ -60,7 +60,10 @@ function show_params(io::IO, x::RayleighTest, ident="")
     println(io, ident, "test statistic:         $(x.Rbar^2 * x.n)")
 end
 
-function pvalue(x::RayleighTest)
+function pvalue(x::RayleighTest; tail=default_tail(x))
+    if tail != default_tail(x) # warn that tail is not used here
+        warn("tail=$tail has no effect on the computation of the p value")
+    end
     Z = x.Rbar^2 * x.n
     x.n > 1e6 ? exp(-Z) :
         exp(-Z)*(1+(2*Z-Z^2)/(4*x.n)-(24*Z - 132*Z^2 + 76*Z^3 - 9*Z^4)/(288*x.n^2))
--- a/src/t.jl
+++ b/src/t.jl
@@ -33,13 +33,13 @@ pvalue(x::TTest; tail=:both) = pvalue(TDist(x.df), x.t; tail=tail)
 default_tail(test::TTest) = :both
 
 # confidence interval by inversion
-function StatsBase.confint(x::TTest, alpha::Float64=0.05; tail=:both)
+function StatsBase.confint(x::TTest, alpha::Float64=0.05; tail=get_tail(x))
     check_alpha(alpha)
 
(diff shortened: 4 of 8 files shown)


src/augmented_dickey_fuller.jl
```diff
@@ -215,4 +216,4 @@ function show_params(io::IO, x::ADFTest, ident)
     println(io, ident, "Critical values at 1%, 5%, and 10%: ", x.cv')
 end
 
-pvalue(x::ADFTest) = HypothesisTests.pvalue(Normal(0, 1), adf_pv_aux(x.stat, x.deterministic); tail=:left)
+pvalue(x::ADFTest; tail=default_tail(x)) = HypothesisTests.pvalue(Normal(0, 1), adf_pv_aux(x.stat, x.deterministic); tail=tail)
```
nalimilan: I don't understand why you don't use `tail=:left` here, since you already know the result of `default_tail` for that test. Same below. Also, 92 char limit.

src/circular.jl
```diff
@@ -60,7 +60,10 @@ function show_params(io::IO, x::RayleighTest, ident="")
     println(io, ident, "test statistic:         $(x.Rbar^2 * x.n)")
 end
 
-function pvalue(x::RayleighTest)
+function pvalue(x::RayleighTest; tail=default_tail(x))
+    if tail != default_tail(x) # warn that tail is not used here
+        warn("tail=$tail has no effect on the computation of the p value")
```
nalimilan: Better throw an error if the argument doesn't work.

src/t.jl
```diff
@@ -33,13 +33,13 @@ pvalue(x::TTest; tail=:both) = pvalue(TDist(x.df), x.t; tail=tail)
 default_tail(test::TTest) = :both
 
 # confidence interval by inversion
-function StatsBase.confint(x::TTest, alpha::Float64=0.05; tail=:both)
+function StatsBase.confint(x::TTest, alpha::Float64=get_alpha(x); tail=get_tail(x))
     check_alpha(alpha)
 
     if tail == :left
-        (-Inf, StatsBase.confint(x, alpha*2)[2])
+        (-Inf, StatsBase.confint(x, alpha*2, tail=:both)[2]) # tail=:both required as recursive anchor
```
nalimilan: "recursive anchor"? I don't understand what happens here.

mirkobunse:
Thank you very much, @nalimilan, for your detailed review! I implemented all changes you requested and even learned about some other nice features of Julia during the process. See my comments for more information and please let me know, if there is anything else to change.

src/HypothesisTests.jl
```diff
@@ -68,35 +68,54 @@ function check_alpha(alpha::Float64)
     end
 end
 
+# Utility to get an optional field (e.g., :tail and :alpha)
+getfield(value::HypothesisTest, name::Symbol, default::Any) =
```
mirkobunse: Removing the "get_" prefix, I simply called the functions `tail` and `alpha`. By default, the tail is `:undefined`, which is the value returned by `default_tail` if not overloaded.

I feel that the method `default_tail` is kinda superseded by this implementation of the `tail` method. For all tests that always use their default tail, the `tail` method returns that default, just like `default_tail` does. For t-tests, which can be configured to use another tail, that other tail is returned (or the default, if no other tail is specified).

`default_tail` can still be used (and behaves just like `tail`) but it emits a deprecation warning.

src/HypothesisTests.jl
```diff
@@ -68,35 +68,54 @@ function check_alpha(alpha::Float64)
...
     println(io, testname(test))
     println(io, repeat("-", length(testname(test))))
+    
+    alpha = get_alpha(test)
+    tail  = get_tail(test)
+    conf_string = string((1 - alpha) * 100) * "%"
 
     # population details
     has_ci = applicable(StatsBase.confint, test)
+    label_len = has_ci ? length(conf_string) + 21 : 22 # 21 is length of ' confidence interval:', 22 that of 'parameter of interest:'
```
mirkobunse: There is a function `prettify_detail` inside `Base.show` now that aligns each line consisting of label and value to the longest possible length.

src/t.jl
mirkobunse: The quantiles are only computed if `tail==:both`. The recursion would be infinite if the argument `tail=:both` would not be provided because `tail` is set to the tail of `x` now, which may be one-sided.

The term "recursive anchor" is chosen because `tail==:both` will not allow infinite recursion.

src/HypothesisTests.jl
mirkobunse: Removing the "get_" prefix, I simply called the functions `tail` and `alpha`. The tail is `:undefined`, which is the value returned by `default_tail` if that method is not overloaded. For all tests that always use some default tail, the `tail` method returns that tail, just like `default_tail` does. For t-tests, which can be configured to use another tail, that other tail is returned (or the default, if no other tail is specified).

I feel that the method `default_tail` is superseded by this implementation of the `tail` method.
`default_tail` can still be used (and behaves just like `tail`) but it emits a deprecation warning.

src/t.jl
mirkobunse: The quantiles are only computed if `tail==:both`. The recursion would be infinite if the argument `tail=:both` would not be provided because `tail` is set to the tail of `x` now, which may be one-sided.

The term "recursive anchor" is chosen because `tail==:both` will not allow infinite recursion.

src/HypothesisTests.jl
mirkobunse: There is a function `prettify_detail` inside `Base.show` now that aligns each line consisting of label and value to the longest possible length.

mirkobunse:
Please note that the Travis build failed for `julia:nightly` because the package `Rmath` (and thus also `StatsFuns` and `Distributions`, on which this package depends) could not be precompiled.

I hope this is no problem for this pull request. The REQUIRE file states that only julia 0.5 is meant to be supported.

src/augmented_dickey_fuller.jl
mirkobunse: I reverted this change by removing the named tail parameter for all but the t-tests

doc/api/confint.rst
```diff
@@ -3,11 +3,12 @@
 Confidence Interval
 ==============================================
 
-.. function:: confint(test::HypothesisTest, alpha=0.05; tail=:both)
+.. function:: confint(test::HypothesisTest, alpha=alpha(test); tail=tail(test))
 
-    Compute a confidence interval C with coverage 1-``alpha``.
+    Compute a confidence interval C with coverage 1-``alpha``
+    (``alpha=0.05`` is the default for all tests).
```
nalimilan: 0.05 is the default for all tests *unless overridden when creating the object*.

doc/parametric/test_t.rst
```diff
@@ -54,5 +66,8 @@ T-test
     .. math::
         \nu_{\chi'} \approx \frac{\left(\sum_{i=1}^n k_i s_i^2\right)^2}
                                  {\sum_{i=1}^n \frac{(k_i s_i^2)^2}{\nu_i}}
-
+
```
nalimilan: Remove trailing spaces here and elsewhere.

src/HypothesisTests.jl
```diff
@@ -72,32 +72,35 @@ end
 function Base.show{T<:HypothesisTest}(io::IO, test::T)
     println(io, testname(test))
     println(io, repeat("-", length(testname(test))))
+    
+    # utilities for pretty-printing
+    conf_string = string(floor((1 - alpha(test)) * 100, 6)) # limit to 6 decimals in %
+    prettify_detail(label::String, value::Any, len::Int) = # len is max length of label
```
nalimilan: I'd say "format" rather than "prettify".

src/HypothesisTests.jl
```diff
@@ -72,32 +72,35 @@ end
...
     println(io, "Population details:")
-    println(io, "    parameter of interest:   $param_name")
-    println(io, "    value under h_0:         $param_under_h0")
-    println(io, "    point estimate:          $param_estimate")
+    println(io, prettify_detail("parameter of interest:", param_name, 32))
+    println(io, prettify_detail("value under h_0:", param_under_h0, 32))
+    println(io, prettify_detail("point estimate:", param_estimate, 32))
     if has_ci
-        println(io, "    95% confidence interval: $(StatsBase.confint(test))")
+        println(io, prettify_detail(conf_string*"% confidence interval:", StatsBase.confint(test), 32))
```
nalimilan: No need for `StatsBase.`, right?

test/anderson_darling.jl
```diff
@@ -9,7 +9,8 @@ x = rand(Normal(), n)
 t = OneSampleADTest(x, Normal())
 @test isapprox(t.A², 0.2013, atol=0.1^4)
 @test isapprox(pvalue(t), 0.8811, atol=0.1^4)
-@test default_tail(t) == :right
+@test tail(t) == :right
+# @test HypothesisTests.default_tail(t) == :right
```
nalimilan: Remove these comments.

doc/api/confint.rst
```diff
@@ -3,11 +3,12 @@
 Confidence Interval
 ==============================================
 
-.. function:: confint(test::HypothesisTest, alpha=0.05; tail=:both)
+.. function:: confint(test::HypothesisTest, alpha=alpha(test); tail=tail(test))
```
nalimilan: `alpha` and `tail` should be exported if they are shown here.

doc/api/confint.rst
```diff
@@ -3,11 +3,12 @@
...
 
-    Compute a confidence interval C with coverage 1-``alpha``.
+    Compute a confidence interval C with coverage 1-``alpha``
+    (``alpha=0.05`` is the default for all tests).
 
-    If ``tail`` is ``:both`` (default), then a two-sided confidence
+    If ``tail`` is ``:both`` (default for most tests), then a two-sided confidence
```
nalimilan: "most" isn't precise enough: give a list of tests for which it is the case below. Same remark for `pvalue`.

doc/parametric/test_t.rst
```diff
@@ -1,45 +1,57 @@
...
 
-.. function:: OneSampleTTest(v::AbstractVector{T<:Real}, mu0::Real=0)
+.. function:: OneSampleTTest(v::AbstractVector{T<:Real}, mu0::Real=0; tail::Symbol=:both, alpha::Real=0.05)
 
     Perform a one sample t-test of the null hypothesis that the data
     in vector ``v`` comes from a distribution with mean ``mu0`` against
     the alternative hypothesis that the distribution does not have mean
     ``mu0``.
+    
+    ``tail`` and ``alpha`` specify the defaults for the application of
```
nalimilan: "for the application of" is an unusual vocabulary. Maybe "when calling"?

src/HypothesisTests.jl
```diff
@@ -72,32 +72,35 @@ end
...
     end
     println(io)
 
     # test summary
-    p = pvalue(test)
-    outcome = if p > 0.05 "fail to reject" else "reject" end
-    tail = default_tail(test)
+    tail = HypothesisTests.tail(test)
```
nalimilan: `HypothesisTests.` isn't needed. Same below in comment.

src/HypothesisTests.jl
```diff
@@ -72,32 +72,35 @@ end
...
+    p = pvalue(test) # obeys value of HypothesisTests.tail(test) if applicable
+    outcome = if p > alpha(test) "fail to reject" else "reject" end
```
nalimilan: It's more standard to use `p > alpha(test) ? ... : ...`. Below, since the conditions are relatively long, would be better to repeat `tailvalue = ` inside each branch (i.e. closer to what it was before).

src/t.jl
```diff
@@ -28,18 +28,18 @@ export OneSampleTTest, TwoSampleTTest, EqualVarianceTTest,
 @compat abstract type TTest <: HypothesisTest end
 @compat abstract type TwoSampleTTest <: TTest end
 
-pvalue(x::TTest; tail=:both) = pvalue(TDist(x.df), x.t; tail=tail)
-
-default_tail(test::TTest) = :both
+pvalue(x::TTest; tail=x.tail) = pvalue(TDist(x.df), x.t; tail=tail)
+tail(x::TTest)  = x.tail    # defaults set by constructors
```
nalimilan: Rather than defining this here, can you just use these definitions for all `HypothesisTest` types, and add the needed fields to all tests? If `tail` doesn't make sense for some types, you can then override the method (to a hardcoded value) instead of adding the field.

src/t.jl
nalimilan: It's just that the term is not obvious, at least not to me. Just say something about blocking infinite recursion.

src/augmented_dickey_fuller.jl
nalimilan: Rather keep it, but call `tail` rather than `default_tail`.

src/t.jl
mirkobunse: Hi @nalimilan,

Adding a `tail` field to all types, which is not used by several of them, is not a quite clean solution. I would rather add the field to those types that actually can make use of it. These are quite many, though: Below, you find a list of types for which the `pvalue` function has a `tail` parameter, and those types for which `pvalue` has a fixed tail.

I will add a `tail` field to those types, for which the `pvalue` function takes a tail parameter.

The alpha value is relevant to all tests. Therefore, I will add an `alpha` field to all types. I will also add a simple function obtaining the test outcome:
```julia
reject_h0(test::HypothesisTest) = pvalue(test) <= test.alpha
```

I am looking forward to your comments!

---

`pvalue` with fixed tail (no tail parameter; 9 types):
- OneSampleADTest
- KSampleADTest
- ADFTest
- BoxPierceTest
- LjungBoxTest
- BreuschGodfreyTest
- RayleighTest
- JarqueBeraTest
- KruskalWallisTest

`pvalue` with tail parameter (15 types):
- BinomialTest
- SignTest
- FisherTLinearAssociation
- JammalamadakaCircularCorrelation
- FisherExactTest
- ExactKSTest
- ApproximateOneSampleKSTest
- ApproximateTwoSampleKSTest
- ExactMannWhitneyUTest
- ApproximateMannWhitneyUTest
- PowerDivergenceTest
- ExactSignedRankTest
- ApproximateSignedRankTest
- ZTest
- TTest

nalimilan: > Adding a tail field to all types, which is not used by several of them, is not a quite clean solution. I would rather add the field to those types that actually can make use of it. These are quite many, though: Below, you find a list of types for which the pvalue function has a tail parameter, and those types for which pvalue has a fixed tail.
> 
> I will add a tail field to those types, for which the pvalue function takes a tail parameter.

Yes, of course, no need to add the field to tests which don't need it, overloading the method will be enough for these.

> The alpha value is relevant to all tests. Therefore, I will add an alpha field to all types.

Right.

> I will also add a simple function obtaining the test outcome:
> 
> reject_h0(test::HypothesisTest) = pvalue(test) <= test.alpha

I'd rather not add this function, or at least not in this PR, as that's a different issue which deserves some discussion.

src/t.jl
```diff
@@ -28,18 +28,18 @@ export OneSampleTTest, TwoSampleTTest, EqualVarianceTTest,
...
+alpha(x::TTest) = x.alpha
 
 # confidence interval by inversion
-function StatsBase.confint(x::TTest, alpha::Float64=0.05; tail=:both)
+function StatsBase.confint(x::TTest, alpha::Float64=x.alpha; tail::Symbol=x.tail)
```
nalimilan: Use `alpha(x)` and `tail(x)` since that gives an identical signature for all tests.

doc/api/pvalue.rst
```diff
@@ -3,11 +3,11 @@
 p-value
 ==============================================
 
-.. function:: pvalue(test::HypothesisTest; tail=:both)
+.. function:: pvalue(test::HypothesisTest; tail=tail(test))
 
     Compute the p-value for a given significance test.
 
-    If ``tail`` is ``:both`` (default), then the p-value for the
```
nalimilan: Would be nice to mention that the default tail depends on the specific test, and give a list of tests with their default.

src/HypothesisTests.jl
```diff
@@ -72,32 +72,38 @@ end
...
-    println(io, "    outcome with 95% confidence: $outcome h_0")
-    if tail == :both
-        println(io, "    two-sided p-value:           $p")
-    elseif tail == :left || tail == :right
-        println(io, "    one-sided p-value:           $p")
+    testtail = tail(test)
+    p = pvalue(test) # obeys value of tail(test) if applicable
+    outcome = p > alpha(test) ? "fail to reject" : "reject"
+    if testtail == :both
+        plabel ="two-sided p-value:"
```
nalimilan: Missing space.

src/HypothesisTests.jl
```diff
@@ -72,32 +72,38 @@ end
...
+    elseif testtail == :left || testtail == :right
+        plabel = "one-sided p-value ($(string(testtail)) tail):"
     else
-        println(io, "    p-value:                     $p")
+        plabel = "p-value:"
     end
+    println(io, "Test summary:")
+    println(io, format_detail("outcome with "*conf_string*"% confidence:", outcome*" h_0", 36))
```
nalimilan: Add spaces around `*` for consistency and readability. Here you can actually use string interpolation instead.

src/HypothesisTests.jl
```diff
@@ -72,32 +72,38 @@ end
...
-        println(io, "    two-sided p-value:           $p")
-    elseif tail == :left || tail == :right
-        println(io, "    one-sided p-value:           $p")
+    testtail = tail(test)
+    p = pvalue(test) # obeys value of tail(test) if applicable
+    outcome = p > alpha(test) ? "fail to reject" : "reject"
+    if testtail == :both
+        plabel ="two-sided p-value:"
+    elseif testtail == :left || testtail == :right
+        plabel = "one-sided p-value ($(string(testtail)) tail):"
```
nalimilan: `string` isn't needed AFAIK.

src/t.jl
mirkobunse: In another branch, I already started to add the `alpha` field to all types, and the `tail` field those types for which the `pvalue` method has a tail parameter. I also develop corresponding unit tests and documentation.

Apparently, these changes are quite extensive. We should discuss what the present PR should be going for.

Initially, this PR was meant to (only) provide the _infrastructure_ for subtypes of HypothesisTest to have `tail` and `alpha` fields. Initially, I was not planning to actually add these fields to all types. Starting to do so confirmed my impression that this should be a different work package.

I see that all tests _should_ have `alpha` and `tail` fields (if applicable), but I suggest to delay these changes to a follow-up PR. I'd like the present PR to focus on the infrastructure, as initially intended. Of course, I will be applying the changes required to make this PR a complete extension to the HypothesisTests package.

nalimilan: I think it makes sense to add support for all tests at the same time. Else the package will be in an inconsistent state. What's the interest of merging the PR in a partial state?

PR #101, opened 2017-07-07
Power Divergence Bug/Test fixes

bdeonovic:
Noticed this fix in a pull request, but the fix was ignored because of ensuing discussion about how to do tests using simulation.


ararslan:
The Travis logs aren't loading for me so I can't see the cause of the CI failure. What is it that this fixes?

bdeonovic:
This pull is exactly same as https://github.com/JuliaStats/HypothesisTests.jl/pull/52

nalimilan:
AFAICT the test failure is due to the fact that you haven't kept enough digits in the expected values. You should copy the full printed value.

Otherwise, looks good to me.

bdeonovic:
Looks like this should be ready to squash and commit

bdeonovic:
Please pull

bdeonovic:
Bump?

src/power_divergence.jl
```diff
@@ -119,31 +120,24 @@ function ci_sison_glaz(x::PowerDivergenceTest, alpha::Float64; skew_correct::Boo
             lambda = x.observed[i]
             #run moments
             a = lambda + c
-            b = max(lambda - c, 0)
-            if lambda > 0
```
andreasnoack: Is `lambda` always positive?

src/power_divergence.jl
```diff
@@ -119,31 +120,24 @@ function ci_sison_glaz(x::PowerDivergenceTest, alpha::Float64; skew_correct::Boo
...
-                poislama = cdf(Poisson(lambda), a)
-                poislamb = cdf(Poisson(lambda), b - 1)
-            else
-                poislama = poislamb = 1.0
-            end
-            den = b > 0 ? poislama-poislamb : poislama
+            b = max(lambda - c, 0.0)
```
andreasnoack: Please don't change these literals. There are a few more below.

src/power_divergence.jl
```diff
@@ -106,9 +106,10 @@ end
 # adapted from SAS macro by May, Johnson (2000)
 function ci_sison_glaz(x::PowerDivergenceTest, alpha::Float64; skew_correct::Bool=true)
     k = length(x.thetahat)
-    probn = inv(pdf(Poisson(x.n), x.n))
-
+    probn = 1 /(cdf(Poisson(x.n), x.n) - cdf(Poisson(x.n), x.n - 1))
```
andreasnoack: Why use the `cdf` here instead of the `pdf`?

bdeonovic: I was trying to figure out where the bug was coming from and so I was making the code as close to the original paper description of the algorithm as I could to narrow down the issue. Turns out the bug was just closer to the bottom of the code. The pdf code should be equivalent here, would you prefer I revert to original line?

src/power_divergence.jl
bdeonovic: Yes, data is validated in PowerDivergenceTest function (x.observed[i] needs to be non-negative)

src/power_divergence.jl
bdeonovic: Sorry can you explain a bit more?

src/power_divergence.jl
andreasnoack: Yes. Please revert. The previous version is simpler and faster.

src/power_divergence.jl
andreasnoack: The original version uses `0` which is just fine here and more generic than the `Float64` literal which might be useful later.

bdeonovic:
Not really sure why nightly is failing to build.

src/power_divergence.jl
```diff
@@ -161,13 +155,13 @@ function ci_sison_glaz(x::PowerDivergenceTest, alpha::Float64; skew_correct::Boo
         s1, s2, s3, s4 = sum(m1), sum(m2), sum(m3), sum(m4)
         z  = (x.n - s1) / sqrt(s2)
         g1 = s3 / s2^(3/2)
-        g2 = s4 * s2^2
+        g2 = s4 / s2^2
 
         poly = 1 + g1 * (z^3 - 3z) / 6 + g2 * (z^4 - 6z^2 + 3) / 24 + g1^2 * (z^6 - 15z^4 + 45z^2 - 15) / 72
         f = poly * exp(-z^2 / 2) / sqrt(2π)
         probx = 1.0
```
andreasnoack: just make this `probx = prod(m5)`

PR #102, opened 2017-08-07
Add Durbin-Watson test

BenjaminBorn:
Implementation of the Durbin-Watson test for serial correlation in the residuals of a regression model.

WIP as I still have to implement Pan's algorithm (Farebrother, 1980) for the computation of exact p-values in small samples.

Everything else is ready and I would be glad to get feedback.

One question: I tried to include a displayed latex equation in the docstring as described in the documentation (and I understand it will not render the Latex code in the REPL or Juno console) but it does not even display the Latex code correctly.

Thanks as always for your help!


src/durbin_watson.jl
```diff
@@ -0,0 +1,138 @@
...
+    DurbinWatsonTest(X::AbstractArray, e::AbstractVector; p_compute::Symbol = :ndep)
+
+Compute the Durbin-Watson test for serial correlation in the residuals of a
+regression model.
+
+`X` is the matrix of regressors from the original regression model and `e` the
+vector of residuals. Note that the Durbin-Watson test is not valid if `X`
+includes a lagged dependent variable. The test statistic is computed as
+```math
+    DW = \frac{\sum_{t=2}^n (e_t - e_{t-1})^2}{\sum_{t=1}^n e_t^2}
```
ararslan: You'll need to escape the backslashes

BenjaminBorn:
I have now added Pan's algorithm (Farebrother, 1980) for the computation of exact p-values in small samples. Tests are comparing output to R.

@ararslan It seems like the p-value is computed three times when calling `t = DurbinWatsonTest(X, resid; p_compute = :exact)`. You can see this when running the last example of the test file (line 193). The warning is printed three times. Is this a general problem of the package that we should investigate or did I do something stupid?

src/durbin_watson.jl
```diff
@@ -92,9 +92,119 @@ function show_params(io::IO, x::DurbinWatsonTest, ident)
...
+  * R. W. Farebrother, 1980, "Algorithm AS 153: Pan's Procedure for the Tail Probabilities
+  of the Durbin-Watson Statistic", Journal of the Royal Statistical Society, Series C
+  (Applied Statistics), Vol. 29, No. 2, pp. 224-227,
+  [http://www.jstor.org/stable/2986316](http://www.jstor.org/stable/2986316).
+
+"""
+function pan_algorithm(a::AbstractArray, x::Float64, m::Int, n::Int)
+
+    try # catch case where ν is empty
+        ν = find(a.>=x)[1]
```
ararslan:
```julia
findfirst(ai -> ai >= x, a)
```
will be more efficient

src/durbin_watson.jl
```diff
@@ -92,9 +92,119 @@ function show_params(io::IO, x::DurbinWatsonTest, ident)
...
+            if ν <= h
+                d  = 2; h  = ν; k  = - k; j1 = 0; j2 = 2; j3 = 3; j4 = 1
+            else
+                d  = - 2; ν  = ν + 2; j1 = m - 2; j2 = m - 1; j3 = m + 1; j4 = m
+            end
+
+            pin = pi / (2n)
+            sum = 0.5 * (k + 1)
+            sgn = k / n
+            n2  = n + n -1
```
ararslan: `2n-1`?

src/durbin_watson.jl
```diff
@@ -92,9 +92,119 @@ function show_params(io::IO, x::DurbinWatsonTest, ident)
...
+                end
+                j2 = 0
+                ν  = 0
+            end
+            return sum
+        end
+    catch
+        sum = 1.0
+    end
+
```
ararslan: Missing a `return` here?

src/durbin_watson.jl
```diff
@@ -92,9 +92,119 @@ function show_params(io::IO, x::DurbinWatsonTest, ident)
...
 function pvalue(x::DurbinWatsonTest; tail=:both)
 
-    if (x.p_compute == :ndep && x.n > 100) || x.p_compute == :approx
+    exact_problem_flag = 0
+    if (x.p_compute == :ndep && x.n <= 100) || x.p_compute == :exact
+        # p-vales based on Pan's algorithm (see Farebrother, 1980)
+
+        # the following setup is, e.g, described in Durbin and Watson (1971)
+        A = diagm(-1 * ones(x.n - 1), -1) + diagm(-1 * ones(x.n - 1), 1) +diagm(
```
ararslan: Can just use `-ones(x.n - 1)` rather than multiplying by `-1`

src/durbin_watson.jl
```diff
@@ -0,0 +1,242 @@
...
+            h = m - ν
+
+            if ν <= h
+                d  = 2; h  = ν; k  = - k; j1 = 0; j2 = 2; j3 = 3; j4 = 1
+            else
+                d  = - 2; ν  = ν + 2; j1 = m - 2; j2 = m - 1; j3 = m + 1; j4 = m
+            end
+
+            pin = pi / (2n)
+            sum = 0.5 * (k + 1)
```
ararslan: Typically it's preferred to avoid float literals in computations such as this, since float literals are always `Float64`, which will promote smaller float types to `Float64` when they're used together in a computation. So in this case it would be `sum = (k + 1) / 2` (though also note that it doesn't _really_ matter, since that will promote to `Float64` anyway, but the general point still stands 😛). It might be worthwhile to check `@code_warntype pan_algorithm(<some args>)` to see whether the function remains type-stable.

src/durbin_watson.jl
```diff
@@ -0,0 +1,242 @@
...
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+# LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+# OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+# WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+export DurbinWatsonTest
+
+struct DurbinWatsonTest <: HypothesisTest
```
ararslan: The package still supports Julia 0.5 but `struct` only parses on 0.6 and later. So this should be `immutable` until we drop 0.5 support.

ararslan:
> Is this a general problem of the package that we should investigate

Yeah, I could be wrong but I'm pretty sure it's the fault of the general `show` method and not a fault of this particular test.

src/durbin_watson.jl
BenjaminBorn: This is great as it allows me to get rid of the try-catch construct.

src/durbin_watson.jl
BenjaminBorn: It seems I don't need one here as it returns the last value automatically.

src/durbin_watson.jl
BenjaminBorn: Thanks, makes sense. Never thought about that.

src/durbin_watson.jl
```diff
@@ -0,0 +1,238 @@
...
+  of the Durbin-Watson Statistic", Journal of the Royal Statistical Society, Series C
+  (Applied Statistics), Vol. 29, No. 2, pp. 224-227,
+  [http://www.jstor.org/stable/2986316](http://www.jstor.org/stable/2986316).
+
+"""
+function pan_algorithm(a::AbstractArray, x::Float64, m::Int, n::Int)
+
+    ν = findfirst(ai -> ai >= x, a)
+    if ν == 0
+        return sum = 1.0
```
ararslan: Shouldn't need the `sum =`, right?

BenjaminBorn:
I have opened an issue about the double computation of the p-value in #105.

BenjaminBorn:
Thanks for the quick and very helpful review Alex!

BenjaminBorn:
Any other comments?

src/durbin_watson.jl
```diff
@@ -0,0 +1,238 @@
...
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+# LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+# OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+# WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+export DurbinWatsonTest
+
+immutable DurbinWatsonTest <: HypothesisTest
+    xmat::Array{Float64}  # regressor matrix
```
nalimilan: If that's a matrix it should be `Matrix{Float64}` to use a concrete type.

src/durbin_watson.jl
```diff
@@ -0,0 +1,238 @@
...
+
+`X` is the matrix of regressors from the original regression model and `e` the vector of
+residuals. Note that the Durbin-Watson test is not valid if `X` includes a lagged dependent
+variable. The test statistic is computed as
+```math
+DW = \\frac{\\sum_{t=2}^n (e_t - e_{t-1})^2}{\\sum_{t=1}^n e_t^2}
+```
+where `n` is the number of observations.
+
+By default, the choice of approach to compute p-values depends on the sample size (`p_compute=:ndep`). For small samples (n<100), Pan's algorithm (Farebrother, 1980) is
```
nalimilan: Break line at 92 chars.

src/durbin_watson.jl
```diff
@@ -0,0 +1,238 @@
...
+  [http://www.jstor.org/stable/2986316](http://www.jstor.org/stable/2986316).
+
+# External links
+
+  * [Durbin-Watson test on Wikipedia:
+  https://en.wikipedia.org/wiki/Durbin–Watson_statistic
+  ](https://en.wikipedia.org/wiki/Durbin–Watson_statistic)
+"""
+function DurbinWatsonTest{T<:Real}(xmat::AbstractArray{T}, e::AbstractArray{T};
+    p_compute::Symbol = :ndep)
```
nalimilan: Incorrect indentation.

src/durbin_watson.jl
```diff
@@ -0,0 +1,238 @@
...
+
+        if p_temp < 0.0 || p_temp > 1.0
+            # println(p_temp)
+            warn("Exact p-values outside [0,1]. Approximate p-values reported instead.")
+            exact_problem_flag = 1
+        end
+    end
+
+    if exact_problem_flag == 1 || (x.p_compute == :ndep && x.n > 100
+        ) || x.p_compute == :approx
```
nalimilan: Parenthesis should be on previous line.

src/durbin_watson.jl
```diff
@@ -0,0 +1,238 @@
...
+
+  * [Durbin-Watson test on Wikipedia:
+  https://en.wikipedia.org/wiki/Durbin–Watson_statistic
+  ](https://en.wikipedia.org/wiki/Durbin–Watson_statistic)
+"""
+function DurbinWatsonTest{T<:Real}(xmat::AbstractArray{T}, e::AbstractArray{T};
+    p_compute::Symbol = :ndep)
+
+    n = length(e)
+    DW = sum(diff(e) .^2) / sum(e .^2)
```
nalimilan: I don't know how much it can matter for performance, but you can use `sum(abs2, diff(e))/sum(abs2, e)` to avoid allocating copies. (`diff` itself allocates a vector, which could be avoided by using a small loop.)

src/durbin_watson.jl
```diff
@@ -0,0 +1,238 @@
...
+  Regression: III", Biometrika, Vol. 58, No. 1, pp. 1-19,
+  [http://www.jstor.org/stable/2334313](http://www.jstor.org/stable/2334313).
+  * R. W. Farebrother, 1980, "Algorithm AS 153: Pan's Procedure for the Tail Probabilities
+  of the Durbin-Watson Statistic", Journal of the Royal Statistical Society, Series C
+  (Applied Statistics), Vol. 29, No. 2, pp. 224-227,
+  [http://www.jstor.org/stable/2986316](http://www.jstor.org/stable/2986316).
+
+"""
+function pan_algorithm(a::AbstractArray, x::Float64, m::Int, n::Int)
+
```
nalimilan: The convention used in Julia and in JuliaStats packages is not to add blank lines at the top nor at the end of functions.

BenjaminBorn:
Thanks Milan! I'll revisit this after the smoke has cleared from the documentation migration.

Issue #103, opened 2017-08-08
Migrate documentation to Documenter


ararslan:
The following things would need to happen as part of the migration:

* Rename doc/ to docs/ (`git mv doc docs`)
* Convert ReST files to Markdown
* Move specific method and type documentation to docstrings

BenjaminBorn:
Would you prefer all of this in one big pull request or would it be ok to go step by step? If it's ok, I would start with a PR that moves specific method and type documentation to docstrings.

ararslan:
Sure, that'd be fantastic!

BenjaminBorn:
I also managed to build the new docs locally using Documenter with native HTML output. I'll send a PR once #106 is merged. If I understand it correctly (I'm a total Documenter noob), we need Travis and Github keys. @ararslan can you set them up for the repository?

ararslan:
Can do.

ararslan:
Keys added and gh-pages branch created.

ararslan:
This task has been valiantly completed by the intrepid @BenjaminBorn.

PR #106, opened 2017-08-21
Move method descriptions to docstrings

BenjaminBorn:
In this pull request, I have copied all available documentation on tests, pvalues, and confidence intervals to docstrings. Please note that I have only done some reformatting. I have not checked the documentation for correctness. This is the first step in the transition to Documenter.jl as discussed in #103. 

When reviewing, please consider the following questions.

1. I have attached the general description of `pvalue()` and `confint()` somewhat arbitrarily to two of the tests (`Binomial.jl` and `t.jl`). Is there a better place to put them?
2. Sometimes a test has different methods, e.g. `OneSampleTTest()`. If possible, I have combined them in one docstring, but sometimes, for clarity reasons, I have included different docstrings for the methods. Is that ok, or should we combine them somehow into one docstring? The disadvantage of having them in different docstrings is that the help output in the REPL can get convoluted, especially as there is not much spacing between the output for the different methods.
3. `pvalue` and `confint` also have different methods with a number of separate docstrings. Is there a way to ensure that the general method, e.g. `pvalue(x::HypothesisTests)`, is the first shown when typing `?pvalue`? As some of the more specialised methods have long docstrings, it could get lost otherwise.
3. The admonition `!!! note` doesn't work in the Jupyter notebook (at least not for me using Safari). It works in the REPL and Juno. Maybe someone can try it. If it's a general problem, we can file an issue with IJulia (or at the appropriate place).
4. There was a copy/paste mistake in the docstring for the binomial confidence intervals. After fixing it, I realised that there is also an open PR (#55) for it. Is there any way of giving credit for the fix to @juliangehring ?

Thanks in advance for your input!


BenjaminBorn:
One idea for question (1) would be to put a "fake" general method in `HypothesisTests.jl` à la
```
"""
    confint(test::HypothesisTest, alpha = 0.05; tail = :both)

Compute a confidence interval C with coverage 1-`alpha`.

...

"""
confint(test::HypothesisTest, alpha = 0.05; tail = :both)
```
And similarly for `pvalue`. Would that be preferable to attaching it to a random method?

ararslan:
A way to generally attach docstrings to functions is to either forward-declare the function with no methods, e.g.
```julia
"""
    confint(...)

...
"""
function confint end
```
or attach it to the function after methods have been defined for it, e.g.
```
"""
    confint(...)

...
"""
confint
```

src/fisher.jl
```diff
@@ -36,11 +36,37 @@ immutable FisherExactTest <: HypothesisTest
 
     # conditional maximum likehood estimate of odd ratio
     ω::Float64
+end
 
-    function FisherExactTest(a::Int, b::Int, c::Int, d::Int)
```
nalimilan: Shouldn't move this function out of the type definition: this would make it an outer constructor rather than an inner constructor. Just attach the docstring to the type itself.

src/fisher.jl
```diff
@@ -36,11 +36,37 @@ immutable FisherExactTest <: HypothesisTest
...
+
+The contingency table is structured as:
+
+| -  | X1 | X2 |
+|:--:|:--:|:--:|
+|*Y1*| a  | b  |
+|*Y2*| c  | d  |
+
+!!! note
+    The print output contains the conditional maximum likelihood estimate of the odd-ratio
```
nalimilan: "odds ratio" (also fix other occurrences below). Also, "print" would better be "show", with a link to the function.

src/fisher.jl
```diff
@@ -59,6 +85,41 @@ function show_params(io::IO, x::FisherExactTest, ident="")
 end
 
 # DOC: for tail=:both there exist multiple ``method``s for computing a pvalue and the corresponding ci.
+"""
+    pvalue(x::FisherExactTest; tail = :both, method = :central)
+
+Compute the p-value for a given significance test.
+
+The one-sided p-values are based on Fisher's non-central hypergeometric distribution
+``f_\\omega(i)`` with odd-ratio ``\\omega``:
```
nalimilan: You can use the omega character directly. Same for `leq` and `geq`.

src/fisher.jl
```diff
@@ -59,6 +85,41 @@ function show_params(io::IO, x::FisherExactTest, ident="")
 end
 
 # DOC: for tail=:both there exist multiple ``method``s for computing a pvalue and the corresponding ci.
+"""
+    pvalue(x::FisherExactTest; tail = :both, method = :central)
+
+Compute the p-value for a given significance test.
```
nalimilan: "for a given Fisher exact test".

src/fisher.jl
```diff
@@ -36,11 +36,37 @@ immutable FisherExactTest <: HypothesisTest
...
+    The print output contains the conditional maximum likelihood estimate of the odd-ratio
+    rather than the sample odds ratio; it maximizes the likelihood given by Fisher's
+    non-central hypergeometric distribution.
+
+Implements: [`pvalue`](@ref), [`confint`](@ref)
+
+# References
+
+  * Fay, M.P. Supplementary material to confidence intervals that match Fisher’s exact or
+    Blaker’s exact tests. Biostatistics, 0(0): 1-13, 2009.
```
nalimilan: "0(0)" is suspicious. Also below.

src/fisher.jl
```diff
@@ -59,6 +85,41 @@ function show_params(io::IO, x::FisherExactTest, ident="")
...
+        p_\\omega = \\sum_{f_\\omega(i)\\leq f_\\omega(a)} f_\\omega(i)
+    ```
+
+!!! note
+    Since the p-value is not necessarily unimodal, the corresponding confidence region might
+    not be an interval.
+
+# References
+
+  * Gibbons, J.D, Pratt, J.W. P-values: Interpretation and Methodology American
```
nalimilan: Missing dot before review name.

src/fisher.jl
```diff
@@ -97,6 +158,18 @@ function pvalue_both_minlike(x::FisherExactTest, ω::Float64=1.0)
 end
 
 # confidence interval by inversion of p-value
+"""
+    confint(x::FisherExactTest, alpha::Float64=0.05; tail=:both, method=:central)
+
+Compute a confidence interval with coverage 1 - `alpha` by inverting the `:central` p-value.
```
nalimilan: Would be nice to repeat the definition of different methods here. AFAIK there's no need to mention the p-value here at all.

src/fisher.jl
```diff
@@ -59,6 +85,41 @@ function show_params(io::IO, x::FisherExactTest, ident="")
...
+    one-sided p-values.
+
+  - Minimum likelihood interval `:minlike`: This p-value is computed by summing all tables
+    with the same marginals that are equally or less probable:
+    ```math
+        p_\\omega = \\sum_{f_\\omega(i)\\leq f_\\omega(a)} f_\\omega(i)
+    ```
+
+!!! note
+    Since the p-value is not necessarily unimodal, the corresponding confidence region might
```
nalimilan: Why mention this here and not in the docstring for `confint`?

src/kolmogorov_smirnov.jl
```diff
@@ -54,6 +54,15 @@ immutable ExactOneSampleKSTest <: ExactKSTest
     δn::Float64 # supremum of the negative CDF differences
 end
 
+"""
+    ExactOneSampleKSTest{T<:Real}(x::AbstractVector{T}, d::UnivariateDistribution)
+
+Perform a one sample Kolmogorov–Smirnov-test of the null hypothesis that the data in vector
```
nalimilan: "Kolmogorov–Smirnov test". Same below.

Would also make sense to mention "exact".

src/kolmogorov_smirnov.jl
```diff
@@ -129,6 +147,21 @@ immutable ApproximateTwoSampleKSTest <: ApproximateKSTest
     δn::Float64 # suproemum of the negative CDF differences
 end
 
+"""
+    ApproximateTwoSampleKSTest{T<:Real, S<:Real}(x::AbstractVector{T}, y::AbstractVector{S})
+
+Perform an asymptotic two sample Kolmogorov–Smirnov-test of the null hypothesis that `x`
+and `y` are drawn from the same distribution against the alternative hypothesis that the
+distribution comes from different distributions.
```
nalimilan: "distribution comes" -> "they come"

src/kolmogorov_smirnov.jl
```diff
@@ -129,6 +147,21 @@ immutable ApproximateTwoSampleKSTest <: ApproximateKSTest
...
+
+Implements: [`pvalue`](@ref)
+
+# External links
+
+  * [Approximation of one-sided test:
+     https://www.encyclopediaofmath.org/index.php/Kolmogorov-Smirnov_test](
```
nalimilan: Better give just the full name of the site rather than repeating the URL.

src/kruskal_wallis.jl
```diff
@@ -32,6 +32,35 @@ immutable KruskalWallisTest <: HypothesisTest
     tie_adjustment::Float64  # adjustment for ties
 end
 
+"""
+    KruskalWallisTest{T<:Real}(groups::AbstractVector{T}...)
+
+Perform Kruskal Wallis rank sum test of the null hypothesis that the location parameters of
```
nalimilan: "Kruskal-Wallis"

src/kruskal_wallis.jl
```diff
@@ -32,6 +32,35 @@ immutable KruskalWallisTest <: HypothesisTest
...
+the distribution of the ``n`` observations are the same in each of the `groups`
```
nalimilan: It's not clear what `n` corresponds to.

src/kruskal_wallis.jl
```diff
@@ -32,6 +32,35 @@ immutable KruskalWallisTest <: HypothesisTest
...
+``\\mathcal{G}`` against the alternative hypothesis that they differ in at least one.
```
nalimilan: "at least one" sounds weird to me. I can't find this description in other references.

src/mann_whitney.jl
```diff
@@ -52,7 +71,17 @@ end
 
 
 ## EXACT MANN-WHITNEY U TEST
+"""
+    ExactMannWhitneyUTest(x::AbstractVector{T<:Real}, y::AbstractVector{T<:Real})
 
+Perform an exact Mann-Whitney U test.
```
nalimilan: Copy the definition of this test from the above docstring, as it should be usable in isolation. Same for the approximate test.

src/mann_whitney.jl
```diff
@@ -52,7 +71,17 @@ end
 
 
 ## EXACT MANN-WHITNEY U TEST
+"""
```
nalimilan: Better attach this docstring to the constructor than to the type since it documents the function. Same below.

src/power_divergence.jl
```diff
@@ -44,6 +43,28 @@ default_tail(test::PowerDivergenceTest) = :right
 
 pvalue(x::PowerDivergenceTest; tail=:right) = pvalue(Chisq(x.df),x.stat; tail=tail)
 
+"""
+    confint(test::PowerDivergenceTest, alpha = 0.05; tail = :both, method = :sison_glaz)
+
+Compute a confidence interval with coverage 1-`alpha` for multinomial proportions using
+one of the following methods. Possible values for `method` are:
+
+  - Sison, Glaz intervals `:sison_glaz` (default)
```
nalimilan: Would be great if you could describe these a bit.

src/power_divergence.jl
```diff
@@ -44,6 +43,28 @@ default_tail(test::PowerDivergenceTest) = :right
...
+
+# References
+
+  * Agresti, Alan. Categorical Data Analysis, 3rd Edition. Wiley, 2013.
+  * Sison, C.P and Glaz, J. Simultaneous confidence intervals and sample size determination
+    for multinomial proportions. Journal of the American Statistical Association,
+    90:366-369, 1995.
+  * Quesensberry, C.P. and Hurst, D.C. Large Sample Simultaneous Confidence Intervals for
+    Multinational Proportions. Technometrics, 6:191-195, 1964.
+  * Gold, R. Z. Tests Auxiliary to :math:`{\chi^{2}}` Tests in a Markov Chain. Annals of
```
nalimilan: You can write  χ² directly.

src/power_divergence.jl
```diff
@@ -214,6 +235,48 @@ end
 
 # Under regularity conditions, their asymptotic distributions are all the same (Drost 1989)
 # Chi-squared null approximation works best for lambda near 2/3
+"""
+    PowerDivergenceTest(x [,y] [, lambda] [,theta0] )
```
nalimilan: Spaces must go after commas. Also mention the default values of `lambda` and `theta`. It could make sense to use two lines for the signature, one without `y` and one with it.

src/power_divergence.jl
```diff
@@ -214,6 +235,48 @@ end
...
+
+Perform a Power Divergence Test.
+
+If `x` is a matrix with one row or column, or if `x` is a vector and `y` is not given, then
+a goodness-of-fit test is performed (``x`` is treated as a one-dimensional contingency
```
nalimilan: Single backticks around `x`. Missing closing parenthesis.

src/power_divergence.jl
```diff
@@ -214,6 +235,48 @@ end
...
+table. The entries of `x` must be non-negative integers. In this case, the hypothesis
+tested is whether the population probabilities equal those in `theta0`, or are all equal if
+`theta0` is not given.
+
+If `x` is a matrix with at least two rows and columns, it is taken as a two-dimensional
+contigency table: the entries of `x` must be non-negative integers. Otherwise, `x` and
```
nalimilan: "contingency". It seems that "the entries of `x` must be non-negative integers" applies all the time, and also to `y` when it is passed, so maybe better mention this at the end in a separate paragraph?

src/power_divergence.jl
```diff
@@ -214,6 +235,48 @@ end
 
 # Under regularity conditions, their asymptotic distributions are all the same (Drost 1989)
 # Chi-squared null approximation works best for lambda near 2/3
+"""
+    PowerDivergenceTest(x [,y] [, lambda] [,theta0] )
+
+Perform a Power Divergence Test.
+
+If `x` is a matrix with one row or column, or if `x` is a vector and `y` is not given, then
```
nalimilan: It's not clear whether "`y` is not given" applies to both cases or only to the first one.

src/power_divergence.jl
```diff
@@ -214,6 +235,48 @@ end
...
+a goodness-of-fit test is performed (``x`` is treated as a one-dimensional contingency
+table. The entries of `x` must be non-negative integers. In this case, the hypothesis
+tested is whether the population probabilities equal those in `theta0`, or are all equal if
+`theta0` is not given.
+
+If `x` is a matrix with at least two rows and columns, it is taken as a two-dimensional
+contigency table: the entries of `x` must be non-negative integers. Otherwise, `x` and
+`y` must be vectors of the same length. The contigency table is calculated using `counts`
+from `Statsbase`. Then the power divergence test is performed of the null hypothesis that
```
nalimilan: Don't mention the package, just use `@ref`.

src/power_divergence.jl
```diff
@@ -214,6 +235,48 @@ end
...
+the joint distribution of the cell counts in a 2-dimensional contingency table is the
+product of the row and column marginals.
+
+The power divergence test is given by
+```math
+    \\dfrac{2}{\\lambda(\\lambda+1)}\\sum_{i=1}^I \\sum_{j=1}^J n_{ij} \\left[(n_{ij}
+    /\\hat{n}_{ij})^\\lambda -1\\right]
+```
+where ``n_{ij}`` is the cell count in the ``i`` th row and ``j`` th column and ``\\lambda``
+is a real number that has the following feature
```
nalimilan: "is a real number that has the following feature" -> "is a real number determing the nature of the test to be performed:".

src/power_divergence.jl
```diff
@@ -214,6 +235,48 @@ end
...
+
+  * ``λ = 1``: this is equal to Pearson's chi-squared statistic
+  * ``λ \\to 0``: it converges to the likelihood ratio test statistic
+  * ``λ \\to -1``: it converges to the minimum discrimination information statistic
+    (Gokhale and Kullback 1978)
+  * ``λ = -2``: it equals Neyman modified chi-squared (Neyman 1949)
+  * ``λ = -1/2``:  it equals the Freeman-Tukey statistic (Freeman and Tukey 1950).
+
+Under regularity conditions, the asymptotic distributions are identical (see Drost et. al.
```
nalimilan: "asymptotic distributions" of what?

src/power_divergence.jl
```diff
@@ -214,6 +235,48 @@ end
...
+1989). The chi-squared null approximation works best for ``λ`` near ``2/3``.
```
nalimilan: What's the "the chi-squared null approximation" in the list above?

src/power_divergence.jl
```diff
@@ -299,6 +367,11 @@ ChisqTest{T<:Integer,U<:AbstractFloat}(x::AbstractVector{T}, theta0::Vector{U} =
     PowerDivergenceTest(reshape(x, length(x), 1), lambda=1.0, theta0=theta0)
 
 #MultinomialLRT
+"""
+    MultinomialLRT(x [,y] [,theta0])
+
+Convenience function for power divergence test with ``λ=0``.
```
nalimilan: Explain more what this test means so that it can be used in isolation. Link to `PowerDivergenceTest`.

src/t.jl
```diff
@@ -33,6 +33,24 @@ pvalue(x::TTest; tail=:both) = pvalue(TDist(x.df), x.t; tail=tail)
...
+Compute a confidence interval C with coverage 1-`alpha`.
+
+If `tail` is `:both` (default), then a two-sided confidence interval is returned. If `tail`
+is `:left` or `:right`, then a one-sided confidence interval is returned.
+
+!!! note
+    Most of the implemented confidence intervals are *strongly consistent*, that is, the
+    confidence interval with coverage 1-`alpha` does not contain the test statistic under
+    ``h_0`` if and only if the corresponding test rejects the null hypothesis
+    ``h_0: \\theta=\\theta_0``:
```
nalimilan: You can write alpha and theta directly using greek letters.

src/t.jl
```diff
@@ -108,6 +153,15 @@ end
 testname(::EqualVarianceTTest) = "Two sample t-test (equal variance)"
 population_param_of_interest(x::TwoSampleTTest) = ("Mean difference", x.μ0, x.xbar) # parameter of interest: name, value under h0, point estimate
 
+"""
+    EqualVarianceTTest(x::AbstractVector{T<:Real}, y::AbstractVector{T<:Real})
+
+Perform a two-sample t-test of the null hypothesis that `x` and `y` come from a
+distributions with the same mean and equal variances against the alternative hypothesis
```
nalimilan: "a distributions  with the same mean and equal variances" -> "distributions with equal means and variances". Same below.

src/wilcoxon.jl
```diff
@@ -27,6 +27,21 @@ export SignedRankTest, ExactSignedRankTest, ApproximateSignedRankTest
...
+    SignedRankTest(x::AbstractVector{T<:Real}, y::AbstractVector{T<:Real})
+
+Perform a Wilcoxon signed rank test of the null hypothesis that the distribution of `x`
+(or the difference `x - y`) has zero median against the alternative hypothesis that the
+median is non-zero.
+
+When there are no tied ranks and ≤50 samples, or tied ranks and ≤15 samples,
+`SignedRankTest` performs an exact signed rank test. In all other cases, `SignedRankTest`
+performs an approximate signed rank test. Behavior may be further controlled by using
+`ExactSignedRankTest` or `ApproximateSignedRankTest` directly.
```
nalimilan: Use `@ref`.

src/mann_whitney.jl
```diff
@@ -27,6 +27,25 @@ export MannWhitneyUTest, ExactMannWhitneyUTest, ApproximateMannWhitneyUTest
...
+Perform a Mann-Whitney U test of the null hypothesis that the probability that an
+observation drawn from the same population as `x` is greater than an observation drawn
+from the same population as `y` is equal to the probability that an observation drawn
+from the same population as `y` is greater than an observation drawn from the same
+population as `x` against the alternative hypothesis that these probabilities are not
+equal.
+
+The Mann-Whitney U test is sometimes known as the Wilcoxon rank sum test.
+
+When there are no tied ranks and ≤50 samples, or tied ranks and ≤10 samples,
```
nalimilan: Use `@ref`.

src/wilcoxon.jl
```diff
@@ -62,6 +77,17 @@ immutable ExactSignedRankTest{T<:Real} <: HypothesisTest
     n::Int                  # number of observations
     median::Float64         # sample median
 end
+"""
+    ExactSignedRankTest(x::AbstractVector{T<:Real}[, y::AbstractVector{T<:Real}])
+
+Perform an exact signed rank U test.
```
nalimilan: Repeat the definition of the test. Same below.

nalimilan:
> I have attached the general description of pvalue() and confint() somewhat arbitrarily to two of the tests (Binomial.jl and t.jl). Is there a better place to put them?

The usual solution to that is to define `function pvalue end` somewhere and attach the generic docstring to that.

> Sometimes a test has different methods, e.g. OneSampleTTest(). If possible, I have combined them in one docstring, but sometimes, for clarity reasons, I have included different docstrings for the methods. Is that ok, or should we combine them somehow into one docstring? The disadvantage of having them in different docstrings is that the help output in the REPL can get convoluted, especially as there is not much spacing between the output for the different methods.

I'd say that's OK as long as the repeated content is not too long. If the REPL output isn't clear, it should be improved, rather than working around it in packages (the HTML manual looks better in general).

> pvalue and confint also have different methods with a number of separate docstrings. Is there a way to ensure that the general method, e.g. pvalue(x::HypothesisTests), is the first shown when typing ?pvalue? As some of the more specialised methods have long docstrings, it could get lost otherwise.

IIRC it depends on the order in which the methods are defined. Using the trick I gave above, you should be able to define it in the right place.

> The admonition !!! note doesn't work in the Jupyter notebook (at least not for me using Safari). It works in the REPL and Juno. Maybe someone can try it. If it's a general problem, we can file an issue with IJulia (or at the appropriate place).

Agreed. Also check that it works in the HTML manual though.

> There was a copy/paste mistake in the docstring for the binomial confidence intervals. After fixing it, I realised that there is also an open PR (#55) for it. Is there any way of giving credit for the fix to @juliangehring ?

Yes, the best way to give her credit is to merge here PR, which I just did. Thanks for pointing it out!

src/anderson_darling.jl
```diff
@@ -30,6 +30,15 @@ immutable OneSampleADTest <: ADTest
     A²::Float64 # Anderson-Darling test statistic
 end
 
+"""
+    OneSampleADTest{T<:Real}(x::AbstractVector{T}, d::UnivariateDistribution)
```
BenjaminBorn: Only in the docstring or should I also go through the function definitions?

nalimilan: You could do both if `T` isn't used in the body of functions, but to keep this PR focused better only touch docstrings.

BenjaminBorn:
Ok, I have (hopefully) addressed all points directly related to the PR. I will now make a pass through the general doc improvement comments and commit them separately, however some specific questions concerning the tests are outside my expertise. Should I just leave them open?

src/kruskal_wallis.jl
BenjaminBorn: At least one group has to stochastically dominate one other group.

src/power_divergence.jl
BenjaminBorn: I think it applies to both cases and have reformulated accordingly.

src/binomial.jl
```diff
@@ -53,20 +66,52 @@ function show_params(io::IO, x::BinomialTest, ident="")
...
+    they are centered around the same value. The Agresti Coull interval has higher or
+    equal coverage.
+  - Arcsine transformation `:arcsine`.
+
+# References
+
+  * Brown, L.D., Cai, T.T., and DasGupta, A. Interval estimation for a binomial proportion.
+    Statistical Science, 16(2):101–117, 2001.
+
+# External links
```
BenjaminBorn: I would prefer to keep the links in a separate category for consistency reasons. That's how I have also done it for the time-series tests.

src/binomial.jl
```diff
@@ -53,20 +66,52 @@ function show_params(io::IO, x::BinomialTest, ident="")
...
+  - Wilson score interval `:wilson`: This interval relies on a normal approximation. In
+    contrast to `:wald`, the standard deviation is not approximated by an empirical
+    estimate resulting in good empirical coverages even for small numbers of draws and
+    extreme success probabilities.
+  - Jeffreys interval `:jeffrey`: Bayesian confidence interval obtained by using a
+    non-informative Jeffreys prior. The interval is very similar to the Wilson interval.
+  - Agresti Coull interval `:agresti_coull`: Simplified version of the Wilson interval;
+    they are centered around the same value. The Agresti Coull interval has higher or
+    equal coverage.
+  - Arcsine transformation `:arcsine`.
```
BenjaminBorn: I have said a word about it ;-) However, I don't know how well they do compared to the other intervals.

src/kruskal_wallis.jl
```diff
@@ -32,6 +32,41 @@ immutable KruskalWallisTest <: HypothesisTest
...
+Perform Kruskal-Wallis rank sum test of the null hypothesis that the `groups`
+``\\mathcal{G}`` come from the same distribution against the alternative hypothesis that
+that at least one group stochastically dominates one other group.
+
+The Kruskal-Wallis test is an extension of the Mann-Whitney U test to more than two groups.
+
+The p-value is computed using a ``χ^2`` approximation to the distribution of the test
+statistic ``H_c=\\frac{H}{C}``:
+```math
+    \\begin{align}
```
ararslan: I forget, does `align` number the lines and `align*` doesn't?

ararslan:
```@docs
    pvalue
    ```

🙂

**Comment by BenjaminBorn** [comment]
````
```@docs
pvalue
```
````
But that prints the docstrings for *all* defined methods of `pvalue`. What if I only want the general one? For example so that I can structure the documentation and put `pvalue(::Binomial)` somewhere else.

File: src/kruskal_wallis.jl (hunk 36 above)
Comment (BenjaminBorn) [inline src/kruskal_wallis.jl]: Good point, `align` numbers (although the Jupyter notebook doesn't print them). I'll change it to `align*` in the next iteration.

**Comment by ararslan** [comment]
Ah, I see. I _think_ you should be able to do this:
```julia
"""
    pvalue(...)

Some general description
"""
pvalue(::HypothesisTest; kwargs...)
```
after methods have been defined for `pvalue`, rather than the 0-method forward declaration.

BenjaminBorn:
Thanks, but even if I put the definition at the end of `HypothesisTests.jl` after everything has been defined (and the docstring can be found, e.g. in the REPL), I still get
> ERROR: LoadError: UndefVarError: HypothesisTest not defined
while loading /Users/bborn/.julia/v0.6/HypothesisTests/docs/make.jl, in expression starting on line 3

when I add `pvalue(::HypothesisTest)` to the `@docs` block.

ararslan:
Hmmmmm, maybe

    ```@meta
    CurrentModule = HypothesisTests
    ```

    ```@docs
    pvalue(::HypothesisTest)
    ```

src/t.jl
```diff
@@ -70,15 +70,42 @@ function show_params(io::IO, x::OneSampleTTest, ident="")
     println(io, ident, "empirical standard error: $(x.stderr)")
 end
 
+"""
+    OneSampleTTest(xbar::Real, stddev::Real, n::Int, μ0::Real = 0)
+
+Perform a one sample t-test of the null hypothesis that `n` values with mean `xbar` and
+sample standard deviation `stddev`  come from a distribution with `μ0` against the
```
nalimilan: Still need to fix "with mean `μ0`".

src/binomial.jl
nalimilan: OK, let's leave this for somebody else then. ;-)

src/power_divergence.jl
```diff
@@ -214,6 +235,48 @@ end
...
+Perform a Power Divergence test.
+
+If `y` is not given and `x` is a matrix with one row or column, or `x` is a vector, then
+a goodness-of-fit test is performed (`x` is treated as a one-dimensional contingency
+table). In this case, the hypothesis tested is whether the population probabilities equal
+those in `theta0`, or are all equal if `theta0` is not given.
+
+If `x` is a matrix with at least two rows and columns, it is taken as a two-dimensional
+contingency table. Otherwise, `x` and `y` must be vectors of the same length. The contingency
+table is calculated using `counts` from [`Statsbase`](@ref). Then the power divergence test
```
nalimilan: "StatsBase". Also, I meant writing``[`StatsBase.counts`](@ref)`` (I'm not sure a link to a module works).

src/power_divergence.jl
```diff
@@ -214,6 +235,48 @@ end
...
+
+Note that the entries of `x` (and `y` if provided) must be non-negative integers.
+
+The power divergence test is given by
+```math
+    \\dfrac{2}{λ(λ+1)}\\sum_{i=1}^I \\sum_{j=1}^J n_{ij} \\left[(n_{ij}
+    /\\hat{n}_{ij})^λ -1\\right]
+```
+where ``n_{ij}`` is the cell count in the ``i`` th row and ``j`` th column and ``λ`` is a
+real number determing the nature of the test to be performed:
```
nalimilan: "determining"

src/power_divergence.jl
nalimilan: Please at least copy the relevant docs from `PowerDivergenceTest` and fix the signature.

src/power_divergence.jl
BenjaminBorn: The `Documenter.jl` documentation states

> Note that depending on what the CurrentModule is set to, a docstring @ref may need to be prefixed by the module which defines it.

However, at least in my local build, links like ``[`StatsBase.counts`](@ref)`` don't seem to work. But maybe we can figure this out in the next PR that will actually set up `Documenter.jl`.

BenjaminBorn:
Thanks @ararslan and @nalimilan for the thorough review. I have hopefully addressed all points now. Once this PR is merged, I will open a new PR to build the docs.

src/fisher.jl
```diff
@@ -71,7 +71,8 @@ immutable FisherExactTest <: HypothesisTest
 end
 
 testname(::FisherExactTest) = "Fisher's exact test"
-population_param_of_interest(x::FisherExactTest) = ("Odds ratio", 1.0, x.ω) # parameter of interest: name, value under h0, point estimate
+population_param_of_interest(x::FisherExactTest) = ("Odds ratio", 1.0, x.ω)
```
nalimilan: Better not change this since that's unrelated. Or at least put the comment before the function rather than after. Same below.

BenjaminBorn: Ok. I have reverted it.

BenjaminBorn:
Done. Let me know if I should do something else.

bjarthur:
this is great!  but...  why is this package not being tested on julia 0.6?

ararslan:
Will be. Once this is merged we should drop 0.5 support, test on 0.6, and have FemtoCleaner do a deprecation fixing pass.

src/fisher.jl
```diff
@@ -59,6 +86,37 @@ function show_params(io::IO, x::FisherExactTest, ident="")
...
 # DOC: for tail=:both there exist multiple ``method``s for computing a pvalue and the corresponding ci.
+"""
+    pvalue(x::FisherExactTest; tail = :both, method = :central)
+
+Compute the p-value for a given Fisher exact test.
+
+The one-sided p-values are based on Fisher's non-central hypergeometric distribution
+``f_ω(i)`` with odds ratio ``ω``:
+```math
+    \\begin{align}
```
ararslan: I guess if you're moving everything to `align*` this should be as well? Aside from that, this all LGTM.

BenjaminBorn: Thanks, fixed.

src/fisher.jl
```diff
@@ -24,6 +24,33 @@
...
+
+The contingency table is structured as:
+
+| -  | X1 | X2 |
+|:--:|:--:|:--:|
+|*Y1*| a  | b  |
+|*Y2*| c  | d  |
+
+!!! note
+    The [`Base.show`](@ref) output contains the conditional maximum likelihood estimate of the odds ratio
```
ararslan: When you build the docs, what does this end up linking to?

BenjaminBorn: As noted above

> The Documenter.jl documentation states
"*Note that depending on what the CurrentModule is set to, a docstring @ref may need to be prefixed by the module which defines it.*"
> However, at least in my local build, links like [`StatsBase.counts`](@ref) don't seem to work. But maybe we can figure this out in the next PR that will actually set up Documenter.jl.

So, at least for now the links to outside modules don't work. Those are:

- ``[`StatsBase.counts`](@ref)`` and  `` [`Base.show`](@ref)`` for which the error is that `!! No doc found for reference`
- ``[`Rmath.pwilcox`](@ref)`` and ``[`Rmath.psignrank`](@ref)`` for which it says `ERROR: UndefVarError(:Rmath)`, so this is maybe not the correct module call.

The Documenter.jl documentation unfortunately doesn't give an example how to link to other modules and I also couldn't find an example in other packages.

Besides these dead links, the docs are still built correctly and everything looks ok. If you have an idea on how to fix it, I can push the fix here. If it takes us longer to figure it out, we might want to merge here and then fix the links in the follow-up PR.

ararslan: I'd just remove the dead ref links for now, e.g. in this case just say "The `show` output contains..."

BenjaminBorn: Done. But you explain this to @nalimilan ;-)

ararslan: Did Milan tell you to link to things that can't be properly linked...?

BenjaminBorn: No no, of course not. I was joking. He told me to include the links but wasn't sure whether it would work. Let's leave them out and we can figure the linking out later.

BenjaminBorn:
Thanks Alex. Sorry for the back and forth.

ararslan:
No problem at all! That's just how PR reviews go. You've done awesome work here, as usual.

ararslan:
@nalimilan Good to merge? Any further cleanup that's necessary can always happen later.

src/fisher.jl
nalimilan: Unfortunately it seems it's not possible at the moment: https://github.com/JuliaDocs/Documenter.jl/issues/425. Let's go without links for now.

nalimilan:
Sorry for the delay. It's fine with me! Though conflicts have to be fixed first.

ararslan:
> Though conflicts have to be fixed first.

Weird, I'm not seeing any merge conflicts?

PR #108, opened 2017-08-26
Drop 0.5 and macOS testing

ararslan:
Once this is merged, FemtoCleaner should jump in and fix everything.

I've removed macOS from the Travis testing matrix since its queue is ridiculous and we don't have any platform-specific code in this package.


ararslan:
Nightly failures are due to StaticArrays via ForwardDiff via Roots.

PR #109, opened 2017-08-27
Documenter setup

BenjaminBorn:
This PR sets up `Documenter.jl`, moves documentation from `.rst` to `.md` files, and deletes the `doc` folder.

The documentation has been successfully built locally. I've not done this before so please let me know if I have to do additional steps to get this built in the GitHub repository.

Fixes https://github.com/JuliaStats/HypothesisTests.jl/issues/59, https://github.com/JuliaStats/HypothesisTests.jl/issues/103.


ararslan:
Looks good from a cursory glance but I'll take a closer look soon. Thanks again for all of your help here!

docs/src/nonparametric.md
```diff
@@ -0,0 +1,70 @@
+# Nonparametric tests
+
+## Anderson-Darling test
+
+The null hypothesis of the Anderson–Darling test is that a dataset comes from a certain
```
nalimilan: Why describe this test and not the others? Isn't that redundant with the docstrings?

docs/src/methods.md
```diff
@@ -0,0 +1,15 @@
+# Methods
+
+## Confidence interval
+```@docs
+confint
+confint(::BinomialTest)
```
nalimilan: I don't think you need to list the specific methods. IIRC only methods from this package will appear.

.travis.yml
```diff
@@ -2,9 +2,8 @@
 language: julia
 os:
   - linux
-  - osx
```
nalimilan: Why drop Mac OS and Julia 0.5?

ararslan: I did in #108, so if this PR is rebased from master, that change will no longer show up in the diff. The Travis Mac queue is too long to be useful, especially since we don't have any platform-specific code in this package. Plus it was just time for 0.5 to be laid to rest so that we can fix the swarm of 0.7 deprecations.

docs/src/methods.md
BenjaminBorn: Actually, I do have to list the specific methods as we defined the general docstring on a zero-method function. If I just list `confint`, it will not include the docstrings for the specific methods in the documentation.

I actually like it this way as it gives us flexibility. We could potentially rearrange the method section or move the specific confidence intervals or p-values to the respective tests.

Note that in the REPL, `?confint` still gives you all methods without having to specify them in the help call.

docs/src/nonparametric.md
BenjaminBorn: I copied this one-to-one from the `.rst` file. Should I delete it?

.travis.yml
BenjaminBorn: Change no longer shows up after rebase.

BenjaminBorn:
I added a link to the (not yet existing) documentation in the readme file.

docs/src/nonparametric.md
nalimilan: I would say yes: either that's completely redundant with the docstring, or the missing pieces should be moved to the docstring.

docs/src/methods.md
nalimilan: OK, I always forget how this works.

docs/src/nonparametric.md
nalimilan: And same remark for Kolmogorov–Smirnov. To nuance my position, I guess it would be fine to keep a sentence here if it describes what is common to the two AD tests and says a word about their differences (something which isn't always very natural to put in docstrings). But that's not a requirement, and there shouldn't be too many repetitions with the docstrings.

BenjaminBorn: Ok, I have replaced them by one-sentence introductions.

README.md
```diff
@@ -35,4 +35,6 @@ pvalue(SignedRankTest(x))
 
 ## Documentation
 
-Full documentation available at [Read the Docs](http://hypothesistestsjl.readthedocs.org/en/latest/index.html).
+Full documentation available at
```
nalimilan: While you're at it, better adopt the standard layout used by other packages. See for example the one for StatsBase: https://raw.githubusercontent.com/JuliaStats/StatsBase.jl/master/README.md

BenjaminBorn: Done.

BenjaminBorn:
Thanks for the review!

README.md
```diff
@@ -1,12 +1,25 @@
...
 
-[![Build Status](https://travis-ci.org/JuliaStats/HypothesisTests.jl.svg?branch=master)](https://travis-ci.org/JuliaStats/HypothesisTests.jl)
-[![Coverage Status](https://coveralls.io/repos/JuliaStats/HypothesisTests.jl/badge.svg?branch=master)](https://coveralls.io/r/JuliaStats/HypothesisTests.jl?branch=master)
-[![HypothesisTests](http://pkg.julialang.org/badges/HypothesisTests_0.5.svg)](http://pkg.julialang.org/?pkg=HypothesisTests)
-[![HypothesisTests](http://pkg.julialang.org/badges/HypothesisTests_0.6.svg)](http://pkg.julialang.org/?pkg=HypothesisTests)
+*HypothesisTests.jl* is a Julia package that implements a wide range of hypothesis tests.
 
-This package implements several hypothesis tests in Julia.
+- **Current Release**:
+  [![HypothesisTests](http://pkg.julialang.org/badges/HypothesisTests.5.svg)
```
ararslan: Since we no longer support 0.5 as of #108, this badge could be dropped. This is more of a note-to-self; you don't have to do that here.

ararslan: Well, that should probably wait anyway until we have a 0.6-only release.

BenjaminBorn:
Thanks Alex. One question, the link to "stable" documentation is dead as there is only a "latest" documentation. Do we have to do anything or will this be filled automatically later on?

ararslan:
The stable docs are populated based on releases. So once we tag a release, Travis will run on the tag, which Documenter detects and then builds and deploys the stable docs. The latest docs are built and deployed on each commit to master.

ararslan:
The docs look _amazing_, by the way. Thanks again!

nalimilan:
Indeed, thanks! Though, if you're looking for a possible improvement, it would be even better to list the contents on the main page, like in StatsBase. That would be particularly useful to directly give a list of supported tests.

BenjaminBorn:
Thanks to both of you for all the help! I will go through the docs again and make a PR with improvements (including the table of contents).

Issue #39, opened 2015-09-09
Float32 error with MannWhitneyUTest


bjarthur:
it works fine for Float64 and all integer types.

'''
julia> using HypothesisTests

julia> MannWhitneyUTest(rand(Float64,5),rand(Float64,8))
Exact Mann-Whitney U test
-------------------------
Population details:
    parameter of interest:   Location parameter (pseudomedian)
    value under h_0:         0
    point estimate:          -0.26900372866519906

Test summary:
    outcome with 95% confidence: fail to reject h_0
    two-sided p-value:           0.3543123543123543 (not signficant)

Details:
    number of observations in each group: [5,8]
    Mann-Whitney-U statistic:             13.0
    rank sums:                            [28.0,63.0]
    adjustment for ties:                  0.0

julia> MannWhitneyUTest(rand(Float32,5),rand(Float32,8))
ERROR: `ExactMannWhitneyUTest{T<:Real}` has no method matching ExactMannWhitneyUTest{T<:Real}(::Float64, ::Array{Float64,1}, ::Float64, ::Int64, ::Int64, ::Float32)

julia> Pkg.installed("HypothesisTests")
v"0.2.9"

julia> VERSION
v"0.3.9"
'''

PR #113, opened 2017-09-06
fix ApproximateMannWhitney for Float32

bjarthur:
this mirrors the fix for ExactMannWhitney in https://github.com/JuliaStats/HypothesisTests.jl/issues/39

### DIFF
--- a/src/mann_whitney.jl
+++ b/src/mann_whitney.jl
@@ -199,8 +199,8 @@ where ``\\mathcal{T}`` is the set of the counts of tied values at each tied posi
 
 Implements: [`pvalue`](@ref)
 """
-function ApproximateMannWhitneyUTest(U::Float64, ranks::AbstractVector{Float64},
-    tie_adjustment::Float64, nx::Int, ny::Int, median::Real)
+function ApproximateMannWhitneyUTest(U::Real, ranks::AbstractVector{T},
+    tie_adjustment::Real, nx::Int, ny::Int, median::Real) where T<:Real
     mu = U - nx * ny / 2
     sigma = sqrt((nx * ny * (nx + ny + 1 - tie_adjustment /
         ((nx + ny) * (nx + ny - 1)))) / 12)
--- a/src/mann_whitney.jl
+++ b/src/mann_whitney.jl
@@ -200,7 +200,7 @@ where ``\\mathcal{T}`` is the set of the counts of tied values at each tied posi
 Implements: [`pvalue`](@ref)
 """
 function ApproximateMannWhitneyUTest(U::Float64, ranks::AbstractVector{Float64},
-    tie_adjustment::Float64, nx::Int, ny::Int, median::T) where T<:Real
+    tie_adjustment::Float64, nx::Int, ny::Int, median::Real)
     mu = U - nx * ny / 2
     sigma = sqrt((nx * ny * (nx + ny + 1 - tie_adjustment /
         ((nx + ny) * (nx + ny - 1)))) / 12)
--- a/src/mann_whitney.jl
+++ b/src/mann_whitney.jl
@@ -165,11 +165,11 @@ end
 
 struct ApproximateMannWhitneyUTest{T<:Real} <: HypothesisTest
     U::Float64              # test statistic: Mann-Whitney-U statistic
-    ranks::Vector{T}        # ranks
+    ranks::Vector{Float64}  # ranks
     tie_adjustment::Float64 # adjustment for ties
     nx::Int                 # number of observations
     ny::Int
-    median::Float64         # sample median
+    median::T               # sample median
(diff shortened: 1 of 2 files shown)


andreasnoack:
It seems a bit odd to make the `struct` parametric when t effectively promotes to `Float64`. @simonster why did you make the change in https://github.com/JuliaStats/HypothesisTests.jl/commit/1c2d2c452c0f2385861f7c48ca23e6ef70cc29be the way you did?

bjarthur:
@nalimilan `ranks` is always Float64 because it's defined to be [here](https://github.com/JuliaStats/HypothesisTests.jl/blob/master/src/common.jl#L62)

@andreasnoack `median()` retains the type so the median field needs to be the same type as the input args

julia nightly is failing on the anderson_darling testset, which is unrelated.  these tests seem to be non-determinstic, as locally they pass sometimes and fail others.

can you please review again?  i think it's ready to merge.

bjarthur:
bump.  @nalimilan approved previously, and i've just resolved the conflicts with master.  can we please merge?  thanks.

src/mann_whitney.jl
```diff
@@ -199,8 +199,8 @@ where ``\\mathcal{T}`` is the set of the counts of tied values at each tied posi
 
 Implements: [`pvalue`](@ref)
 """
-function ApproximateMannWhitneyUTest(U::Real, ranks::AbstractVector{T},
-    tie_adjustment::Real, nx::Int, ny::Int, median::Float64) where T<:Real
+function ApproximateMannWhitneyUTest(U::Float64, ranks::AbstractVector{Float64},
+    tie_adjustment::Float64, nx::Int, ny::Int, median::T) where T<:Real
```
nalimilan, line 203:
```suggestion
    tie_adjustment::Float64, nx::Int, ny::Int, median::Real)
```

nalimilan (approved):

bjarthur:
the only CI test that fails is anderson-darling for macOS 1.3.  this test passes for me locally on my mac using julia v1.3.  not sure what the difference is.  moreover, not sure how the changes proposed in this PR relate to anderson-darling.  lastly, why do we support 1.3 in HypothesisTests?  1.6 is the current LTS, and locally it passes there as well. could we just merge this PR?

src/mann_whitney.jl
```diff
@@ -199,8 +199,8 @@ where ``\\mathcal{T}`` is the set of the counts of tied values at each tied posi
 
 Implements: [`pvalue`](@ref)
 """
-function ApproximateMannWhitneyUTest(U::Real, ranks::AbstractVector{T},
-    tie_adjustment::Real, nx::Int, ny::Int, median::Float64) where T<:Real
+function ApproximateMannWhitneyUTest(U::Float64, ranks::AbstractVector{Float64},
+    tie_adjustment::Float64, nx::Int, ny::Int, median::Real)
```
ararslan, line 203: The changes to the type restrictions could conceivably be breaking for anyone calling this method directly. AFAICT they shouldn't be needed since the inner constructor will `convert` its arguments to the types of the fields, which permits a larger variety of inputs than `Float64`.

test/mann_whitney.jl
```diff
@@ -54,6 +54,10 @@ end
     @test abs(@inferred(pvalue(ExactMannWhitneyUTest(Float32[1:10;], Float32[2:11;]))) - 0.5096) <= 1e-4
 end
 
+@testset "Issue #113" begin
+  @test abs(pvalue(ApproximateMannWhitneyUTest(Float32[1:10;], Float32[2:11;])) - 0.4948) <= 1e-4
```
ararslan, line 58:
```suggestion
    @test abs(pvalue(ApproximateMannWhitneyUTest(Float32[1:10;], Float32[2:11;])) - 0.4948) <= 1e-4
```

ararslan:
> the only CI test that fails is anderson-darling for macOS 1.3. this test passes for me locally on my mac using julia v1.3. not sure what the difference is. moreover, not sure how the changes proposed in this PR relate to anderson-darling.

Probably a flaky test, definitely unrelated to the PR

> why do we support 1.3 in HypothesisTests?

No one has submitted a PR to change the compatibility

ararslan:
> i don't understand why the changes to type restrictions are needed either, but the test in this PR for Float32 fails without them

I checked out your branch, changed the method signature back to what it was, and ran tests, and they passed for me locally. 🤔

ararslan:
Actually when I made the change back to what it was, I missed one particular parameter, which solves the mystery. 😄 Only the `mean::Float64` -> `median::Real` change is needed for the method signature. I've put that in a suggestion.

ararslan (approved):
Thanks for the contribution and for your patience!

Issue #50, opened 2016-04-18
Documentation for p-value in FisherExactTest


andreasnoack:
We use a different method than R. I think we use 2x"the smaller tail" whereas R sums all values less than or equal to the reference p-value. It would be great to have it documented. Maybe we also want to use R's method.

simonster:
I think this is documented [here](http://hypothesistestsjl.readthedocs.org/en/latest/api/pvalue.html#pvalue-fisher) and the R method is available as `method=:minlike` but the documentation could probably be reorganized so this is easier to find.

andreasnoack:
Ah thanks. Perfectly documented. I agree with you that since I didn't think of clicking on the `pvalue` link, it might be possible to make it more obvious that additional information is available there.

BenjaminBorn:
Can this be closed? 

The link to the p-value documentation is now quite prominent in the [FisherExactTest documentation](http://juliastats.github.io/HypothesisTests.jl/stable/nonparametric.html#Fisher-exact-test-1) and it can be seen without clicking on the link that a specific method for the FisherExactTest exists.

I'm also happy to add a sentence to the documentation if it's not clear enough.

nalimilan:
Maybe we could add more information directly on the `FisherExactTest` docstring, since by default a p-value will be printed without even calling `p-value`? Possibly just something like "see [`p-value(::FisherExactTest)`](@ref) for details about the computation of the default p-value".

PR #116, opened 2017-10-16
Adjust FisherExactTest documentation

BenjaminBorn:
Make people aware that p-value and confidence interval are non-standard and link the information.

See #50.


ararslan:
Nightly error is because we don't have nightly binaries right now.

Issue #117, opened 2017-10-17
Restrict printed test output to 3 or 4 decimal places


BenjaminBorn:
Right now, the test output is printed with 17 decimal places. I would propose to only print 3 or 4.

What would be the best approach? Could I just use `round` on the floating output in the definition of show in HypothesisTests.jl?

BenjaminBorn:
Addressed in #119.

Issue #67, opened 2016-07-25
Use StatsBase.confint instead of ci


ararslan:
Currently HypothesisTests defines a function `ci` for computing confidence intervals. However, it seems to me that we should be using `confint` from StatsBase and extending it for each test. Is there a reason why we should continue to prefer `ci` or would it make sense to switch over to `confint`?

andreasnoack:
They should definitely use the same name. I'll vote for `confint`.

ararslan:
Agreed. It's certainly a more descriptive name than `ci`, and folks using this package that are familiar with other packages that *do* extend `confint` would likely be confused if we don't conform.

PR #119, opened 2017-10-31
Make printing of p-values pretty

BenjaminBorn:
Addresses the p-value part of #117.

Open questions:

1. I also wanted to restrict the confidence interval to four digits, however I ran into problems. My idea was to use elementwise round such that
`println(io, "    95% confidence interval: $(round.(StatsBase.confint(test), 4))")`
However, the confidence interval for the `PowerDivergenceTest` is saved as a `Vector{Tuple{Float64,Float64}}` and `round.()` does not work with with a vector of tuples. Is there a more elegant way of doing it or should I write a loop over the elements of the vector?

2. There are also a number of other outputs printed that should be restricted in their number of decimal digits shown. However, there is a lot of variation between tests so it would make more sense to adjust them at the test level. I'm happy to go through the tests but wanted to make sure that rounding the printed output to four digits is the preferred approach.


BenjaminBorn:
I have now also rounded the confidence interval in the printed output to four decimal places. The approach has been discussed on [Discourse](https://discourse.julialang.org/t/elementwise-operation-on-array-of-tuples/6812).

I think point (2) can be dealt with later.

src/HypothesisTests.jl
```diff
@@ -111,7 +111,8 @@ function Base.show(io::IO, test::T) where T<:HypothesisTest
     println(io, "    value under h_0:         $param_under_h0")
     println(io, "    point estimate:          $param_estimate")
     if has_ci
-        println(io, "    95% confidence interval: $(StatsBase.confint(test))")
+        println(io,
+            "    95% confidence interval: $((x -> round.(x, 4)).(StatsBase.confint(test)))")
```
nalimilan: I'd rather save the intermediate result to make the expression less complex. Also, as I said on Discourse, I think `map` is more standard for this, anonymous functions with dot syntax look weird.

src/HypothesisTests.jl
```diff
@@ -122,11 +123,11 @@ function Base.show(io::IO, test::T) where T<:HypothesisTest
     println(io, "Test summary:")
     println(io, "    outcome with 95% confidence: $outcome h_0")
     if tail == :both
-        println(io, "    two-sided p-value:           $p")
+        println(io, "    two-sided p-value:           $(StatsBase.PValue(p))")
```
nalimilan: Here too it would make sense to store the `PValue` in an object so that you don't need to call this for each case.

BenjaminBorn: I have a more fundamental question. Couldn't we, as has been done for the confidence interval (see https://github.com/JuliaStats/HypothesisTests.jl/issues/67), use StatsBase.PValue throughout and extend it for our tests?

nalimilan: Well, `PValue` is mostly useful for printing AFAIK. It could be annoying to return such an object to users, as it doesn't support any operation defined on numbers.

BenjaminBorn: I just realised this as well.

BenjaminBorn:
I have added two tests but wasn't sure how to test the confidence interval printing.

src/HypothesisTests.jl
```diff
@@ -111,22 +111,25 @@ function Base.show(io::IO, test::T) where T<:HypothesisTest
     println(io, "    value under h_0:         $param_under_h0")
     println(io, "    point estimate:          $param_estimate")
     if has_ci
-        println(io, "    95% confidence interval: $(StatsBase.confint(test))")
+        ci = map(x -> round.(x, 4), StatsBase.confint(test))
```
nalimilan: No need for the dot, right? Also no need for a line break below.

test/power_divergence.jl
```diff
@@ -26,6 +26,7 @@ for i = 1:length(c)
 end
 
 @test pvalue(m) ≈ 2.9535891832117357e-7
+@test "$(StatsBase.PValue(pvalue(m)))" == "<1e-6"
```
nalimilan: The point isn't to test StatsBase, it's to test the result of `sprint(show, m)`. That should allow you to test both the p-value and the confidence interval at the same time. Use triple-quoted strings to hold the expected output. BTW, I think this should go to a separate file, e.g. `show.jl`, since there's no point in testing the output for all types as long as the code is common.

src/HypothesisTests.jl
BenjaminBorn: Thanks. But it doesn't work without the dot
`MethodError: no method matching round(::Tuple{Float64,Float64}, ::Int64)`

test/power_divergence.jl
BenjaminBorn: Sorry, I didn't know how to get the output from show. Using `sprint(show, m)` dumps the whole output into a string. What is the most elegant way of extracting, e.g., the confidence interval? I don't know what you mean by "Use triple-quoted strings to hold the expected output". Could you give me a link to an example?

BenjaminBorn: Ok, I think I have now understood what you meant and included three tests in a new file show.jl.

One last question: I tried including line breaks in the output of `sprint(show, m)` but then the tests would fail. Is there a way of not having all of the output in one enormous line?

nalimilan: You can just introduce normal line breaks, just like in a copy/paste of the output. Any indentation will automatically be removed by comparing with the position of triple quotes on the first line.

src/HypothesisTests.jl
nalimilan: Sorry, I hadn't realized there could be multiple CIs.

Something I noticed in the tested outputs is that `0.3` is not printed as `0.300`, which would be better to make it clear that there are three digits of precision and to ease the comparison with other values. You can achieve that with `@sprintf("%.3f", x)`.

BenjaminBorn: Unfortunately, that doesn't seem to work on the array of tuples:
```
using HypothesisTests
d = [[762,484] [327,239] [468,477]]
m = PowerDivergenceTest(d)
ci = map(x -> round.(x, 4), StatsBase.confint(m))
@sprintf("%.4f", ci)

MethodError: no method matching isfinite(::Array{Tuple{Float64,Float64},1})
```

But I probably don't understand what you have in mind.

nalimilan: Yes, you'll have to use something like `map(x -> @sprintf("%.4f", x), ci)` just like you did for `round`. BTW, no need to call `round` beforehand, so you can just replace that call.

test/show.jl
```diff
@@ -0,0 +1,73 @@
...
+    one-sided p-value:           <1e-6
+
+Details:
+    Sample size:        2757
+    statistic:          30.070149095754687
+    degrees of freedom: 2
+    residuals:          [2.19886, -2.50467, 0.41137, -0.468583, -2.84324, 3.23867]
+    std. residuals:     [4.50205, -4.50205, 0.699452, -0.699452, -5.31595, 5.31595]
+"""
+show(IOBuffer(), m)
```
nalimilan: This isn't needed, `sprint` does the same thing. Same below.

test/show.jl
```diff
@@ -0,0 +1,73 @@
+using HypothesisTests, Base.Test
+
+# based on power_divergence.jl tests
+d = [[762,484] [327,239] [468,477]]
+m = PowerDivergenceTest(d)
+
+@test sprint(show, m) == """Pearson's Chi-square Test
+-------------------------
```
nalimilan: I think the most common convention is to put `"""` on the second line, with four spaces of indentation (just like for any other code), and then align all others lines to `Pearson`. That way it's clear that the code line continues, and the output appears exactly as it's printed on screen.

test/show.jl
```diff
@@ -0,0 +1,73 @@
...
+Population details:
+    parameter of interest:   Multinomial Probabilities
+    value under h_0:         [0.255231, 0.19671, 0.11594, 0.0893561, 0.193574, 0.14919]
+    point estimate:          [0.276387, 0.175553, 0.118607, 0.0866884, 0.16975, 0.173014]
+    95% confidence interval: Tuple{Float64,Float64}[(0.2579, 0.2955), (0.1571, 0.1947), (0.1001, 0.1378), (0.0682, 0.1058), (0.1513, 0.1889), (0.1545, 0.1922)]
```
nalimilan: While you're at it, if you want to make more changes in another PR, it looks like this output could be improved by turning it into a table similar to `StatsBase.CoefTable` (probably using that type), with one parameter per row, and one column for the value under h_0, on for the point estimate, and two for the CI. It would also make sense to stop printing `Tuple{Float64, Float64}`. That would only apply to the situation where there are several parameters to estimate of course.

test/show.jl
```diff
@@ -0,0 +1,73 @@
...
+
+Test summary:
+    outcome with 95% confidence: reject h_0
+    one-sided p-value:           <1e-6
+
+Details:
+    Sample size:        2757
+    statistic:          30.070149095754687
```
nalimilan: This one would also benefit from rounding.

src/HypothesisTests.jl
BenjaminBorn: Yeah, I tried that yesterday but unfortunately that still doesn't work with the array of tuples.
```
using HypothesisTests
d = [[762,484] [327,239] [468,477]]
m = PowerDivergenceTest(d)
ci = StatsBase.confint(m)
map(x -> @sprintf("%.4f", x), ci)

MethodError: no method matching isfinite(::Tuple{Float64,Float64})
```

Do I need to loop through the array?

test/show.jl
BenjaminBorn: Yes. See point 2 in my opening comment. The problem is that the output in the "Details" section varies from test to test. So I think I need to adjust this at the individual test level. I think that's another PR.

src/HypothesisTests.jl
nalimilan: Ah, yes, you'll need two nested calls to `map`.

BenjaminBorn: It also works with a comprehension
```
ci = [ map(x -> @sprintf("%.4f", x), y) for y in StatsBase.confint(m) ]
```
Unfortunately, ``println()`` then produces
```
95% confidence interval: Tuple{String,String}[("0.2579", "0.2955"), ("0.1571", "0.1947"), ("0.1001", "0.1378"), ("0.0682", "0.1058"), ("0.1513", "0.1889"), ("0.1545", "0.1922")]
```
So it's now an array of tuples of strings (albeit rounded nicely). Would you prefer this or should I go back to the just rounded output for now? Following your suggestion below, I think it makes sense to restructure this in a table anyways.

nalimilan: Ah, right, you'd need to print the parts of the tuple manually. Let's keep `round` for now and improve this later by using a table instead.

BenjaminBorn: Perfect. Thanks for your help here!

BenjaminBorn:
Any other comments?

PR #120, opened 2017-10-31
Fix deprecation

BenjaminBorn:
Came across this when running the test suite. 

Btw., are these mini-PRs welcome or are they too annoying for reviewers?


ararslan:
> are these mini-PRs welcome or are they too annoying for reviewers?

They're very welcome. Thanks for the contribution!

PR #132, opened 2018-08-30
Updates for Julia 1.0

ararslan:
Drop Compat, bump minimum Julia version to 0.7, deploy docs from 1.0


ararslan:
Looks like Distributions is broken on 1.0.

PR #136, opened 2018-10-01
Change of API of UnequalVarianceTTest

pevnak:
Hi,

I have extended API of `UnequalVarianceTTest`, such that user can either use the old one providing lists of values of individual measurements or their mean and variance. I needed this extension, because I am calculating statistics online. This extension simplifies my code and make it more readable.
Thanks for considering this.


wildart:
LGTM

nalimilan:
Why not, but the docstring should give the meaning of the arguments in the second signature, and we should apply this to `EqualVarianceTTest` too. Also there's no need to force mean and var to be of the same type. Finally, adding a test would be good to ensure we don't treat this as a purely internal function.

PR #138, opened 2018-10-15
fix binom test edge case

bkamins:
This PR fixes implementation of Clopper-Pearson `BinomialTest` which currently fails in edge cases (number of successes equal to `0` or number of trials).

I did not correct the other tail to account for presence of edge case (this is what is also done in R).

I additionally cleaned up the layout of test cases (now `SignTest` is in separate testset).

### DIFF
--- a/src/binomial.jl
+++ b/src/binomial.jl
@@ -130,7 +130,8 @@ end
 
 # Clopper-Pearson interval (confidence interval by inversion)
 function ci_clopper_pearson(x::BinomialTest, alpha::Float64=0.05)
-    (quantile(Beta(x.x, x.n - x.x + 1), alpha/2), quantile(Beta(x.x + 1, x.n - x.x), 1-alpha/2))
+    (x.x == 0 ? 0.0 : quantile(Beta(x.x, x.n - x.x + 1), alpha/2),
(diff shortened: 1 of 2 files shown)


src/binomial.jl
```diff
@@ -130,7 +130,8 @@ end
 
 # Clopper-Pearson interval (confidence interval by inversion)
 function ci_clopper_pearson(x::BinomialTest, alpha::Float64=0.05)
-    (quantile(Beta(x.x, x.n - x.x + 1), alpha/2), quantile(Beta(x.x + 1, x.n - x.x), 1-alpha/2))
+    (x.x == 0 ? 0.0 : quantile(Beta(x.x, x.n - x.x + 1), alpha/2),
```
ararslan: This means that the output values will always be `Float64`. Perhaps we should be more clever with types somehow? Or will the beta quantile always be `Float64`?

bkamins: Given current implementation it will be always `Float64` as `BinomialTest` is not parametric so we always call `quantile(Beta(::Int, ::Int), ::Float64)` which even if `quantile` is reimplemented should return `Float64` (currently no matter what input `quantile` would return `Float64` as it calls a wrapper of R function - however, I guess, it could change in the future).

Also `@code_warntype` goes through OK for `pvalue` function here.

ararslan: Good enough for me

Issue #142, opened 2019-01-08
Wrong and slow CI for PowerDivergenceTest with :sison_glaz method


bkamins:
Here is an example of the problem:
```
julia> M = [10^6, 10^6]
2-element Array{Int64,1}:
 1000000
 1000000

julia> @time confint(ChisqTest(M))
  6.237456 seconds (35 allocations: 6.734 KiB)
2-element Array{Tuple{Float64,Float64},1}:
 (0.0, 1.0)
 (0.0, 1.0)
```

Not only it is very slow but also `95% confidence interval: Tuple{Float64,Float64}[(0.0, 1.0), (0.0, 1.0)]` is wrong IMHO.

Compare e.g.:
```
julia> @time confint(ChisqTest(M), method=:gold)
  0.000028 seconds (40 allocations: 6.500 KiB)
2-element Array{Tuple{Float64,Float64},1}:
 (0.49911667776081814, 0.5008833222391819)
 (0.49911667776081814, 0.5008833222391819)

julia> @time confint(ChisqTest(M), method=:bootstrap)
  0.002813 seconds (33.14 k allocations: 1.353 MiB)
2-element Array{Tuple{Float64,Float64},1}:
 (0.49931748750000005, 0.5007065375)
 (0.4992934625, 0.5006825125000001)

julia> @time confint(ChisqTest(M), method=:quesenberry_hurst)
  0.000031 seconds (36 allocations: 6.375 KiB)
2-element Array{Tuple{Float64,Float64},1}:
 (0.49920754546127, 0.50079245453873)
 (0.49920754546127, 0.50079245453873)
```

PR #143, opened 2019-01-17
automatic selection of CI method for PowerDivergenceTest

bkamins:
Fixes https://github.com/JuliaStats/HypothesisTests.jl/issues/142.

I have pushed this PR, but I am not sure who is maintaining HypothesisTests.jl for a review.


src/power_divergence.jl
```diff
@@ -79,6 +81,9 @@ function StatsBase.confint(x::PowerDivergenceTest, alpha::Float64=0.05;
         i = StatsBase.confint(x, alpha*2,method=method, GC=GC)
         Tuple{Float64,Float64}[(i[j][1], 1.0) for j in 1:m]
     elseif tail == :both
+        if method == :auto
+            method = minimum(x.expected) > 100 ? :quesenberry_hurst : :sison_glaz
+        end
         if method == :gold
```
nalimilan: `elseif`?

test/power_divergence.jl
```diff
@@ -158,6 +174,8 @@ confint(m, method = :sison_glaz, tail=:right)
 @test_throws ArgumentError confint(m, method=:FOO)
 @test_throws ArgumentError confint(m, tail=:fox)
 
+@test confint(m, method = :sison_glaz) == confint(m, method = :auto) == confint(m)
```
nalimilan: Also test a case where the other method is used?

bkamins: It is tested by the other set of tests above.

src/power_divergence.jl
bkamins: No it is not `elseif` - I update the `method` variable value if needed and reuse it later in the other `if`.

bkamins:
In R you have https://cran.r-project.org/web/packages/MultinomialCI/MultinomialCI.pdf. It uses the Sison and Glaz method and gives correct results even for large values and is faster than our implementation.

That is why in general this PR is only a work-around as it is clear that our implementation of Sison and Glaz has a bug or could be improved.

That is why I have asked who is maintaining HypothesisTests.jl because what I propose is only a temporary solution to the problem.

bkamins:
I have already did some time ago https://github.com/JuliaStats/HypothesisTests.jl/issues/142.

bkamins:
Does JuliaStats team want to close this or merge (just cleaning up the status of my old open PRs 😄)?

src/power_divergence.jl
```diff
@@ -44,12 +44,14 @@ default_tail(test::PowerDivergenceTest) = :right
...
 
 """
-    confint(test::PowerDivergenceTest, alpha = 0.05; tail = :both, method = :sison_glaz)
+    confint(test::PowerDivergenceTest, alpha = 0.05; tail = :both, method = :auto)
 
 Compute a confidence interval with coverage 1-`alpha` for multinomial proportions using
 one of the following methods. Possible values for `method` are:
 
-  - `:sison_glaz` (default): Sison-Glaz intervals
+  - `:auto` (default): Sison-Glaz intervals unless the minimum of expected cell counts
```
ararslan: The wording here is a bit confusing. I'd say
```suggestion
  - `:auto` (default): If the minimum of the expected cell counts exceeds 100, Sison-Glaz
    intervals are used, otherwise Quesenberry-Hurst.
```

src/power_divergence.jl
```diff
@@ -44,12 +44,14 @@ default_tail(test::PowerDivergenceTest) = :right
...
+     is greater than 100 when Quesenberry-Hurst invervals are selected
```
ararslan:
```suggestion
```

src/power_divergence.jl
```diff
@@ -44,12 +44,14 @@ default_tail(test::PowerDivergenceTest) = :right
...
 
 """
-    confint(test::PowerDivergenceTest, alpha = 0.05; tail = :both, method = :sison_glaz)
+    confint(test::PowerDivergenceTest, alpha = 0.05; tail = :both, method = :auto)
 
 Compute a confidence interval with coverage 1-`alpha` for multinomial proportions using
 one of the following methods. Possible values for `method` are:
 
-  - `:sison_glaz` (default): Sison-Glaz intervals
+  - `:auto` (default): If the minimum of the expected cell counts exceeds 100, Sison-Glaz
```
bkamins:
```suggestion
  - `:auto` (default): If the minimum of the expected cell counts exceeds 100, Quesenberry-Hurst
```

src/power_divergence.jl
```diff
@@ -44,12 +44,14 @@ default_tail(test::PowerDivergenceTest) = :right
...
+    intervals are used, otherwise Quesenberry-Hurst.
```
bkamins:
```suggestion
    intervals are used, otherwise Sison-Glaz.
```

bkamins:
Thank you for the fix. I have amended it slightly (the rule was the other way around: Sison-Glaz is used for small counts)

nalimilan:
OK, let's go with that then!

PR #145, opened 2019-01-26
Add Hotelling's T^2 test

ararslan:
Imported from https://github.com/ararslan/MultivariateTests.jl, updated for the modern era (originally written for 0.6). The test values were validated against SAS when I wrote it.


src/common.jl
```diff
@@ -57,3 +57,11 @@ function tiedrank_adj!(ord::AbstractVector, v::AbstractArray)
 end
 
 tiedrank_adj(v::AbstractArray) = tiedrank_adj!(Vector{Float64}(undef, length(v)), v)
+
+# Pool covariance matrices, overwriting the first
+function poolcov!(Sx::AbstractMatrix, nxm1::Int, Sy::AbstractMatrix, nym1::Int)
+    @inbounds for i = eachindex(Sx, Sy)
```
nalimilan: Can write this using dot syntax nowadays.

nalimilan: Too bad.

The PR looks good to me, but I know so little about this that I don't dare approving.

ararslan:
It matches SAS and examples from a textbook so I'd say let's go for it. If something is wrong, people will let us know.

PR #146, opened 2019-01-26
Add partial correlation test

ararslan:
NOTE: This will have merge conflicts with #145.

Imported from https://github.com/ararslan/MultivariateTests.jl. Requires StatsBase 0.27.0, which introduced `partialcor`, also taken from MultivariateTests. Test values validated against SAS when originally written.


mschauer:
I tried to compare with https://github.com/mschauer/CausalInference.jl/blob/f354ca8ce9052293023f7ccefda299bf946dd977/src/skeleton.jl#L105, but could not match up turns, does the atan belong there somewhere or do I have it wrong?

ararslan:
> but could not match up turns

I'm not familiar with that terminology. Can you elaborate?

mschauer:
Apparently I meant to write: match up the *terms* in both expressions

ararslan:
Oh, haha, that makes more sense. I don't know about the use of trig functions there; my original implementation was based on the book and notes from a course on multivariate statistics. The online version of the notes that mentions this test is here: https://newonlinecourses.science.psu.edu/stat505/node/47/. They just formulate the test statistic basically like a regular correlation.

simonster:
`(log1p(test.r) - log1p(-test.r)) / 2` is the same as `atanh(test.r)` if that helps.

ararslan:
If you can think of a test case in which such catastrophic cancelation could occur, I'd be happy to add it.

mschauer:
If you consider writing a test I would rather replace by the difference by `atanh`, because such tests have already been written for `atanh`. I checked
```
f(r) = (log1p(r) - log1p(-r)) / 2
julia> using Test

julia> @test f(-1+eps()) == atanh(-1+eps())
Test Passed
```
Again I don't think anything goes wrong, I just wanted to report I did due diligence.

nalimilan:
@ararslan Do you want to finish this? Looks like it's quite close.

ararslan:
I think I was blocked on coming up with an appropriate test as Moritz mentioned.

mschauer:
I think this would do

ararslan:
You mean the PR as-is?

ararslan:
Rebased.

src/partialcor.jl
```diff
@@ -0,0 +1,61 @@
...
+"""
+    PartialCorTest(x, y, Z)
+
+Perform a t-test for the hypothesis that ``\\text{Cor}(x,y|Z=z) = 0``, i.e. the partial
+correlation of vectors `x` and `y` given the matrix `Z` is zero.
+
+Implements `pvalue` and `confint`. See also [`partialcor`](@ref).
+"""
+struct PartialCorTest <: HypothesisTest
+    r::Real
```
nalimilan: I think I've asked this before, but shouldn't it be a type parameter?

mschauer: It should

src/partialcor.jl
```diff
@@ -0,0 +1,61 @@
...
+    n::Int
+    k::Int
+    t::Real
+
+    # Error checking is done in `partialcor`
+    function PartialCorTest(x::AbstractVector, y::AbstractVector, Z::AbstractMatrix)
+        r = partialcor(x, y, Z)
```
mschauer: Is `partialcor` guaranteed in `[0,1]`?

src/partialcor.jl
```diff
@@ -0,0 +1,61 @@
...
+testname(::PartialCorTest) = "Test for partial correlation"
+population_param_of_interest(p::PartialCorTest) = ("Partial correlation", zero(p.r), p.r)
+
+StatsBase.nobs(p::PartialCorTest) = p.n
+StatsBase.dof(p::PartialCorTest) = p.n - 2 - p.k
+
+function StatsBase.confint(test::PartialCorTest, alpha::Float64=0.05)
+    dof(test) > 1 || return (-one(test.r), one(test.r)) # Otherwise we can get NaNs
+    q = quantile(Normal(), 1 - alpha / 2)
+    fisher = (log1p(test.r) - log1p(-test.r)) / 2
```
mschauer:
```suggestion
    fisher = atanh(test.r)
```

src/partialcor.jl
```diff
@@ -0,0 +1,61 @@
...
+
+testname(::PartialCorTest) = "Test for partial correlation"
+population_param_of_interest(p::PartialCorTest) = ("Partial correlation", zero(p.r), p.r)
+
+StatsBase.nobs(p::PartialCorTest) = p.n
+StatsBase.dof(p::PartialCorTest) = p.n - 2 - p.k
+
+function StatsBase.confint(test::PartialCorTest, alpha::Float64=0.05)
+    dof(test) > 1 || return (-one(test.r), one(test.r)) # Otherwise we can get NaNs
+    q = quantile(Normal(), 1 - alpha / 2)
```
mschauer: t-test should go with t based confidence interval?

src/partialcor.jl
```diff
@@ -0,0 +1,61 @@
...
+    fisher = (log1p(test.r) - log1p(-test.r)) / 2
+    bound = q / sqrt(test.n - 3 - test.k)
+    lo = 2 * (fisher - bound)
+    hi = 2 * (fisher + bound)
+    elo = Statistics.clampcor(expm1(lo) / (exp(lo) + 1))
```
mschauer:
```suggestion
    elo = Statistics.clampcor(tanh(fisher - bound)) 
```
This makes it much clearer that this is the inverse transformation

src/partialcor.jl
```diff
@@ -0,0 +1,61 @@
...
+    ehi = Statistics.clampcor(expm1(hi) / (exp(hi) + 1))
+    return (elo, ehi)
+end
+
+default_tail(::PartialCorTest) = :both
+pvalue(test::PartialCorTest; tail=:both) = pvalue(TDist(dof(test)), test.t, tail=tail)
```
mschauer: Or here: Normal

src/partialcor.jl
```diff
@@ -0,0 +1,61 @@
...
+        r = partialcor(x, y, Z)
+        n, k = size(Z)
+        t = r * sqrt((n - 2 - k) / (1 - r^2))
+        return new(r, n, k, t)
+    end
+
+    function PartialCorTest(x::AbstractVector, y::AbstractVector, z::AbstractVector)
+        r = partialcor(x, y, z)
+        n = length(z)
+        t = r * sqrt((n - 3) / (1 - r^2))
```
mschauer: Shouldn't the test be based on the fisher-z transform?

src/partialcor.jl
ararslan: Yep. As implemented in StatsBase, it returns the result of `clampcor`.

ararslan:
Looks like there are some numerical differences >0.001.

mschauer:
I understand this know, the test and the CI are unrelated... the test is exact using a different test statistic which is (only under the null hypothesis rho=0) t distributed and the CI is approximate using fishers z as statistic which is approximately normal under a range of partial correlation values (this is the story for correlation estimation and it seems that this translates to partial correlations.)

ararslan:
That's vaguely what I had thought but I didn't remember well enough to say definitively.

src/partialcor.jl
mschauer: Resolved by comment

src/partialcor.jl
```diff
@@ -0,0 +1,62 @@
...
+
+export PartialCorTest
+
+"""
+    PartialCorTest(x, y, Z)
+
+Perform a t-test for the hypothesis that ``\\text{Cor}(x,y|Z=z) = 0``, i.e. the partial
+correlation of vectors `x` and `y` given the matrix `Z` is zero.
+
+Implements `pvalue` and `confint`. See also [`partialcor`](@ref).
```
mschauer:
```suggestion
Additionally implements an approximative confidence interval `confint` using Fisher's z-transform. 

See also [`partialcor`](@ref).


# External ressources

* [Partial correlation on Wikipedia](https://en.wikipedia.org/wiki/Partial_correlation#As_conditional_independence_test) (for the construction of the confidence interval)

*[Section Testing using Student's t-distribution from Pearson correlation coefficient on Wikipedia](https://en.wikipedia.org/wiki/Pearson_correlation_coefficient#Testing_using_Student's_t-distribution)

* [K.J. Levy and S.C. Narula (1978): Testing Hypotheses concerning Partial Correlations: Some Methods and Discussion. International Statistical Review 46(2).](https://www.jstor.org/stable/1402814)

```

src/partialcor.jl
```diff
@@ -0,0 +1,62 @@
+# Partial correlation
+
+export PartialCorTest
+
+"""
+    PartialCorTest(x, y, Z)
+
+Perform a t-test for the hypothesis that ``\\text{Cor}(x,y|Z=z) = 0``, i.e. the partial
+correlation of vectors `x` and `y` given the matrix `Z` is zero.
```
mschauer:
```suggestion
correlation of vectors `x` and `y` given the matrix `Z` is zero and implements `pvalue`.
```

src/partialcor.jl
```diff
@@ -0,0 +1,62 @@
...
+StatsBase.nobs(p::PartialCorTest) = p.n
+StatsBase.dof(p::PartialCorTest) = p.n - 2 - p.k
+
+function StatsBase.confint(test::PartialCorTest{T}, alpha::Float64=0.05) where T
+    df = dof(test)
+    df > 1 || return (-one(T), one(T))  # Otherwise we can get NaNs
+    q = quantile(TDist(df), 1 - alpha / 2)
+    fisher = atanh(test.r)
+    bound = q / sqrt(test.n - 3 - test.k)
+    lo = 2 * (fisher - bound)
```
mschauer:
```suggestion
```

src/partialcor.jl
```diff
@@ -0,0 +1,62 @@
...
+    hi = 2 * (fisher + bound)
```
mschauer:
```suggestion
```

src/partialcor.jl
```diff
@@ -0,0 +1,62 @@
...
+testname(::PartialCorTest) = "Test for partial correlation"
+population_param_of_interest(p::PartialCorTest) = ("Partial correlation", zero(p.r), p.r)
+
+StatsBase.nobs(p::PartialCorTest) = p.n
+StatsBase.dof(p::PartialCorTest) = p.n - 2 - p.k
+
+function StatsBase.confint(test::PartialCorTest{T}, alpha::Float64=0.05) where T
+    df = dof(test)
+    df > 1 || return (-one(T), one(T))  # Otherwise we can get NaNs
+    q = quantile(TDist(df), 1 - alpha / 2)
```
mschauer:
```suggestion
    q = quantile(Normal(), 1 - alpha / 2)
```
Let's go back to Normal quantiles for the confidence interval as the more common choice.

mschauer:
I'll merge. One question: this is the right thing to do if you want to do a correlation tests. Should we define an alias emphasising  testing for (non-partial) Pearson correlation coefficients?

ararslan:
Hm, we could define it instead as `CorrelationTest` where the two-argument method performs a normal correlation test and the three-argument method does the partial correlation test.

src/partialcor.jl
```diff
@@ -0,0 +1,72 @@
...
+population_param_of_interest(p::PartialCorTest) = ("Partial correlation", zero(p.r), p.r)
+
+StatsBase.nobs(p::PartialCorTest) = p.n
+StatsBase.dof(p::PartialCorTest) = p.n - 2 - p.k
+
+function StatsBase.confint(test::PartialCorTest{T}, alpha::Float64=0.05) where T
+    dof(test) > 1 || return (-one(T), one(T))  # Otherwise we can get NaNs
+    q = quantile(Normal(), 1 - alpha / 2)
+    fisher = atanh(test.r)
+    bound = q / sqrt(test.n - 3 - test.k)
```
ararslan:
```suggestion
    bound = q / sqrt(dof(test) - 1)
```

ararslan:
I'm not sure when I'll have time to make further changes to this PR, unfortunately.

ararslan:
Rebased and renamed to `CorrelationTest`. The two-argument method that performs the "regular" correlation test is not implemented here.

mschauer:
@aguang We will merge this without, but then it should be fairly straight forward to implement the two argument version on top of it. @ararslan Thank you for bearing with me, the princess is not in another castle this time.

Issue #150, opened 2019-02-21
We should support getting standard error of difference for two-sample t-tests


pdeffebach:
Inspired by [this tweet](https://twitter.com/amandayagan/status/1098314654470819840), which highlights that it is difficult to get the standard error of a two-sample t test using R's `t.test` function:

pdeffebach:
We should also support the magnitude of differences between the two means.

nalimilan:
More developments on the R side: https://stat.ethz.ch/pipermail/r-devel/2019-February/077335.html

PR #151, opened 2019-02-21
Add stderror and meandiff functions for TTest

pdeffebach:
Fixes issue #150 by adding two small functions for `TTest` abstract types. 

There are probably many more types we can define this for if we want. 

It's clear that a `getproperty` give the user what they need, but it's good practice to wrap all of these in a function.


nalimilan:
I guess it makes sense to use `stderror` since it already exists, but is it really worth exporting a new `meandiff` function for that?

A few other ideas:
- print the standard error in the test output?
- add it to the docstring
- add tests

pdeffebach:
Fair enough about the `meandiff`. 

Should I go ahead and make all of these changes for all functions that have `confint` defined? If you have a confidence interval, surely you have a standard error.

nalimilan:
> Should I go ahead and make all of these changes for all functions that have `confint` defined? If you have a confidence interval, surely you have a standard error.

If it makes sense for those tests, yes.

pdeffebach:
Now that I'm on winter break I will work on this more. I will start by making a list of all the tests and whether or not `stderror` is well defined.

pdeffebach:
To determine which tests should be updated I searched for `confint` and the following tests have `confint` implemented:

* t-test: Implemented easily
* z: Implemented easily
* Binomial: Tough because there are multiple methods of calculating it. It's not clear to me whether the standard error is just `sqrt(p * (1-p) / n)` for everything and the confidence intervals are calculated differently for each test. 
* Partial Correlation test: Wikipedia tells me the standard error is just [sqrt(1 / dof(test)](https://en.wikipedia.org/wiki/Pearson_correlation_coefficient#Testing_using_Student's_t-distribution). So this is implemented. 
* Fisher's exact test: I think the [Fisher's Hon-central Hypergeometric geometric distribution](https://en.wikipedia.org/wiki/Fisher%27s_noncentral_hypergeometric_distribution) has a well-defined  standard deviation available in `Distributions`. 
* Power Divergence Test: Googling tells me that this statistic doesn't have a standard error associated with it. 
* Signed Rank Test: As far as I can tell the standard error is not commonly used for this test.

pdeffebach:
I totally forgot about this pull request for over a year, but I just added tests. I think this is good to be merged. 

I didn't add `stderror` for Fischer's exact test because I don't understand it enough. I can always add it later.

nalimilan:
Thanks. There should probably be documentation for these methods, just like what we have for `confint` and `pvalue`?

pdeffebach:
Thanks! added tests for the empty function and then edited all the `implements:...` docstrings in the docstrings for the types.

pdeffebach:
This PR can be merged I think. Sorry this has languished.

PR #152, opened 2019-02-22
Wald-Wolfowitz (Runs) Test

mileslucas:
I was inspired to add the equivalent of R's `lawstat::runs.test`. Following the original R code as well as the wikipedia entry for the Wald-Wolfowitz test, I believe I've implemented an accurate Hypothesis Test. This is non-parametric and I've added a line in the docs under the correct markdown file for it. In addition, I've added a couple of testsets. I was only able to get the tests to run (and pass) with the line `using .HypothesisTests` in `test/wald_wolfowitz.jl` but after I made sure the tests passed I maintained the convention in the other testing files.

Note: I am very new to Julia, but I believe I have implemented this simple Hypothesis Test appropriately. I am very open to feedback as well as any documentation suggestions, or anything else.


src/wald_wolfowitz.jl
```diff
@@ -0,0 +1,57 @@
...
+Performs the Wald-Wolfowitz (or Runs) test to determine
+randommness for a data-sequence. The null hypothesis is that the data is random, 
+or independent.
+
+Implements: [`pvalue`](@ref)
+"""
+function WaldWolfowitzTest(x::AbstractVector{T}) where T<:Real
+    n = length(x)
+    med = median(x)
+    num_above = count(x .>= med)
```
nalimilan: Use this to avoid an allocation:
```suggestion
    num_above = count(v -> v >= med, x)
```
Same below.

(You could also use a single loop over `x` an increment both variables, maybe it would be slightly faster.)

src/wald_wolfowitz.jl
```diff
@@ -0,0 +1,57 @@
...
+    num_below = count(x .< med)
+
+    # Get the expected value and standard deviation
+    μ = 1 + 2 * num_above * num_below / n
```
nalimilan: Something like this will prevent overflow with long vectors (on 32-bit mainly):
```suggestion
    μ = 1 + 2 * num_above * (num_below / n)
```

src/wald_wolfowitz.jl
```diff
@@ -0,0 +1,57 @@
...
+    σ = sqrt((μ - 1) * (μ - 2) / (n - 1))
+
+    # Get the number of runs
+    signs = sign.(x .- med)
```
nalimilan: Better compute this on the fly to avoid another allocation.

test/wald_wolfowitz.jl
```diff
@@ -0,0 +1,37 @@
...
+
+@testset "Wald-Wolfowitz test" begin
+
+@testset "Independent Observations" begin
+    # Get set of independent observations
+    x = randn(1000)
+    tst = WaldWolfowitzTest(x)
+
+    # Should not be significant dependence
+    @test pvalue(tst) > 0.05
```
nalimilan: This is too broad. Better hardcode the exact result for a given input vector and check it with another implementation. Then you don't need separate tests for independent and dependent observations. However it can make sense to test different inputs, in particular a boolean array and an array with only equal values.

Also test other functions implemented by the test (see what other tests do).

src/wald_wolfowitz.jl
```diff
@@ -0,0 +1,57 @@
+export 
+    WaldWolfowitzTest
+
+struct WaldWolfowitzTest <: HypothesisTest
+    N_above::Int
+    N_below::Int
+    N_runs::Int
+    μ::Real
```
nalimilan: This should be a type parameter to ensure type stability. One parameter is probably enough to all three fields. See how other tests handle it.

src/wald_wolfowitz.jl
```diff
@@ -0,0 +1,57 @@
...
+default_tail(::WaldWolfowitzTest) = :both
+
+
+function show_params(io::IO, x::WaldWolfowitzTest, ident="")
+    println(io, ident, "number of runs:  $(x.N_runs)")
+    println(io, ident, "z-statistic:     $(x.z)")
+end
+
+"""
+    WaldWolfowitzTest(x::AbstractVector{T<:Real})
```
nalimilan:
```suggestion
    WaldWolfowitzTest(x::AbstractVector{<:Real})
```

src/wald_wolfowitz.jl
```diff
@@ -0,0 +1,57 @@
+export 
+    WaldWolfowitzTest
+
+struct WaldWolfowitzTest <: HypothesisTest
+    N_above::Int
```
nalimilan: `nabove` and so on would probably be more standard.

src/wald_wolfowitz.jl
```diff
@@ -0,0 +1,57 @@
...
+
+function show_params(io::IO, x::WaldWolfowitzTest, ident="")
+    println(io, ident, "number of runs:  $(x.N_runs)")
+    println(io, ident, "z-statistic:     $(x.z)")
+end
+
+"""
+    WaldWolfowitzTest(x::AbstractVector{T<:Real})
+
+Performs the Wald-Wolfowitz (or Runs) test to determine
```
nalimilan: AFAICT the convention is to mention the null hypothesis in the first sentence (see manual).

src/wald_wolfowitz.jl
```diff
@@ -0,0 +1,57 @@
...
+randommness for a data-sequence. The null hypothesis is that the data is random,
```
nalimilan: Wikipedia describes the test as applying to two-valued data sequences, so it would be useful to explain that values are split at the median (which as a special case works when there are only two values).

ararslan:
Welcome, and thanks for the cool contribution!

You mentioned that you "follow[ed] the original R code." Note that there could be licensing concerns, as this package is MIT-licensed, but GPL is more common in the R world, and the GPL requires derived works to be distributed under GPL as well. Do you know the license for the R code you used as a reference?

nalimilan:
Given how short the actual algorithm is, and that my comments imply modifying each line, I don't think the end result will qualify as derivative work. :-) What remains is really the core of the math, which isn't copyrightable.

mileslucas:
I would tend to agree with @nalimilan , but regardless the original code is from [lawstat](https://cran.r-project.org/web/packages/lawstat/index.html). The only "derivative" part of this was the method of determining the runs (using the signs of the difference with the median). 

I brought the package up because my Stats professor is using this `lawstat::runs.test` along with another one as part of hypothesis testing for linear regression assumptions, but I'm trying to use Julia for the course and saw an easy opportunity to contribute. Theoretically, this could have been implemented without ever referencing R!

src/wald_wolfowitz.jl
mileslucas: I don't understand your comments, unfortunately. In other files, I see the use of `abstract type ____ <: HypothesisTest` but I see no need to have an intermediate abstract type rather than a struct.

ararslan:
Sounds good to me. :slightly_smiling_face:

src/wald_wolfowitz.jl
nalimilan: I mean doing `μ::T` and `struct WaldWolfowitzTest{T<:Real} <: HypothesisTest`.

mileslucas:
These recent pushes resolve most of the above suggestions/issues, feel free to take another look around.

I've actually changed my method of determining the number of runs @ararslan so there should be absolutely no issues, in my opinion, with license. I've created a function for `TwoValuedWaldWolfowitzTest`, which is technically the de facto and the method I implemented previously is a special case, however, as far as I know, the many-valued approach is more common and should be considered the default for statistical use. I make use of it, however, by taking the numerical vector and transforming it into a bitarray so I properly encapsulate and reuse the methods.

Any more comments about documentation or standards is very welcome, thanks for the help.

src/wald_wolfowitz.jl
mileslucas: Thanks for the clarification, will work on this now

src/wald_wolfowitz.jl
```diff
@@ -0,0 +1,68 @@
...
+
+function show_params(io::IO, x::WaldWolfowitzTest, ident="")
+    println(io, ident, "number of runs:  $(x.nruns)")
+    println(io, ident, "z-statistic:     $(x.z)")
+end
+
+"""
+    WaldWolfowitzTest(x::AbstractVector{<:Real})
+
+Performs the Wald-Wolfowitz (or Runs) test of the null hypothesis that the given data is random, or independent.
```
nalimilan: Just a matter of style, but can we really say that "data is independent"? The Wikipedia page says that *elements* are independent, which makes more sense to me.

mileslucas: Does "independently sampled" seem more accurate?

src/wald_wolfowitz.jl
```diff
@@ -0,0 +1,68 @@
...
+    println(io, ident, "number of runs:  $(x.nruns)")
+    println(io, ident, "z-statistic:     $(x.z)")
+end
+
+"""
+    WaldWolfowitzTest(x::AbstractVector{<:Bool})
+    WaldWolfowitzTest(x::AbstractVector{<:Real})
+
+Performs the Wald-Wolfowitz (or Runs) test of the null hypothesis that the given data is random, or independently sampled.
+The data can come as many-valued or two-valued (boolean). If many-valued, the sample is transfromed by labelling each
```
nalimilan:
```suggestion
The data can come as many-valued or two-valued (Boolean). If many-valued, the sample is transformed by labelling each
```

src/wald_wolfowitz.jl
```diff
@@ -0,0 +1,68 @@
...
+pvalue(test::WaldWolfowitzTest; tail=:both) = pvalue(Normal(0.0, 1.0), test.z; tail=tail)
+
+
+function show_params(io::IO, x::WaldWolfowitzTest, ident="")
+    println(io, ident, "number of runs:  $(x.nruns)")
+    println(io, ident, "z-statistic:     $(x.z)")
+end
+
+"""
+    WaldWolfowitzTest(x::AbstractVector{<:Bool})
```
nalimilan:
```suggestion
    WaldWolfowitzTest(x::AbstractVector{Bool})
```

src/wald_wolfowitz.jl
```diff
@@ -0,0 +1,68 @@
...
+    WaldWolfowitzTest(x::AbstractVector{<:Real})
+
+Performs the Wald-Wolfowitz (or Runs) test of the null hypothesis that the given data is random, or independently sampled.
+The data can come as many-valued or two-valued (boolean). If many-valued, the sample is transfromed by labelling each
+element as above or below the median.
+
+Implements: [`pvalue`](@ref)
+"""
+function WaldWolfowitzTest(x::AbstractVector{T}) where T<:Bool
```
nalimilan:
```suggestion
function WaldWolfowitzTest(x::AbstractVector{Bool})
```

src/wald_wolfowitz.jl
```diff
@@ -0,0 +1,68 @@
...
+        end
+    end
+
+    # calculate simple z-statistic
+    z = (nruns - μ) / σ
+    WaldWolfowitzTest(nabove, nbelow, nruns, μ, σ, z)
+
+end
+
+function WaldWolfowitzTest(x::AbstractVector{T}) where T<:Real
```
nalimilan:
```suggestion
function WaldWolfowitzTest(x::AbstractVector{<:Real})
```

src/wald_wolfowitz.jl
```diff
@@ -0,0 +1,68 @@
...
+    med = median(x)
+    transformed = x .>= med
+    WaldWolfowitzTest(transformed)
+end
+
+"""
+Alias for: [`WaldWolfowitzTest`](@ref)
```
nalimilan: I don't think it's a good idea to define aliases.

test/wald_wolfowitz.jl
```diff
@@ -0,0 +1,47 @@
...
+@testset "Wald-Wolfowitz test" begin
+
+@testset "Many-Valued Observations" begin
+    # Get set of dependent observations
+    x = 1:1000
+    tst = WaldWolfowitzTest(x)
+    
+    # Should have significant dependence
+    @test tst.z ≈ -31.575 atol=1e-3
+    @test pvalue(tst) < 2.2e-16
```
nalimilan: Also check exact value. Same below.

test/wald_wolfowitz.jl
```diff
@@ -0,0 +1,47 @@
...
+
+    # Test consistency of z-statistics
```
nalimilan: This isn't really useful if you test the exact value above.

test/wald_wolfowitz.jl
```diff
@@ -0,0 +1,47 @@
...
+using Distributions
+
+@testset "Wald-Wolfowitz test" begin
+
+@testset "Many-Valued Observations" begin
+    # Get set of dependent observations
+    x = 1:1000
+    tst = WaldWolfowitzTest(x)
+    
+    # Should have significant dependence
```
nalimilan: Remove this comment, it makes it sound like you don't test the exact value.

src/wald_wolfowitz.jl
nalimilan: "elements are independently sampled" should be OK I guess.

src/wald_wolfowitz.jl
mileslucas: In my testing, the array `x .>=med` is of type `bitmask` so just saying `Bool` did not work.

nalimilan: Not it's a `BitArray` but that's OK. `Bool` is a concrete type anyway, it doesn't have subtypes.

mileslucas:
Long time no see! Latest push addresses most recent comments, including removing `RunsTest` as an alias, updating the tests to use exact values, and minor documentation and typing changes.

mileslucas:
Is there any update on this?

src/wald_wolfowitz.jl
```diff
@@ -0,0 +1,60 @@
...
+    for k in 1:(n-1)
+        if x[k] != x[k+1]
+            nruns += 1
+        end
+    end
+
+    # calculate simple z-statistic
+    z = (nruns - μ) / σ
+    WaldWolfowitzTest(nabove, nbelow, nruns, μ, σ, z)
+
```
nalimilan:
```suggestion
```

src/wald_wolfowitz.jl
```diff
@@ -0,0 +1,60 @@
...
+    nabove = sum(x)
+    nbelow = n - nabove
+    
+    # Get the expected value and standard deviation
+    μ = 1 + 2 * nabove * (nbelow / n)
+    σ = sqrt((μ - 1) * (μ - 2) / (n - 1))
+
+    # Get the number of runs
+    nruns = 1
+    for k in 1:(n-1)
```
nalimilan:
```suggestion
    @inbounds for k in 1:(n-1)
```

src/wald_wolfowitz.jl
```diff
@@ -0,0 +1,60 @@
...
+        end
+    end
+
+    # calculate simple z-statistic
+    z = (nruns - μ) / σ
+    WaldWolfowitzTest(nabove, nbelow, nruns, μ, σ, z)
+
+end
+
+function WaldWolfowitzTest(x::AbstractVector{<:Real})
```
nalimilan: This could just be a one-liner:
```julia
WaldWolfowitzTest(x::AbstractVector{<:Real}) = WaldWolfowitzTest(x .>= median(x))
```

test/wald_wolfowitz.jl
```diff
@@ -0,0 +1,37 @@
...
+    tst = WaldWolfowitzTest(x)
+    
+    @test tst.z ≈ -31.575 atol=1e-3
+    @test pvalue(tst) ≈ 8.052e-219 atol=1e-222
+
+    # Test consistency of z-statistics
+    @test pvalue(tst) == pvalue(Normal(tst.μ, tst.σ), tst.nruns)
+    @test pvalue(tst, tail=:left) == pvalue(Normal(tst.μ, tst.σ), tst.nruns, tail=:left)
+    @test pvalue(tst, tail=:right) == pvalue(Normal(tst.μ, tst.σ), tst.nruns, tail=:right)
+    show(IOBuffer(), tst)
```
nalimilan: Can you test the output? `sprint(show, tst)` will return it as a string.

src/wald_wolfowitz.jl
```diff
@@ -0,0 +1,60 @@
...
+function show_params(io::IO, x::WaldWolfowitzTest, ident="")
+    println(io, ident, "number of runs:  $(x.nruns)")
+    println(io, ident, "z-statistic:     $(x.z)")
+end
+
+"""
+    WaldWolfowitzTest(x::AbstractVector{Bool})
+    WaldWolfowitzTest(x::AbstractVector{<:Real})
+
+Performs the Wald-Wolfowitz (or Runs) test of the null hypothesis that the given data is random, or independently sampled.
```
nalimilan:
```suggestion
Perform the Wald-Wolfowitz (or Runs) test of the null hypothesis that the given data is random, or independently sampled.
```

src/wald_wolfowitz.jl
```diff
@@ -0,0 +1,60 @@
...
+The data can come as many-valued or two-valued (Boolean). If many-valued, the sample is transfromed by labelling each
```
nalimilan:
```suggestion
The data can come as many-valued or two-valued (Boolean). If many-valued, the sample is transformed by labelling each
```
Please also cut lines at 92 characters.

mileslucas:
merged the current master and implemented above changes

src/wald_wolfowitz.jl
```diff
@@ -12,10 +12,10 @@ end
 testname(::WaldWolfowitzTest) = "Wald-Wolfowitz Test"
 population_param_of_interest(x::WaldWolfowitzTest) = ("Number of runs", x.μ, x.nruns) # parameter of interest: name, value under h0, point estimate
 default_tail(::WaldWolfowitzTest) = :both
-pvalue(test::WaldWolfowitzTest; tail=:both) = pvalue(Normal(0.0, 1.0), test.z; tail=tail)
+pvalue(test::WaldWolfowitzTest; tail = :both) = pvalue(Normal(0.0, 1.0), test.z; tail = tail)
```
nalimilan: Please revert the spacing changes: AFAICT we don't use spaces like this in other files.

nalimilan:
Looks like you need to rebase on master (not sure why your merge didn't work).

mileslucas:
I've rebased and double checked the spacing

nalimilan:
Sorry, it still seems that the rebase didn't work. Have you run `git pull` while on the `master` branch? You can also run `git fetch && git rebase origin/master`.

src/wald_wolfowitz.jl
```diff
@@ -0,0 +1,55 @@
...
+    z::T            # test z-statistic from Normal distribution
+end
+
+testname(::WaldWolfowitzTest) = "Wald-Wolfowitz Test"
+population_param_of_interest(x::WaldWolfowitzTest) = ("Number of runs", x.μ, x.nruns) # parameter of interest: name, value under h0, point estimate
+default_tail(::WaldWolfowitzTest) = :both
+pvalue(test::WaldWolfowitzTest; tail = :both) = pvalue(Normal(0.0, 1.0), test.z; tail = tail)
+
+
+function show_params(io::IO, x::WaldWolfowitzTest, ident = "")
```
nalimilan:
```suggestion
function show_params(io::IO, x::WaldWolfowitzTest, ident="")
```

src/wald_wolfowitz.jl
```diff
@@ -0,0 +1,55 @@
...
+    nruns::Int      # Number of runs
+    μ::T            # Expected mean
+    σ::T            # Expected variance
+    z::T            # test z-statistic from Normal distribution
+end
+
+testname(::WaldWolfowitzTest) = "Wald-Wolfowitz Test"
+population_param_of_interest(x::WaldWolfowitzTest) = ("Number of runs", x.μ, x.nruns) # parameter of interest: name, value under h0, point estimate
+default_tail(::WaldWolfowitzTest) = :both
+pvalue(test::WaldWolfowitzTest; tail = :both) = pvalue(Normal(0.0, 1.0), test.z; tail = tail)
```
nalimilan:
```suggestion
pvalue(test::WaldWolfowitzTest; tail=:both) = pvalue(Normal(0.0, 1.0), test.z; tail=tail)
```

mileslucas:
I don't understand what is wrong with the branches...I've merged/rebased of the juliastats/hypothesistests:origin branch and both times the pull request says there are no conflicts with the base branch. I don't see what the issue is.

nalimilan:
Not sure what happened, but one of the previous merges appears to have done something weird. When trying to rebase manually I get conflicts between commits from this PR, which is really weird, and then it fails. I've started from scratch, copying files manually. Please check that I didn't lose anything in the process before deleting your branch.

In general, better avoid working from the `master` branch of your fork, that's likely what caused the bug.

test/runtests.jl
```diff
@@ -32,4 +32,6 @@ include("t.jl")
 include("wilcoxon.jl")
 include("z.jl")
 include("hotelling.jl")
+include("wald_wolfowitz.jl")
```
nalimilan:
```suggestion
```

src/HypothesisTests.jl
```diff
@@ -178,5 +178,7 @@ include("jarque_bera.jl")
 include("durbin_watson.jl")
 include("permutation.jl")
 include("hotelling.jl")
+include("wald_wolfowitz.jl")
```
nalimilan:
```suggestion
```

mileslucas:
Ah, I can see how it got messed up, my mistake! I've removed those two unnecessary includes. Hopefully it should be good to go now.

mileslucas:
Man, sorry for all the trouble! Thanks for all the help and guidance!

mileslucas:
Bumping. I've re-merged the current master. My tests still pass, I've added version 0.9.0 to `Project.toml` based on the latest gh release being 0.8.0. Would be nice to see this merged!

nalimilan:
Sorry, I should have merged this before. The rebase seems to have gone wrong: maybe you rebased on your own fork instead of on master?

mileslucas:
Oh my, git issues, again. I most recently ran 
```
$ git pull js master
```
where `js` is this remote (`https://github.com/juliastats/hypothesistests.jl.git`) and I was on my master branch.

nalimilan:
Try `git reset --hard origin/master` from your master branch. Beware that it will destroy any changes you may have on that branch.

mileslucas:
> `git reset --hard origin/master` from your master branch. 

No effect!

mileslucas:
I'm going to close this in favor of #165. We won't have the explicit git history that this has, but to be honest it's not much code total and we have this PR to reference if the git history gods unleash a flood on us.

PR #161, opened 2019-07-25
Fix show methods to respect IOContext settings

rofinn:
I believe this should allow tests to pass on julia >= 1.2


ararslan:
Looks good to me! Seems to pass on nightly, which I guess is a good enough metric. We could also add 1.2 to Travis.

PR #164, opened 2019-08-20
Some love

Nosferican:
- Drop REQUIRE
- Use 1.2 flavor for Project.toml
- Update README (PkgEval)
- License for licensee detection
- Compat for StatsBase

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,8 @@ os:
   - linux
   - osx
   - windows
-  - nightly
 julia:
-  - 0.7
   - 1.0
-  - 1.1
   - 1.2
   - 1.3
 notifications:
@@ -19,13 +16,7 @@ jobs:
   include:
     - os: windows
       arch: x86
-      julia: 0.7
-    - os: windows
-      arch: x86
       julia: 1.0
-    - os: windows
-      arch: x86
-      julia: 1.1
     - os: windows
       arch: x86
       julia: 1.2
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,34 +1,42 @@
-# Documentation: http://docs.travis-ci.com/user/languages/julia/
 language: julia
-
 os:
   - linux
-
-julia:
-  - 1.2
+  - osx
+  - windows
   - nightly
-
-matrix:
-  allow_failures:
-    - julia: nightly
-
+julia:
+  - 0.7
+  - 1.0
+  - 1.1
+  - 1.2
+  - 1.3
 notifications:
   email: false
-
-# uncomment the following lines to override the default test script
-#script:
-#  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-#  - julia -e 'Pkg.clone(pwd()); Pkg.build("HypothesisTests"); Pkg.test("HypothesisTests"; coverage=true)'
-
-after_success:
-  - julia -e 'using Pkg; Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())';
-
 jobs:
+  fast_finish: true
+  allow_failures:
+    - julia: nightly
   include:
-    - stage: "Documentation"
+    - os: windows
+      arch: x86
+      julia: 0.7
+    - os: windows
+      arch: x86
+      julia: 1.0
+    - os: windows
+      arch: x86
+      julia: 1.1
+    - os: windows
+      arch: x86
       julia: 1.2
+    - os: windows
+      arch: x86
+      julia: 1.3
+    - stage: "Documentation"
       os: linux
+      julia: 1.2
       script:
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -5,9 +5,7 @@ os:
   - linux
 
 julia:
-  - 0.7
-  - 1.0
-  - 1.1
+  - 1.2
```
ararslan: I don't see the point in dropping support for earlier Julia versions

Nosferican: It is using `test/Project.toml`/`docs/Project.toml` which requires `julia = "1.2"` for the CI. Before that Pkg.jl only supports `extras`/`targets`. I could go back to the `extras` / `targets` if we want to keep testing those versions.

ararslan: Yes I think we should keep testing those versions

LICENSE.md
```diff
@@ -1,22 +1,9 @@
-HypothesisTests.jl is licensed under the MIT License:
```
nalimilan: This is the form used in Examples.jl. Are it needs to be changed? If so, it would be nice to update Example.jl.

test/Project.toml
```diff
@@ -0,0 +1,8 @@
+[deps]
```
nalimilan: Are you sure this is the recommended approach (again, based on Example.jl)?

docs/Project.toml
```diff
@@ -3,4 +3,4 @@ Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 HypothesisTests = "09f84164-cd44-5f33-b23f-e6b0d136a0d5"
 
 [compat]
-Documenter = "~0.20"
+Documenter = "0.23"
```
nalimilan: This is the recommended form: https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#.travis.yml-Configuration-1
```suggestion
Documenter = "~0.23"
```

LICENSE.md
Nosferican: Aye. It should be updated in Example.jl. The current one is not machine readable for licensee (the Ruby gem GitHub uses to detect licenses). Especially since the MIT licenses are a family, but the Open Source Initiative standardized the text.

I usually use the ISC which is the variant of the MIT license without language deemed unnecessary following the Berne Convention. Then I just copy the standardized text from the OSI website (that gets picked up by GitHub as the correct license).

test/Project.toml
Nosferican: https://julialang.github.io/Pkg.jl/dev/creating-packages/#Test-specific-dependencies-in-Julia-1.2-and-above-1 It is in the Pkg.jl documentation, but for Julia 1.2 onward.

docs/Project.toml
Nosferican: I am checking with Documenter.jl if they are in fact using patch versions for breaking changes during 0 major releases. If they are following the standard semver it should be the caret specifier (the default one).

.travis.yml
Nosferican: Should those be `0.7, 1.0, 1.1, 1.2, 1.3, nightly`?

> "Given that v1.1 is no more supported and superseded by v1.2, would it make sense to stop testing packages for v1.1? I'm not totally comfortable spawning dozens of CI jobs for each commit and pull request." - Mosè Giordano 
> "I would say it makes sense." - Kristoffer Carlsson
from Slack this morning, would it make sense to drop `0.7` and `1.1` then?

test/Project.toml
nalimilan: Interesting. Though this approach doesn't work on Julia 1.0 AFAICT, so we need to stick with the old one for now.

LICENSE.md
nalimilan: Good to know. Can you file a PR against Example.jl? Better ensure that package uses the same approach as here for consistency.

test/Project.toml
Nosferican: Aye. Ideally we could do some Travis logic to support the current system when Julia is the LTS à la Project/REQUIRE.

docs/Project.toml
nalimilan: Whatever the reason, I think we'd better do what the docs recommend. IIRC they indeed broke compatibility recently, so we can expect them to do so in the future too.

Nosferican: The docs have been updated to use the default semver caret instead of tilde (https://github.com/JuliaDocs/Documenter.jl/pull/1105).

.travis.yml
```diff
@@ -1,36 +1,33 @@
...
   - 1.0
-  - 1.1
-  - nightly
-
-matrix:
-  allow_failures:
-    - julia: nightly
-
+  - 1.2
+  - 1.3
```
Nosferican: Should it allow failures for 1.3?

PR #165, opened 2019-10-09
Waldwolf Wolfowitz Test

mileslucas:
This is just #152 except I care about the git this time. I've tested the code and docs- this should have a much cleaner git history (@nalimilan)


nalimilan:
Thanks!

PR #167, opened 2019-10-28
Adding homoscedasticity test

raphaelsaavedra:
Hi, I'm interested in contributing with a homoscedasticity test. The main reference is Durbin & Koopman, Time Series Analysis by State Space Methods, 2nd ed., chapter 2.12.1.

This is my first time using the package so please tell me if something is off. I tried to make it as similar as possible to how the other tests were implemented.


src/homoscedasticity.jl
```diff
@@ -0,0 +1,79 @@
+# homoscedasticity.jl
```
nalimilan: Please use the same header as other files.

src/homoscedasticity.jl
```diff
@@ -0,0 +1,79 @@
...
+  * James Durbin and Siem Jan Koopman, "Time Series Analysis by State Space Methods", 2nd Ed.,
+    2012, Oxford Statistical Science Series.
+
+# External links
+
+  * [Homoscedasticity on Wikipedia](https://en.wikipedia.org/wiki/Homoscedasticity)
+"""
+function HomoscedasticityTest(y::AbstractVector{T}, h::Int) where T <: Real
+    n = length(y)
+    @assert h <= n/2
```
nalimilan: Throw `ArgumentError` instead.

src/homoscedasticity.jl
```diff
@@ -0,0 +1,79 @@
...
+# References
+
+  * James Durbin and Siem Jan Koopman, "Time Series Analysis by State Space Methods", 2nd Ed.,
+    2012, Oxford Statistical Science Series.
+
+# External links
+
+  * [Homoscedasticity on Wikipedia](https://en.wikipedia.org/wiki/Homoscedasticity)
+"""
+function HomoscedasticityTest(y::AbstractVector{T}, h::Int) where T <: Real
```
nalimilan:
```suggestion
function HomoscedasticityTest(y::AbstractVector{<: Real}, h::Integer)
```

test/homoscedasticity.jl
```diff
@@ -0,0 +1,66 @@
...
+        0.741863, 0.145903, -0.450616, 0.384291, 0.0129242, -1.11626, 1.45973, -1.27855, -0.154354, 
+        1.67222, -0.685274, -0.742184, 0.488662, 0.805217, -1.22203, -1.11926, 1.29633, -0.415697, 
+        -0.319425, -1.14136, -0.59144, -0.350336, -0.426482, 0.275284, 1.58645]
+
+    t = HomoscedasticityTest(y_h0, 100)
+
+    @test t.n == 200
+    @test t.h == 100
+    @test t.H ≈ 1.115413 rtol = 1e-4
+    @test pvalue(t) ≈ 0.586011 rtol = 1e-4
```
nalimilan: Also test other tail variants.

raphaelsaavedra:
Hi @nalimilan,

Thanks for the review. Unfortunately, I don't have a proper name for the test. My reference is Durbing & Koopman and they only mention it as H test (name they gave to the test statistic). 

Screen Shot 2019-11-02 at 16 33 58

I will try to find out if the test has a proper name.

src/homoscedasticity.jl
raphaelsaavedra: By this, do you mean use a proper name for the test? I am not sure it has one but I will try to find out.

nalimilan: Sorry I hadn't realized files in this package use this header. We don't repeat the file name and the full license in other packages.

nalimilan:
It sounds weird that they would propose a new test with so little developments. It would be good to find a longer reference and a standard name.

raphaelsaavedra:
Well it's mostly because it's simply an F-test that is meant to be applied to residuals. If you find it acceptable, I can rename it to `HomoscedasticityFTest` or something like that.

mschauer:
It would be good to provide the F test independently from the application to time series and make the HomoscedasticityFTest test a thin wrapper. (I hope that is not too much feature creep)

raphaelsaavedra:
Very good point @mschauer. I am changing the struct to `FTest` so that we have a more general form that can be used to test the variances of two different samples and then `HomoscedasticityFTest` will simply be a particular case of `FTest`.

src/f.jl
```diff
@@ -0,0 +1,101 @@
...
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+# LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+# OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+# WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+export FTest, HomoscedasticityFTest
+
+struct FTest <: HypothesisTest
+    n1::Float64   # sample size of sample 1
```
nalimilan:
```suggestion
    n1::Float64   # size of sample 1
```

src/f.jl
```diff
@@ -0,0 +1,101 @@
...
+    n2::Float64   # sample size of sample 2
```
nalimilan:
```suggestion
    n2::Float64   # size of sample 2
```

src/f.jl
```diff
@@ -0,0 +1,101 @@
...
+Compute the test statistic of an F test: the null hypothesis is that two real-valued vectors
+`y1` and `y2` have equal variances.
+
+# References
+
+  * George E. P. Box, "Non-Normality and Tests on Variances", Biometrika 40 (3/4): 318–335, 1953.
+
+# External links
+
+  * [F-test on Wikipedia](https://en.wikipedia.org/wiki/F-test)
```
nalimilan:
```suggestion
  * [F-test of equality of variances on Wikipedia](https://en.wikipedia.org/wiki/F-test_of_equality_of_variances)
```

nalimilan:
There are several kinds of F-tests, as can be seen on Wikipedia. Better use a more specific name than `FTest` to avoid confusion with e.g. `ftest` in GLM. We currently have `EqualVarianceTTest`, but unfortunately "equal variance" there refers to the assumption of the t-test, not to the hypothesis which is tested. So maybe use just `VarianceFTest`? `EqualityOfVariancesFTest` would be more explicit but quite verbose.

Anyway, probably better list this test under "Parametric tests", and put only the homoscedasticity test under time series.

Regarding `HomoscedasticityFTest`, I don't think it's appropriate to have a function which looks like a constructor, but returns an object of a different type. It should either use a separate type, or make it a constructor of `VarianceFTest`. Not sure which approach is better. I guess the advantage of the latter is that it would allow printing the value of the `h` argument, and it would be simpler for the docs (list it with time series). The drawback is of course the increased complexity, but the code is quite short. What do you think?

src/f.jl
nalimilan: Can you copy the named of the fields from `EqualVarianceTTest`? Also use the same conventions for printing.

raphaelsaavedra:
I am making the requested changes. I agree with your point regarding `HomoscedasticityFTest`, but if we change it to a constructor of `VarianceFTest` I'm not sure how to present the different docstrings for methods of the same name under parametric and time series. Do you know how to do this?

nalimilan:
You can give the signature of each method instead of just its name. But I'm still not completely sure whether it's better to use a single type or two different types...

raphaelsaavedra:
I think having a single constructor for `VarianceFTest` is pretty intuitive. One method that receives two vectors and another one that receives one vector and the number of observations to consider in each "side".

nalimilan:
OK. Though I guess both could be listed under "Parametric tests", as it could apply to series other than time series.

raphaelsaavedra:
See if it's ok now :) don't hesitate to ping me to make additional changes

src/f.jl
```diff
@@ -0,0 +1,104 @@
...
+    2012, Oxford Statistical Science Series.
+
+# External links
+
+  * [Homoscedasticity on Wikipedia](https://en.wikipedia.org/wiki/Homoscedasticity)
+"""
+function VarianceFTest(y::AbstractVector{<: Real}, h::Integer)
+    n = length(y)
+    h > n/2 && throw(ArgumentError("The number of observations considered in each end must not be more than half the total number of observations"))
+    y1 = y[n-h+1:n]
```
nalimilan:
```suggestion
    y1 = view(y, n-h+1:n)
```

src/f.jl
```diff
@@ -0,0 +1,104 @@
...
+    y2 = y[1:h]
```
nalimilan:
```suggestion
    y2 = view(y, 1:h)
```

Also, why isn't this `y1`? That would sound more logical (especially for the printing).

test/f.jl
```diff
@@ -0,0 +1,110 @@
...
+        t = VarianceFTest(y1_h1, y2_h1)
+
+        @test t.n_x == 200
+        @test t.n_y == 120
+        @test t.df_x == 199
+        @test t.df_y == 119
+        @test t.F ≈ 0.36754 rtol = 1e-4
+        @test pvalue(t) < 1e-5
+        @test default_tail(t) == :both
+        @test pvalue(t; tail = :left) < 1e-5
```
nalimilan: Why choose this level exactly? Can this be made stricter? Same for next one and the following block.

src/f.jl
```diff
@@ -0,0 +1,104 @@
...
+function VarianceFTest(y1::AbstractVector{<: Real}, y2::AbstractVector{<: Real}; F0::Real = 1.0)
+    n1, n2 = length(y1), length(y2)
+    F = var(y1) / var(y2)
+    return VarianceFTest(n1, n2, n1-1, n2-1, F, F0)
+end
+
+"""
+    VarianceFTest(y::AbstractVector{<: Real}, h::Integer)
+
+Compute the test statistic of an homoscedasticity F test: the null hypothesis is that a real-valued
```
nalimilan:
```suggestion
Perform an homoscedasticity F-test of the null hypothesis that a real-valued
```

src/f.jl
```diff
@@ -0,0 +1,104 @@
...
+vector `y` has constant variance. The test is conducted over the ratio of the sample variances of the
+first `h` observations and the last `h` observations of the series.
+
```
nalimilan:
```suggestion

Implements: [`pvalue`](@ref)

```

src/f.jl
```diff
@@ -0,0 +1,104 @@
...
+    F::Real    # test statistic
+    F0::Real   # variance ratio under h_0
+end
+
+"""
+    VarianceFTest(y1::AbstractVector{<: Real}, y2::AbstractVector{<: Real})
+
+Compute the test statistic of an F test: the null hypothesis is that two real-valued vectors
+`y1` and `y2` have equal variances.
+
```
nalimilan:
```suggestion

Implements: [`pvalue`](@ref)

```

src/f.jl
```diff
@@ -0,0 +1,104 @@
...
+  * [F-test of equality of variances on Wikipedia](https://en.wikipedia.org/wiki/F-test_of_equality_of_variances)
+"""
+function VarianceFTest(y1::AbstractVector{<: Real}, y2::AbstractVector{<: Real}; F0::Real = 1.0)
+    n1, n2 = length(y1), length(y2)
+    F = var(y1) / var(y2)
+    return VarianceFTest(n1, n2, n1-1, n2-1, F, F0)
+end
+
+"""
+    VarianceFTest(y::AbstractVector{<: Real}, h::Integer)
```
nalimilan: Maybe this should be a keyword argument, with a more specific name? An integer could have several meanings (that we may want to support in the future), notably the first `h` against the rest. Or we could allow choosing a different number of first vs. last items. Not sure what would be the best name: `firstlast`?

raphaelsaavedra: I liked the keyword argument idea.

raphaelsaavedra:
I implemented your suggestions @nalimilan, please tell me if it's okay now.

src/f.jl
```diff
@@ -0,0 +1,107 @@
...
+# References
+
+  * James Durbin and Siem Jan Koopman, "Time Series Analysis by State Space Methods", 2nd Ed.,
+    2012, Oxford Statistical Science Series.
+
+# External links
+
+  * [Homoscedasticity on Wikipedia](https://en.wikipedia.org/wiki/Homoscedasticity)
+"""
+function VarianceFTest(y::AbstractVector{<: Real}; firstlast::Integer = div(length(y), 2))
```
nalimilan: This isn't correct as it overwrites the previous method (keyword arguments don't participate in dispatch). So you need to use a single method with `firstlast=nothing` by default and handle this internally.

mschauer:
I understand that this is opposite to what you wanted to do in the first place, but I would be most comfortable with just adding the `VarianceFTest(y1::AbstractVector{<: Real}, y2::AbstractVector{<: Real})` and defer any time series applications to that to user space or more specialised packages as
```
    y1 = @view y[n-h+1:n]
    y2 = @view y[1:h]
```
is easily done on the  callers site and that `VarianceFTest` with one argument means splitting the vector is a bit too magic. As a compromise, having a *method* `homoscedasticityFTest(y::AbstractVector{<: Real}; firstlast::Integer = div(length(y), 2)) = ...` returning a VarianceFTest would be okay with me.

raphaelsaavedra:
Thinking about it, I have to agree with @mschauer. The homoscedasticity test is such a simple case of `VarianceFTest` that it does not justify its own constructor. I don't even think having its own method is necessary. If @nalimilan agrees, I can just remove the homoscedasticity version of `VarianceFTest` and anyone who wants to run that test can simply do as mentioned by @mschauer.

nalimilan:
Fine with me.

test/f.jl
```diff
@@ -0,0 +1,35 @@
...
+
+        t = VarianceFTest(y1_h0, y2_h0)
+
+        @test t.n_x == 500
+        @test t.n_y == 400
+        @test t.df_x == 499
+        @test t.df_y == 399
+        @test t.F ≈ 0.974693 rtol = 1e-5
+        @test pvalue(t) ≈ 0.784563 rtol = 1e-5
+        @test default_tail(t) == :both
```
nalimilan: Can you test the two tails for the p-value here too?

src/f.jl
```diff
@@ -0,0 +1,81 @@
...
+
+# References
+
+  * George E. P. Box, "Non-Normality and Tests on Variances", Biometrika 40 (3/4): 318–335, 1953.
+
+# External links
+
+  * [F-test of equality of variances on Wikipedia](https://en.wikipedia.org/wiki/F-test_of_equality_of_variances)
+"""
+function VarianceFTest(y1::AbstractVector{<: Real}, y2::AbstractVector{<: Real}; F0::Real = 1.0)
```
nalimilan: Remove spaces around `=` for all keyword arguments here and elsewhere.

src/f.jl
```diff
@@ -0,0 +1,81 @@
...
+
+export VarianceFTest
+
+struct VarianceFTest <: HypothesisTest
+    n_x::Int   # size of sample 1
+    n_y::Int   # size of sample 2
+    df_x::Int  # degrees of freedom 1
+    df_y::Int  # degrees of freedom 2
+    F::Real    # test statistic
+    F0::Real   # variance ratio under h_0
```
mschauer: This is not used

test/f.jl
mschauer: How did you get the ground truth?

test/f.jl
```diff
@@ -0,0 +1,37 @@
+using HypothesisTests, Test
+using HypothesisTests: default_tail
+
+@testset "F-tests" begin
+    @testset "Basic variance F-test" begin
+        Random.seed!(12)
+        y1_h0 = 4 .+ randn(500)
+        y2_h0 = 4 .+ randn(400)
+
```
mschauer: Can you repeat the test also with vectors switched?

test/f.jl
```diff
@@ -0,0 +1,37 @@
...
+        t = VarianceFTest(y1_h0, y2_h0)
+
+        @test t.n_x == 500
+        @test t.n_y == 400
+        @test t.df_x == 499
+        @test t.df_y == 399
+        @test t.F ≈ 0.974693 rtol = 1e-5
```
nalimilan: Also remove spaces around `=` here and below.

test/f.jl
raphaelsaavedra: I have compared with `var.test` in R for validation.

src/f.jl
raphaelsaavedra: It's used in `show`:

```
julia> t
Variance F-test
---------------
Population details:
    parameter of interest:   variance ratio
    value under h_0:         1.0
    point estimate:          0.8073932940769317

Test summary:
    two-sided p-value:           0.7552

Details:
    number of observations: [10, 10]
    F statistic:            0.8073932940769317
    degrees of freedom:     [9, 9]
```

test/f.jl
raphaelsaavedra: Done

mschauer: But it is not used in computing a p value... it is not providing the functionality it suggests

src/f.jl
mschauer: But it is not used in computing a p value... it is not providing the functionality it suggests

raphaelsaavedra: I agree it does not provide any functionality. I can just remove F0 and leave value under h_0 as 1.0 as it will always be the case

src/f.jl
```diff
@@ -0,0 +1,80 @@
...
+    println(io, ident, "degrees of freedom:     [$(x.df_x), $(x.df_y)]")
+end
+
+function pvalue(x::VarianceFTest; tail=:both)
+    dist = FDist(x.df_x, x.df_y)
+    if tail == :both
+        Δ = abs(cdf(dist, x.F) - 0.5)
+        F1 = quantile(dist, 0.5 - Δ)
+        F2 = quantile(dist, 0.5 + Δ)
+        return cdf(dist, F1) + (1 - cdf(dist, F2))
```
mschauer:
```suggestion
        return 1 - 2abs(cdf(dist, x.F) - 0.5)
```
Simplifies to this line.  The three lines above are not needed then. But what was intended in the first place?

mschauer: Ah, I see, that level of indirection perhaps came from computing the critical region first and then the p-value from that.

raphaelsaavedra: You're correct. I simplified the calculation using your version. Thanks!

raphaelsaavedra:
Thanks, is it possible to make a new release so that we can use this test?

nalimilan:
Sure, see https://github.com/JuliaStats/HypothesisTests.jl/pull/168.

raphaelsaavedra:
Thank you!

PR #169, opened 2019-11-29
Drop 0.7, add package compat bounds, remove REQUIRE

ararslan:
And bump package version to 0.9.0, since we're dropping a Julia version.

Closes #168 
Closes #164

PR #174, opened 2020-01-08
Use strict version bounds for dependencies

nalimilan:
Allow the oldest release of dependencies that works with Julia 1.0.

We will then enable CompatHelper.


nalimilan:
Actually there was already https://github.com/JuliaStats/HypothesisTests.jl/pull/169.

PR #176, opened 2020-01-09
Make alpha a keyword argument to confint

nalimilan:
This is more consistent with StatsBase, which uses a `level` keyword argument for statistical models. In the future, we can add `level` to this package, and `alpha` to StatsBase.

### DIFF
--- a/src/wilcoxon.jl
+++ b/src/wilcoxon.jl
@@ -161,7 +161,7 @@ function pvalue(x::ExactSignedRankTest; tail=:both)
     end
 end
 
-StatsBase.confint(x::ExactSignedRankTest; alpha::Real=0.05, tail=:both) = calculate_ci(x.vals, alpha, tail=tail)
+StatsBase.confint(x::ExactSignedRankTest; level::Real=0.95, tail=:both) = calculate_ci(x.vals, level, tail=tail)
 
 
 ## APPROXIMATE SIGNED RANK TEST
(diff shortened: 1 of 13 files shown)


src/wilcoxon.jl
```diff
@@ -161,7 +161,7 @@ function pvalue(x::ExactSignedRankTest; tail=:both)
     end
 end
 
-StatsBase.confint(x::ExactSignedRankTest, alpha::Real=0.05; tail=:both) = calculate_ci(x.vals, alpha; tail=tail)
+StatsBase.confint(x::ExactSignedRankTest; alpha::Real=0.05, tail=:both) = calculate_ci(x.vals, alpha, tail=tail)
```
mschauer:
```suggestion
StatsBase.confint(x::ExactSignedRankTest; alpha::Real=0.05, tail=:both) = calculate_ci(x.vals, alpha; tail=tail)
```
Do we have a preference whether to use `;` (as below) or `,` in a function call?

nalimilan: The package seems to be inconsistent in that regard, so I tried to keep the existing style.

mschauer: Yeah, I saw that, only this one appeared deliberate.

nalimilan: There's another one just below... :-/

ararslan: FWIW two major style guides (YASGuide and BlueStyle) advise using `;` to separate keyword arguments. Even though I've worked with both on different codebases, I still find I use `,` out of habit...

nalimilan:
I've added a commit to deprecate `alpha` in favor of `level`. That's consistent with StatsBase, which matters if we don't want to support both (@ararslan was opposed to it). It's also more standard, as can be seen from the docstring which referred to `1-alpha`.

nalimilan:
Let's release 0.9.0 for now, and if we don't want to change anything to the API in a few months, we'll tag 1.0. https://github.com/JuliaRegistries/General/pull/8107

Issue #185, opened 2020-03-08
Easier and Detailed Documentation with Examples


Sinansi:
I kindly ask the owners of this library to provide easier and more detailed documentation.
I find it challenging to use this library, despite how straight forward it should be. Not all people have PhD and acquainted with Julia like you do. It would be great if you can provide examples.

Regarding, the Augmented Dickey-Fuller test, I am trying to follow the documentation, but keep getting errors.

random_vector = rand(1000)

HypothesisTests.ADFTest(random_vector)
Not working

HypothesisTests.ADFTest(random_vector, 2)
Not working

HypothesisTests.ADFTest(random_vector, "none", 2)
Not working

HypothesisTests.ADFTest(y=random_vector, deterministic="none", lag=2)
Not working

Can someone solve the enigma? How does this function like to take its arguments?

Thank you!

Sinansi:
Anyone can help on this issue please?

Sinansi:
Now I understand why Julia keep falling against Python in popularity, and soon it is going to get kicked out of the TIOBE index. The documentation is not as friendly as in Python. 

Not sure why Julia software developers prevent declaring the data type and data shape of function arguments in their documentation. I found in the source code that the arguments have to be passed as symbols, not as string as widely known.

I assumed I could get help on this issue, but very friendly community indeed.

Thank you!

matbesancon:
I am not qualified to answer here, not being familiar with the package, but for help questions like this, you should try asking on a more Q&A oriented platform (developers are more likely to keep issues for development issue and help out on forums).
Check out the different ways to reach the julia community here: julialang.org/community/
The best format may be discourse:
discourse.julialang.org

ararslan:
Hi @Sinansi. Please note that this issue was open only for a couple of days. The developers are all busy folks with other commitments who are volunteering their time to develop and maintain packages like this. Although we try our best to do so, we may not be able to respond to issues right away. For immediate feedback on questions, I encourage you to post on the [Julia Discourse forum](https://discourse.julialang.org).

If you've figured out how to solve your problem, I encourage you to contribute your knowledge to the package in the form of a pull request. That will help other folks with your same question about the interface.

PR #186, opened 2020-03-10
put signature into docs

SimonDanisch:
If the signature is fully typed, its helpful to put them in the docs!
Help with #185


pabloferz:
Probably the documentation should also state the options for the `deterministic` variable explicitly as symbols? (`:none` instead of `none`).

ararslan:
Closed by #187

PR #188, opened 2020-03-14
T-tests Request different degrees of freedom 

### LINKED ISSUE #181: Request different degrees of freedom (OneSampleTTest)
In the OneSampleTTest functions, the degrees of freedom are fixed as n-1. But if we want to perform a t-test for the p regression parameters, the degrees of freedom must be equal to n-p. Is it possible to generalize the OneSampleTTest function by:

function OneSampleTTest(xbar::Real, stddev::Real, n::Int, μ0::Real=0, df::Int = n-1)
    stderr = stddev/sqrt(n)
    t = (xbar-μ0)/stderr
    OneSampleTTest(n, xbar, df, stderr, t, μ0)
end

?

birm:
closes #181

### DIFF
--- a/src/t.jl
+++ b/src/t.jl
@@ -71,7 +71,7 @@ function show_params(io::IO, x::OneSampleTTest, ident="")
 end
 
 """
-    OneSampleTTest(xbar::Real, stddev::Real, n::Int, μ0::Real = 0)
+    OneSampleTTest(xbar::Real, stddev::Real, n::Int, μ0::Real = 0, df::Int = n-1)
 
 Perform a one sample t-test of the null hypothesis that `n` values with mean `xbar` and
 sample standard deviation `stddev`  come from a distribution with mean `μ0` against the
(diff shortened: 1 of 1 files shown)


wildart:
LGTM

src/t.jl
```diff
@@ -144,13 +143,12 @@ have different means but equal variances.
 
 Implements: [`pvalue`](@ref), [`confint`](@ref)
 """
-function EqualVarianceTTest(x::AbstractVector{T}, y::AbstractVector{S}, μ0::Real=0) where {T<:Real,S<:Real}
+function EqualVarianceTTest(x::AbstractVector{T}, y::AbstractVector{S}, μ0::Real=0, df::Int=length(x)+length(y)-2) where {T<:Real,S<:Real}
```
nalimilan:
```suggestion
function EqualVarianceTTest(x::AbstractVector{<:Real}, y::AbstractVector{<:Real}, μ0::Real=0,
                            df::Int=length(x)+length(y)-2)
```

src/t.jl
```diff
@@ -71,18 +71,17 @@ function show_params(io::IO, x::OneSampleTTest, ident="")
 end
 
 """
-    OneSampleTTest(xbar::Real, stddev::Real, n::Int, μ0::Real = 0)
+    OneSampleTTest(xbar::Real, stddev::Real, n::Int, μ0::Real = 0, df::Int = n-1)
```
nalimilan: The docstring should mention `df`. Probably better call it `dof` by the way, since it's the name of the function in StatsBase.

src/t.jl
```diff
@@ -144,13 +143,12 @@ have different means but equal variances.
 
 Implements: [`pvalue`](@ref), [`confint`](@ref)
 """
-function EqualVarianceTTest(x::AbstractVector{T}, y::AbstractVector{S}, μ0::Real=0) where {T<:Real,S<:Real}
```
nalimilan: Should update the docstring.

src/t.jl
birm: I've changed df to dof all within t.jl, but I'm concerned that this is a breaking change.

PR #189, opened 2020-03-17
Faster Mann-Whitney

alyst:
Some small optimizations to make Mann-Whitney faster:
* use views to avoid unnecessary array creation (it has a big impact, actually)
* use `@inbounds` in `tiderank_adj!()` (should help the other users of this method as well)
* tiny optimization in the math in `mwuenumerate()`


wildart:
LGTM

PR #190, opened 2020-03-18
Dunn test

wildart:
Added Dunn test: a non-parametric pairwise multiple comparisons procedure based on rank sums.


hoyttyoh:
This seems like a very useful nonparametric test to include to support Kruskal-Wallis. Any reason it went cold?

AdamWysokinski:
Will this ever be merged?

PR #196, opened 2020-04-23
fix direction of comparison for two-tailed permutation test

### LINKED ISSUE #195: p-value for permutation tests is opposite the convention for other tests
The p-value for permutation tests is computed as the fraction of samples from the permuted distribution that are LESS extreme than the observed value, whereas for other tests the p value corresponds to the probability of observing a MORE extreme value under the null hypothesis.  This behavior is apparently intentional since it's included in the tests:

https://github.com/JuliaStats/HypothesisTests.jl/blob/cb3e7acae56c4ccb6f1f86d985a76e32fc1087ab/test/permutation.jl#L3-L14

but it's confusing and also leads to strange printing of the results  of the permutation test (this came up on slack recently) since the p value for a test where the observed statistic is much more extreme than any of the permuted values prints as "fails to reject h0".  So either the p value should be changed to make it more consistent with other tests/standard convention (my preference), or the printing of the permutation test results shoudl be special-cased to be clear that the p value has a different meaning.

kleinschmidt:
Previously counted number of permuted test statistics LESS extreme than the
observed statistic.  this makes the both-tail behavior consisten with right/left
tail behavior by computing the fraction of permuted statistics that are as
or more extreme.  the tests have been updated to reflect that this makes the
two-tailed test more conservative (because ties count towards h0 under this
definition).

fixes #195

----

#

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "HypothesisTests"
 uuid = "09f84164-cd44-5f33-b23f-e6b0d136a0d5"
-version = "0.9.2"
+version = "0.10.0"
 
 [deps]
 Combinatorics = "861a8166-3701-5b0c-9a16-15d98fcdc6aa"
--- a/src/permutation.jl
+++ b/src/permutation.jl
@@ -42,7 +42,7 @@ end
 
 function pvalue(apt::PermutationTest; tail=:both)
     if tail == :both
-        count = sum(abs(apt.observation) >= abs(x) for x in apt.samples)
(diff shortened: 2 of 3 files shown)


kleinschmidt:
BTW I think this is a bug but it will change the behavior of permutation tests so I'm not sure whether this should be a patch or a minor (breaking) release....

kleinschmidt:
Okay, I've bumped the version to 0.10.0.  I'll merge soon if there are no objections.

PR #204, opened 2020-05-23
Tests for Equality of Variances

wildart:
Implemented tests for equality of variances:
- [x] One-way ANOVA Test
- [x] Levene's Test
- [x] Brown-Forsythe Test
- [x] Fligner-Killeen Test


wildart:
Ready for review.

mschauer:
It looks like Levene's test is just a form of one-way ANOVA with F-distributed test statistic. Is this somehow already covered elsewhere?

wildart:
Indeed, it is a variant of ANOVA.

Question is why there is no simple 1-way extendable ANOVA in this package that would allow to perform above tests?

I've seen this discourse thread [ANOVA Tests in Julia?](https://discourse.julialang.org/t/anova-tests-in-julia/1421/56). I could try to spin this PR in a proper ANOVA interface.

mschauer:
This looks like the right approach to me. Should the  Levene, Brown-Forsythe, Fligner-Killeen return an ANOVA object or have their own objects inheriting from that?

wildart:
I do not think it is necessary to create separate types for each test. I designed `VarianceEqualityTest` with enough flexibility to cover the above test, and any future implementations that fit the particular profile.

docs/src/parametric.md
```diff
@@ -32,4 +36,34 @@ UnequalVarianceZTest
...
\ No newline at end of file
+```
+
+## One-way ANOVA Test
+
+```@docs
+OneWayANOVATest
+```
+
+Example:
```
nalimilan, line 47: We don't show examples for other tests AFAIK. Can't you put it in the docstring instead?

src/var_equality.jl
```diff
@@ -0,0 +1,190 @@
...
+    description::Tuple{String, String, String} # test name, parameter of interest, test statistic name
+end
+
+population_param_of_interest(t::VarianceEqualityTest) = (t.description[2], "all equal", NaN)
+testname(t::VarianceEqualityTest) = t.description[1]
+teststatisticname(t::VarianceEqualityTest{TD}) where {TD <: ContinuousDistribution} =
+    length(t.description[3]) != 0 ? t.description[3] : (TD <: FDist ? "F" : "χ²")
+
+StatsBase.nobs(t::VarianceEqualityTest) = t.Nᵢ
+StatsBase.dof(t::VarianceEqualityTest{Chisq}) where T = t.DFt
```
nalimilan, line 20: `T` doesn't exist. Is this method actually covered by tests?

src/var_equality.jl
```diff
@@ -0,0 +1,190 @@
...
+StatsBase.dof(t::VarianceEqualityTest{FDist}) where T = (t.DFt, t.DFe)
+
+teststatistic(t::VarianceEqualityTest{FDist}) = (t.DFe/t.DFt)*sum(t.SStᵢ)/sum(t.SSeᵢ)
+teststatistic(t::VarianceEqualityTest{Chisq}) = let y=sum(t.SStᵢ)/sum(t.SSeᵢ); y*(t.DFe+t.DFt)/(1 + y) end # sum(t.SStᵢ)/t.s²
```
nalimilan, line 24: Use the long function form?

nalimilan:
Do these tests have any relation with the equality of variance tests [we already support](https://juliastats.org/HypothesisTests.jl/latest/parametric/#F-test-1)?

katdeane:
@mschauer, are there any updates on this? This is an ideal solution for a use-case that I'm currently running and I was wondering if there's a timeline on merging it. Thanks!

wildart:
I'll try to find some time next week to finish it.

wildart:
> Do these tests have any relation with the equality of variance tests we already support?

They are related, https://en.wikipedia.org/wiki/F-test_of_equality_of_variances#Properties, but computationally more to the ANOVA.

mschauer (approved):

nalimilan:
OK (the point of my question was to know whether they should use a common type).

Can you wrap all lines at 92 chars? Otherwise, looks good to me.

nalimilan (approved):

PR #205, opened 2020-05-24
Fix Diebold-Mariano as a TTest

guilhermebodin:
As a TTest it should have a `xbar`, `stderr`, and `\mu0`. @PaulSoderlind pointed out in #203 an error and it is due to not having these in the `DieboldMarianoTest` struct

### DIFF
--- a/src/diebold_mariano.jl
+++ b/src/diebold_mariano.jl
@@ -26,19 +26,20 @@ export DieboldMarianoTest
 
 struct DieboldMarianoTest <: TTest
     n::Int         # number of observations
+    xbar::Real     # estimated mean
     df::Int        # degrees of freedom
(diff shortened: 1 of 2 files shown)


PaulSoderlind:
One more gripe (sorry...): I would suggest cutting lines 31-42 of the test file.

Motivation: (a) the random numbers are not guaranteed to be the same across Julia versions (you need [StableRNGs.jl](https://github.com/rfourquet/StableRNGs.jl) for that). Second, the motivation/comment on line 31 may not always hold - especially not in out-of-sample setting.

mschauer:
Good point, I agree to the removal.

guilhermebodin:
Removed :)

Issue #199, opened 2020-05-14
Test for heteroskedasticity (White's etc)


PaulSoderlind:
I believe we need a test for heteroskedasticity (in a regression model). Or is it already available?

The idea of White's test is to use regress residual^2 on the regressors, their squares and (optionally) also their cross products - and test whether all slopes are zero or not. This can be implemented in terms of the R2 of this auxilliary regression. The Breusch-Pagan version is to use just the linear terms. 

EDIT: suggested function
```
"""
WhitesTest(X,e,TestType="")

Compute the Breusch-Pagan or White's test of heteroskedasticity of the residuals
of a regression model. It is assumed that `X` includes a constant.

# Input
- `X::Array`:          TxK, regressors
- `e::Vector`:         T-vector, residuals
- `TestType::String`:  "Linear" (Breusch-Pagan test), "LinearAndSquares" (White's without cross-products), "" (White's with cross-products)

# Output
- `LM::Float64`:    test statistic, distributed as Chisq(df)
- `LMp::Float64`:   p value of test
- `df::Int`:        degrees of freedom in test

"""
function WhitesTest(X,e,TestType="")

  (T,K) = (size(X,1),size(X,2))    #nobs,nvars

  if TestType == "Linear"
    z  = X
  elseif TestType == "LinearAndSquares"
    z = [X X.^2]
  else               #linear, squares and cross-products
    z = fill(NaN,T,round(Int,K*(K+1)/2))   #loop is easier and quicker than index tricks
    vv = 1
    for i = 1:K, j = i:K
      z[:,vv] = X[:,i].*X[:,j]             #eg. x1*x1, x1*x2, x2*x2
      vv      = vv + 1
    end
  end

  df  = rank(z) - 1                         #number of independent regressors in z
  e2  = e.^2
  b   = z\e2
  res = e2 - z*b
  R2  = 1 - var(res)/var(e2)
  LM  = T*R2                                #T*R2 in original papers, could also use T*R2/(1-R2)
  LMp = 1 - cdf(Chisq(df),LM)

  return LM, LMp, df

end
```

I might need help with applying the common HypothesisTests.jl interface.

Should I go ahead?

PaulSoderlind:
I've now tested and also benchmarked against R. The interface should be doable, but I have one _question_: the test comes in three flavours (see `TestType` above). What is the preferred way of interfacing that? Does the package have a standard for such cases?

wildart:
[...]

PR #206, opened 2020-05-26
add White's test (heteroskedasticity)

PaulSoderlind:
This PR adds White's (and Breusch-Pagan's) test for heteroskedasticity in a regression. It is one of the most commonly applied diagnostic tests (along with various tests for autocorrelation which already are in the package). The test gives the same result as `bptest` in R.

Closes issue https://github.com/JuliaStats/HypothesisTests.jl/issues/199


azev77:
Thanks this looks great! It might be nice to explicitly define `BreuschPaganTest` in case someone looking for it doesn't realize it's a special case of White Test. 
Something that reuses code like: 
```julia
BreuschPaganTest(X, e) = WhiteTest(X, e, :Linear)
```

PaulSoderlind:
@azev77 Thanks for the feedback. One way of doing that is to 

(1) add the function as you suggest, export it
(2) output a string from `WhiteTest()` that describes the test (eg. `TestDescription = "Breusch-Pagan's test for heteroskedasticity (linear)"` depending on the `TestType`
(3) which is used for `testname(t::WhiteTest) = t.TestDescription`

Sounds OK?

azev77:
Sounds good

PaulSoderlind:
added Breusch-Pagan test and more precise `testname(t::WhiteTest)`

azev77:
Is ‘ population_param_of_interest’ standard? 
I like ‘estimand’. 
I’m not recommending a modification, just discussing for the future...

PaulSoderlind:
@azev77 Thanks. I just tried to follow what appears to be the default in this package.

Another thing: it would be nice to make this package and `CovarianceMatrices.jl` communicate better. The easiest way is to make something like https://github.com/gragusa/CovarianceMatrices.jl/issues/49 happen. Please drop me a line in case you are interested in contributing.

src/white.jl
```diff
@@ -0,0 +1,107 @@
...
+  * [White's test on Wikipedia](https://en.wikipedia.org/wiki/White_test)
+"""
+function WhiteTest(X::AbstractVecOrMat{<:Real}, e::AbstractVector{<:Real}, TestType=:White)
+    (n,K) = (size(X,1),size(X,2))    #nobs,nvars
+
+    n == length(e) || throw(DimensionMismatch("inputs must have the same length"))
+
+    if TestType == :Linear
+        z  = X
+        TestDescription = string("Breusch-Pagan's test for heteroskedasticity (linear)")
```
wildart, line 63: There is no need to call `string` function here, string is any double-quoted text. Same below.

PaulSoderlind: Indeed. These are left-overs from another attempt. The fix will be in the next push.

azev77:
@PaulSoderlind 
check out a [recent paper](https://www.sciencedirect.com/science/article/abs/pii/S030440761630197X) (Romano & Wolf, JE, 2017)
"Therefore, it is tempting to decide based on the data which route to take: OLS or WLS. Specifically,we suggest applying a test for conditional heteroskedasticity.  Several such tests exists, the mostpopular ones being the tests of **Breusch and Pagan** (1979) and **White** (1980); also see **Koenker**(1981) and **Koenker and Bassett** (1982).  If the null hypothesis of conditional homoskedasticity itnot rejected by such a test, use the OLS estimator; otherwise, use the WLS estimator.  We call theresulting estimator theadaptive least squares(ALS) estimator.  The motivation is as follows.  Underconditional homoskedasticity, the ALS estimator will be equal to the WLS estimator with a smallprobability only (roughly equal to the nominal size of the test).  Therefore, in this case, ALS isexpected to be more efficient than WLS in finite samples, though still less efficient than OLS."

azev77:
@PaulSoderlind 
in your code you regress `e2` on `z=[X, X.^2]`
I've [seen some](https://jsf.pm-research.com/content/22/1/35.abstract) regress `e2` on `z=[y_hat, y_hat^2`

PaulSoderlind:
@azev77 Yes, there are lots of versions of these tests. For instance, what I called the BP test is really the BP/Koenker (1981) test.

The code above can be used to perform some of the other versions applied in other packages: just specify `X` as desired to have 
1.  just some of the original regressors 
2. the predicted values (`y_hat`) instead of the original regressors. 

Maybe we should add this information to the documentation of the function?

As you point out, these are restricted versions, sometimes applied save dfs/increase power. I believe Stata has some interesting routines for this, and even through on a Bonferroni approach for multiple testing. I have Julia code for Bonferroni/Holm so that could be done here - at some point in time.

PaulSoderlind:
@wildart I added a comment along the lines of the comment by @azev77.

Good to go?

wildart:
You forgot to add the test file to `test/runtests.jl`

PaulSoderlind:
@wildart I added white.jl to `test/runtests.jl`. I also added `BreuschPaganTest` to doc and test.

The test passes. I did notice, however, that `Pkg.test(HypothesisTests)` fails because of `show.jl`. This appears to be unrelated to my PR.

wildart:
Disregard nightly builds fails, especially if it doesn't related to your PR.

PaulSoderlind:
It seems the feedback so far is positive. Anything else I can do to make this move forward?

wildart:
@ararslan @mschauer Can we merge this PR?

src/white.jl
```diff
@@ -0,0 +1,110 @@
...
+    LM::Float64              #test statistic, distributed as Chisq(dof)
+    TestDescription::String  #short description of the test
+end
+
+"""
+    WhiteTest(X, e, TestType)
+
+Compute White's (or Breusch-Pagan's) test for heteroskedasticity.
+
+`X` is a matrix of regressors and `e` is the vector of residuals from the original model.
```
nalimilan, line 38: `X` can also be a vector.

src/white.jl
```diff
@@ -0,0 +1,110 @@
...
+export WhiteTest, BreuschPaganTest
+
+struct WhiteTest <: HypothesisTest
+    dof::Int                 #degrees of freedom in test
+    LM::Float64              #test statistic, distributed as Chisq(dof)
+    TestDescription::String  #short description of the test
+end
+
+"""
+    WhiteTest(X, e, TestType)
```
nalimilan, line 34: `TestType` should use lowercase. Also, maybe better make it a keyword argument
```suggestion
    WhiteTest(X, e; type=:White)
```

src/white.jl
```diff
@@ -0,0 +1,110 @@
...
+
+Compute White's (or Breusch-Pagan's) test for heteroskedasticity.
+
+`X` is a matrix of regressors and `e` is the vector of residuals from the original model.
+`TestType` is either `:Linear` for the Breusch-Pagan/Koenker test, `:LinearAndSquares` for
```
nalimilan, line 39: I'd use lowercase `:linear` and `linearandsquares` (or `linear_and_squares`), for consistency with e.g. `pvalue`'s `tail` argument.

src/white.jl
```diff
@@ -0,0 +1,110 @@
...
+White's test with linear and squared terms only (no cross-products), or `:White` (the default)
+for the full White's test (linear, squared and cross-product terms). `X` should include a
+constant. In some applications, `X` is a subset of the regressors in the original model,
+or just the fitted values. This saves degrees of freedom and may give a more powerful test.
+The test statistic is T*R2 where R2 is from the regression of `e^2` on the terms
+mentioned above. Under the null hypothesis it is distributed as `Chisquare(dof)`
```
nalimilan, line 45: Better use the actual name of the type in Distributions.jl:
```suggestion
mentioned above. Under the null hypothesis it is distributed as `Chisq(dof)`
```

src/white.jl
```diff
@@ -0,0 +1,110 @@
...
+# External links
+  * [White's test on Wikipedia](https://en.wikipedia.org/wiki/White_test)
+"""
+function WhiteTest(X::AbstractVecOrMat{<:Real}, e::AbstractVector{<:Real}, TestType=:White)
+    (n,K) = (size(X,1),size(X,2))    #nobs,nvars
+
+    n == length(e) || throw(DimensionMismatch("inputs must have the same length"))
+
+    if TestType == :Linear
+        z  = X
```
nalimilan, line 65:
```suggestion
        z = X
```

src/white.jl
```diff
@@ -0,0 +1,110 @@
...
+        TestDescription = "Breusch-Pagan's test for heteroskedasticity (linear)"
+    elseif TestType == :LinearAndSquares
+        z = [X X.^2]
+        TestDescription = "White's test for heteroskedasticity (linear and squares)"
+    else               #linear, squares and cross-products
+        z = fill(NaN,n,round(Int,K*(K+1)/2))   #loop is easier and quicker than index tricks
+        vv = 1
+        for i = 1:K, j = i:K
+            z[:,vv] = X[:,i].*X[:,j]             #eg. x1*x1, x1*x2, x2*x2
+            vv      = vv + 1
```
nalimilan, line 75: Better avoid unnecessary copies:
```suggestion
        @views for i = 1:K, j = i:K
            z[:, vv] .= X[:, i] .* X[:, j]             #eg. x1*x1, x1*x2, x2*x2
            vv += 1
```

test/white.jl
```diff
@@ -0,0 +1,224 @@
...
+
+    w_test = WhiteTest(X, e)
+    @test pvalue(w_test) ≈ 0.3458 atol=atol
+
+    show(IOBuffer(), w_test)
+
+    @test_throws DimensionMismatch WhiteTest(rand(3), rand(4))
+
+   bp_test = BreuschPaganTest(X, e,)
+   @test pvalue(bp_test) ≈ 0.1287 atol=atol
```
nalimilan, line 222:
```suggestion
    bp_test = BreuschPaganTest(X, e,)
    @test pvalue(bp_test) ≈ 0.1287 atol=atol
```

src/white.jl
```diff
@@ -0,0 +1,110 @@
...
+BreuschPaganTest(X, e) = WhiteTest(X, e, :Linear)
+
+
+testname(t::WhiteTest)                     = t.TestDescription
+population_param_of_interest(t::WhiteTest) = ("T*R2", 0, t.LM)
+default_tail(test::WhiteTest)              = :right
+
+function show_params(io::IO, t::WhiteTest, ident)
+    println(io, ident, "T*R^2 statistic:        ", t.LM)
+    println(io, ident, "degrees of freedom:     ", t.dof)
```
nalimilan, line 106: Better print the test type here by storing it in a field, rather than in the description. See how this is done in `KSampleADTest`.

src/white.jl
PaulSoderlind: OK, will change.

src/white.jl
PaulSoderlind: lowercase is fine

keyword seems trickier: it seems to interfere with the dispatch, probably because the struct which is also called `WhiteTest` (as per the convention in this package). I'll send more information on this in a later email.

src/white.jl
PaulSoderlind: OK

src/white.jl
PaulSoderlind: OK

src/white.jl
PaulSoderlind: OK

src/white.jl
PaulSoderlind: OK, except that I dislike the ` X[:, j]` thing. I would prefer no space

src/white.jl
PaulSoderlind: Sure, can be done. Either that way of by changing the `testname()` to print the proper heading. The latter is perhaps better?

PaulSoderlind:
@nalimilan 

Here is the problem with keyword arguments that I referred to above (my setting is Win 10, Julia 1.4.2/1.5.0-beta1.0)

```
struct foo
    A::Int
    B::Symbol
end

#function foo(x,B=:bar)          #works
function foo(x;B=:bar)           #gives InexactError: Int64(1.2)
  x = round(Int,x)
  return foo(x,B)
end

z = foo(1.2,:xx)
```

nalimilan:
Sorry, I don't understand the problem. You need to write `z = foo(1.2, B=:xx)` when using keyword arguments. `z = foo(1.2, :xx)` calls the default constructor, which you can override if needed using an inner constructor (see [the manual](https://docs.julialang.org/en/latest/manual/constructors/#man-inner-constructor-methods-1)). But I don't think that's the case here?

src/white.jl
```diff
@@ -0,0 +1,110 @@
...
+# LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+# OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+# WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+export WhiteTest, BreuschPaganTest
+
+struct WhiteTest <: HypothesisTest
+    dof::Int                 #degrees of freedom in test
+    LM::Float64              #test statistic, distributed as Chisq(dof)
+    TestDescription::String  #short description of the test
```
nalimilan, line 30:
```suggestion
    dof::Int       # degrees of freedom in test
    lm::Float64    # test statistic, distributed as Chisq(dof)
    descr::String  # short description of the test
```

nalimilan: Oh, and what does `LM` stand for?

src/white.jl
nalimilan: Not sure what "proper" means, but my suggestion was to have `testname` always return the same texte ("White's test for heteroskedasticity"), but print additional information saying which variant is used.

src/white.jl
nalimilan: The only things that matters is consistency -- but I don't know which convention is used in this package, if any.

src/white.jl
azev77: I thought LM = Lagrange Multiplier test statistic? 
(the only other context is LM = Linear Model)

PaulSoderlind: yes, Lagrange Multiplier test statistic. Perhaps we could use `LMstat` or `lmstat` instead

PaulSoderlind: I looked around in the lit: and it's LM everywhere. I'll stick to that.

PaulSoderlind:
@nalimilan I have made the changes you requested and make an update of the PR. And, yes, you are right about the keyword argument.

PaulSoderlind:
@nalimilan and @azev77 

I have some **2nd thoughts** about one of latest design changes: `X` should really by an `AbstractMatrix` (not an `AbstractArray`) and we should check whether it actually includes a (non-zero) constant column. Motivation: the test really needs `X` to have a constant and at least one more column. If not, the results are not valid (so a silent failure). 

To test the `X` matrix: (comments welcome):
```
constant_cols = mapslices(col->all(diff(col).==0),X,dims=1)   #loops over columns, test if constant
if !any(X[1,vec(constant_cols)] .!= 0)          #test if any constant column is non-zero
  throw(ArgumentError("one of the colums of X must be a non-zero constant"))
end
```

This clearly allocates memory, but just one column at a time.

PaulSoderlind:
I have now implemented (almost all of) the suggestions by @nalimilan limilan. In addition, I have added checks on `X` having at least two columns and including a non-zero constant column (or else the test can silently fail). Tests for these checks have been added.

@wildart good to go?

src/white.jl
```diff
@@ -0,0 +1,115 @@
...
+export WhiteTest, BreuschPaganTest
+
+struct WhiteTest <: HypothesisTest
+    dof::Int                 #degrees of freedom in test
+    LM::Float64              #Lagrange Multiplier test statistic, distributed as Chisq(dof)
+    testtype::Symbol
+end
+
+"""
+    WhiteTest(X, e; testtype = :White)
```
nalimilan, line 34: No need to repeat "test"?
```suggestion
    WhiteTest(X, e; type = :White)
```

src/white.jl
```diff
@@ -0,0 +1,115 @@
...
+
+Compute White's (or Breusch-Pagan's) test for heteroskedasticity.
+
+`X` is a matrix of regressors and `e` is the vector of residuals from the original model.
```
nalimilan, line 38: Could you say that observations are in rows and/or that variables are in columns? There are different conventions and it can be confusing.

src/white.jl
```diff
@@ -0,0 +1,115 @@
...
+# References
+  * H. White, (1980): A heteroskedasticity-consistent covariance matrix estimator and a direct test for heteroskedasticity, Econometrica, 48, 817-838.
+  * T.S. Breusch & A.R. Pagan (1979), A simple test for heteroscedasticity and random coefficient variation, Econometrica, 47, 1287-1294
+  * R. Koenker (1981), A note on studentizing a test for heteroscedasticity, Journal of Econometrics, 17, 107-112
+
+# External links
+  * [White's test on Wikipedia](https://en.wikipedia.org/wiki/White_test)
+"""
+function WhiteTest(X::AbstractMatrix{<:Real}, e::AbstractVector{<:Real}; testtype = :White)
+
```
nalimilan, line 61:
```suggestion
```

src/white.jl
```diff
@@ -0,0 +1,115 @@
...
+    (n,K) = (size(X,1),size(X,2))    #nobs,nvars
+    n == length(e) || throw(DimensionMismatch("inputs must have the same length"))
+
+    K >= 2 || throw(ArgumentError("X must have >= 2 columns"))
+
+    intercept_cols = mapslices(col->all(diff(col).==0) && (first(col) != 0),X,dims=1)
+    any(intercept_cols) || throw(ArgumentError("one of the colums of X must be a non-zero constant"))
```
nalimilan, line 68: This will avoid lots of allocations (I hope I got it right but you see the idea):
```suggestion
    intercept_col = any(eachcol(X)) do col
        first(col) != 0 && all(==(first(col)), col)
    end
    intercept_col || throw(ArgumentError("one of the colums of X must be a non-zero constant"))
```

src/white.jl
```diff
@@ -0,0 +1,115 @@
...
+    return WhiteTest(dof, LM, testtype)
+end
+
+"""
+    BreuschPaganTest(X, e)
+
+Compute Breusch-Pagan's test for heteroskedasticity.
+
+`X` is a matrix of regressors from the original model and `e` the vector of residuals.
+See `WhiteTest` for further details.
```
nalimilan, line 99:
```suggestion

This is equivalent to `WhiteTest(X, e, testtype=:linear)`.
See [`WhiteTest`](@ref) for further details.
```

test/white.jl
```diff
@@ -0,0 +1,228 @@
...
+
+    show(IOBuffer(), w_test)
+
+    @test_throws DimensionMismatch WhiteTest([rand(3) ones(3)], rand(4))
+
+    @test_throws ArgumentError WhiteTest(ones(4,1), rand(4))
+
+    @test_throws ArgumentError WhiteTest(rand(4,2), rand(4))
+
+    bp_test = BreuschPaganTest(X, e,)
```
nalimilan, line 225:
```suggestion
    bp_test = BreuschPaganTest(X, e)
```

test/white.jl
```diff
@@ -0,0 +1,228 @@
...
+    @test pvalue(bp_test) ≈ 0.1287 atol=atol
+
```
nalimilan, line 227: Can you add a test for `show`?

src/white.jl
nalimilan: Though better use lowercase `lm`, in Julia uppercase is used for types.

PaulSoderlind:
@nalimilan Thanks for the suggestions. I believe I implemented all of them. Yes, your `intercept_col` is a better idea than my old code. I changed slightly to
`intercept_col = any(col -> first(col) != 0 && all(==(first(col)), col),eachcol(X))`

However, I now notice travis test failures on Julia 1.0 (passes the 1.3 test). Because of `eachcol`?

nalimilan:
> ``
> 
> However, I now notice travis test failures on Julia 1.0 (passes the 1.3 test). Because of `eachcol`?

Ah, yes, you'll need to add a dependency on Compat 2.2.0.

test/white.jl
```diff
@@ -205,24 +205,25 @@ e = [ 5.19346;
...
     @test pvalue(w_test) ≈ 0.1287 atol=atol
 
-    w_test = WhiteTest(X, e, testtype = :linear_and_squares)
+    w_test = WhiteTest(X, e, type = :linear_and_squares)
     @test pvalue(w_test) ≈ 0.2774 atol=atol
 
     w_test = WhiteTest(X, e)
     @test pvalue(w_test) ≈ 0.3458 atol=atol
-
     show(IOBuffer(), w_test)
```
nalimilan, line 216: Sorry I should have been more specific: this should also check the contents of the `IOBuffer` using e.g. `String(take!(buf)) == """..."""`.

PaulSoderlind: @nalimilan Regarding the `eachcol` and Compat. I don't know whether the maintainers like adding another dep. Perhaps this old-fashioned approach could do the trick until support for < Julia 1.1 is dropped?
```
intercept_col = false
for i = 1:K
    col = view(X,:,i)
    intercept_col = first(col) != 0 && all(==(first(col)), col)
    intercept_col && break
end
```

PaulSoderlind: @nalimilan Regarding testing the IOBuffer, I could do something like
```
buffer = IOBuffer()
show(buffer, w_test)
str = String(take!(buffer))
@test str == refstr
```
where `refstr` is a predfined string with the expected output. 

But, I see a **problem**: the expected output contains `point estimate:          5.612418635833716`.  That's a lot of digits, and strictly requiring that they all match is somewhat in conflict with the other tests that use `isapprox`. Is there a way around this (other than always rounding the printed results?)

nalimilan: You could use `$value` to interpolate the exact value in the string.

Regarding Compat, that's always a tradeoff, but I'd say it's a small and common enough dependency that it's OK. Adding version checks everywhere is precisely what Compat allows to avoid.

PaulSoderlind: The latest push includes 
1. a test for show. (I did't find anything similar in this package, so I am not entirely sure what the expectation is.) 
2. a loop for testing if `X` has an intercept term. Not so pretty, but compatible with Julia 1.0.5 and actually pretty fast.

test/white.jl
```diff
@@ -212,8 +212,31 @@ e = [ 5.19346;
     @test pvalue(w_test) ≈ 0.2774 atol=atol
 
     w_test = WhiteTest(X, e)
-    @test pvalue(w_test) ≈ 0.3458 atol=atol
-    show(IOBuffer(), w_test)
+    w_pval = pvalue(w_test)
+    @test w_pval ≈ 0.3458 atol=atol
+
+    refstr = """
+White's (or Breusch-Pagan's) test for heteroskedasticity
```
nalimilan, line 219: You should be able to add a four-space indentation here without breaking the test.

src/white.jl
```diff
@@ -64,7 +64,13 @@ function WhiteTest(X::AbstractMatrix{<:Real}, e::AbstractVector{<:Real}; type =
 
     K >= 2 || throw(ArgumentError("X must have >= 2 columns"))
 
-    intercept_col = any(col -> first(col) != 0 && all(==(first(col)), col),eachcol(X))
+   # intercept_col = any(col -> first(col) != 0 && all(==(first(col)), col),eachcol(X))
```
nalimilan, line 67:
```suggestion
```

src/white.jl
```diff
@@ -102,11 +108,11 @@ BreuschPaganTest(X, e) = WhiteTest(X, e, type = :linear)
 
 testname(t::WhiteTest) = "White's (or Breusch-Pagan's) test for heteroskedasticity"
 
-population_param_of_interest(t::WhiteTest) = ("T*R2", 0, t.lm)
+population_param_of_interest(t::WhiteTest) = ("T*R2", 0, round(t.lm,digits=4))
```
nalimilan, line 111: Currently we don't round value in other tests, so better don't round here either. It would probably make sense to apply some rounding when printing, but that should be applied to all tests at the same time, in another PR.

PaulSoderlind: Thanks. I have skipped the round, except for testing `pvalue` which seems has a  built-in `round(z,digits=4)`. I also fixed the indentation of that """...""" string.

PR #208, opened 2020-06-08
PermutationTest overhaul

### LINKED ISSUE #207: Permutation test structs and constructors
`ExactPermutationTest` and `ApproximatePermutationTest` functions currently return instances of `PermutationTest`. Should they instead return instances of `ExactPermutationTest` and `ApproximatePermutationTest`, where

```julia
abstract type AbstractPermutationTest{T<:Real} <: HypothesisTest end

struct ExactPermutationTest{T<:Real} <: AbstractPermutationTest{T}
    # ...
end

struct ApproximatePermutationTest{T<:Real} <: AbstractPermutationTest{T}
    # ...
end

### PR DESCRIPTION
**Pull request by harryscholes**
This PR is an overhaul of the `PermutationTest` and

- Deprecates `ExactPermutationTest` and `ApproximatePermutationTest`
- Implements two constructors of `PermutationTest` for exact and approximate tests
- Changes the order of the constructor arguments, putting the function argument first, which allows the `do` block syntax to be used

Closes #207

### DIFF
--- a/docs/src/nonparametric.md
+++ b/docs/src/nonparametric.md
@@ -69,6 +69,6 @@ ApproximateSignedRankTest
 ## Permutation test
 
 ```@docs
-ExactPermutationTest
-ApproximatePermutationTest
+PermutationTest(::Function, ::AbstractVector, ::AbstractVector)
+PermutationTest(::Function, ::AbstractVector, ::AbstractVector, ::Integer)
 ```
--- a/docs/src/nonparametric.md
+++ b/docs/src/nonparametric.md
@@ -69,6 +69,5 @@ ApproximateSignedRankTest
 ## Permutation test
 
 ```@docs
-PermutationTest(::Function, ::AbstractVector, ::AbstractVector)
-PermutationTest(::Function, ::AbstractVector, ::AbstractVector, ::Integer)
+PermutationTest
 ```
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: docs/src/nonparametric.md [hunk 1]
```diff
@@ -69,6 +69,6 @@ ApproximateSignedRankTest
 ## Permutation test
 
 ```@docs
-ExactPermutationTest
-ApproximatePermutationTest
+PermutationTest(::Function, ::AbstractVector, ::AbstractVector)
+PermutationTest(::Function, ::AbstractVector, ::AbstractVector, ::Integer)
```
harryscholes, line 73: Is this okay, or should `PermutationTest` have a single docstring that documents the exact and approximate variants?

wildart:
How about to combine both tests into one?

For example, the last parameter can be set with some default value
```
PermutationTest(f::Function, x::AbstractVector{<:Real}, y::AbstractVector{<:Real}, n::Int=0)
```
so if `n` is 0 then the exact test is performed, otherwise approximate.

harryscholes:
Having one method, with `n = 0` as a default argument doesn't feel like a very Julian solution to me. I think having multiple methods for the approx and exact tests is preferable. But we could have a single docstring covering both tests

```julia
"""
    PermutationTest(f::Function, x::AbstractVector{<:Real}, y::AbstractVector{<:Real} [, n::Integer])
"""
```

wildart:
> But we could have a single docstring covering both tests

That would work as well

harryscholes:
Cool. I've unified the docstrings now

harryscholes:
Hi @wildart just wondering if you'd seen the single `PermutationTest` docstring I made?

wildart:
Yes, it's fine.

PR #214, opened 2020-10-07
Correlation test

nklb:
This PR adds the correlation test for two vector inputs x and y. It
- adds an according constructor for the CorrelationTest struct
- extends the docstring and modifies the title in the documentation
- adds a test using the already present file nutrient.txt (I've made sure to get the same output as cor.test in R)
- fixes the naming of the test (we test for partial correlation instead of correlation if k !=0, it was the other way around)

### DIFF
--- a/src/correlation.jl
+++ b/src/correlation.jl
@@ -30,6 +30,14 @@ struct CorrelationTest{T<:Real} <: HypothesisTest
     k::Int
     t::T
 
+    # Error checking is done in `cor`
+    function CorrelationTest(x::AbstractVector, y::AbstractVector)
+        r = cor(x, y)
+        n = length(x)
+        t = r * sqrt((n - 2) / (1 - r^2))
+        return new{typeof(r)}(r, n, 0, t)
+    end
+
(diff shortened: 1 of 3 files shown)


src/correlation.jl
```diff
@@ -47,10 +60,10 @@ struct CorrelationTest{T<:Real} <: HypothesisTest
 end
 
 testname(p::CorrelationTest) =
-    string("Test for nonzero ", p.k == 0 ? "partial " : "", " correlation")
+    string("Test for nonzero ", p.k != 0 ? "partial " : "", "correlation")
```
mschauer, line 63: Oh!

mschauer (approved):

src/correlation.jl
nalimilan:
```suggestion
    string("Test for nonzero ", p.k == 0 ? "partial " : "", "correlation")
```

nalimilan:
Thanks!

nalimilan:
I've tagged 0.10.2: https://github.com/JuliaRegistries/General/pull/24290

PR #216, opened 2020-10-07
Always print compactly in show method for PowerDivergenceTest

andreasnoack:
Run CI on release version of Julia


mschauer (approved):

andreasnoack:
So I had to add two more commits to make tests pass. The first commit because of the RNG breakage in Julia since the tests in f.jl were hardcoded so I've added a test dependency on StableRNGs. The second commit because the compact printing of the output made the show method test in white.jl fail, so I've adjusted it to use the compact formatting.

test/show.jl
```diff
@@ -51,6 +52,54 @@ m = PowerDivergenceTest(d)
         residuals:          [0.0, -1.11803, 1.11803]
         std. residuals:     [0.0, -1.36931, 1.36931]
     """
+else
+@test sprint(show, m, context=:compact => true) ==
```
nalimilan, line 56: I'd have used a 4-space indentation.

nalimilan (approved):

PR #229, opened 2021-04-06
Enable GitHub Actions for CI

nalimilan:


nalimilan:
Not sure why CI doesn't run on this PR. :-/

nalimilan:
This failure is weird. It's the same on all versions, yet tests passed on master back in November...

ararslan (approved):
Change looks good but yeah the failures are bizarre

nalimilan:
OK, actually these are due to a change in Distributions 0.24.14: https://github.com/JuliaStats/Distributions.jl/pull/1269. AFAICT it's just that it generates different random values, so there's no bug.

I've pushed a fix. Technically tests will only pass with Distributions >= 0.24.14, but since we don't require that version in general I haven't updated the bounds.

nalimilan:
There's still a failure on 32-bit due to trying to take the square root of a negative variance. The error is exactly the same (including the value) after changing the random seed, which is quite surprising. AFAICT this indicates a real bug at least in the handling of corner cases, so I suggest merging this PR with the remaining failures so that we don't forget we need to improve this.

ararslan (approved):

ararslan:
Could also open an issue regarding the 32-bit failures

nalimilan:
Filed https://github.com/JuliaStats/HypothesisTests.jl/issues/230.

PR #236, opened 2021-06-03
Making CorrelationTest nonparametric

ayushpatnaikgit:
The cor function performs pearson rank correlation, which assumes that the datapoints are normally distributed. The corspearman performs the spearman rank correlation test, which doesn't assume any underlying distribution. The latter is more robust and better suited for this package.

### DIFF
--- a/src/correlation.jl
+++ b/src/correlation.jl
@@ -1,6 +1,4 @@
 # Correlation
-# TODO: Implement `CorrelationTest` for two arguments
-
 using Statistics: clampcor
 
 export CorrelationTest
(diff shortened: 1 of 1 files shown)


azev77:
Why not allow both options, w the default being ‘corspearman’?

nalimilan:
Just like "correlation" (and therefore `cor`) refers to the Pearson correlation most of the time, "correlation test" is generally taken to refer to the Pearson correlation. Could you (as @azev suggested) add an argument allowing users to pass the correlation function they want? And then tests to check that the result is correct (checking against another implementation)?

nalimilan:
Closing in favor of  #304.

PR #237, opened 2021-06-07
Add t-test for cases where only stats are known

rikhuijzer:
I was trying to reproduce some t-tests conducted in a paper, but could not with HypothesisTests to my surprise.

This PR adds a EqualVarianceTTest for situations when only the statistics (mean, variance and sample size) are known.

The new extra EqualVarianceTTest is tested in the tests and I also verified it with a paper: I got exactly the same results as a paper reported. The paper was using SPSS.


rikhuijzer:
In 4be6509, I've fixed the indentation by using tabs in `test/t.jl` in line with the rest of that file. `src/t.jl` is using spaces. To ease review, I won't refactor the indentation in `test/t.jl`.

rikhuijzer:
@andreasnoack could you take a look at these changes? :)

Issue #126, opened 2018-01-17
p val > 1 in ExactMannWhitneyUTest 


nw11:
```
A = [ 1.34937, 1.75722,0.276514, 1.04546 , 1.69085 , 0.738085, 2.36313 ]
B = [ 2.62325 , 1.16533 , 1.1327  ,  0.728714]

julia>ExactMannWhitneyUTest(A,B)
Exact Mann-Whitney U test
-------------------------
Population details:
    parameter of interest:   Location parameter (pseudomedian)
    value under h_0:         0
    point estimate:          0.20035500000000006

Test summary:
    outcome with 95% confidence: fail to reject h_0
    two-sided p-value:           1.0727272727272728

Details:
    number of observations in each group: [7, 4]
    Mann-Whitney-U statistic:             14.0
    rank sums:                            [41.0, 25.0]
    adjustment for ties:                  0.0
```

ararslan:
Wow, that's extraordinarily probable.

andreasnoack:
I'm not sure this is really a bug. It is a standard issue with two-sided p-values. I've previously made [this comment](https://github.com/JuliaStats/HypothesisTests.jl/issues/98#issuecomment-301532196) although the issue here isn't that the distribution is asymmetric but that the test statistic happens to be the middle outcome of the discrete distribution. Hence, the middle observation is counted twice. I don't think it is big deal. Asymptotically, it won't be larger than one.

ararslan:
It's pretty jarring as a user. Perhaps we could just add a `clamp(pval, 0, 1)` when showing the value?

nalimilan:
I guess we should do whatever other software do.

ararslan:
I don't recall ever having seen a value outside of [0, 1] for p-values in R or SAS.

bjarthur:
glad i'm not the only one who came across this!

rpgowers:
I also found this for a case where U1 = U2, e.g.
```
a = [12,10,7,6,3,1]
b = [11,9,8,5,4,2]
```
Where 

`MannWhitneyUTest(a,b)`

yields an error:
```
Exact Mann-Whitney U test
-------------------------
Population details:
    parameter of interest:   Location parameter (pseudomedian)
    value under h_0:         0
    point estimate:          0.0

Error showing value of type ExactMannWhitneyUTest{Float64}:
ERROR: p-values must be in [0; 1]
```
whilst
[...]

PR #243, opened 2021-09-10
fix pvalue > 1 in  MannWhitneyUTest (#126)

pdeffebach:
Fix taken from [here](https://github.com/SurajGupta/r-source/blob/a28e609e72ed7c47f6ddfbb86c85279a0750f0b7/src/library/stats/R/wilcox.test.R#L76) in the R source code. 

Tests seem to fail for FischerExactTest for unrelated reasons.


src/mann_whitney.jl
```diff
@@ -139,10 +139,11 @@ function pvalue(x::ExactMannWhitneyUTest; tail=:both)
         # cannot account for ties
         if tail == :both
             if x.U < x.nx * x.ny / 2
-                2 * pwilcox(x.U, x.nx, x.ny, true)
+                p = pwilcox(x.U, x.nx, x.ny, true)
             else
-                2 * pwilcox(x.U - 1, x.nx, x.ny, false)
+                p =  pwilcox(x.U - 1, x.nx, x.ny, false)
             end
+            min(2 * p, 1)
```
pdeffebach, line 146:
```suggestion
            min(2 * p, 1.0)
```

src/mann_whitney.jl
```diff
@@ -227,7 +228,8 @@ function pvalue(x::ApproximateMannWhitneyUTest; tail=:both)
         1
     else
         if tail == :both
-            2 * ccdf(Normal(), abs(x.mu - 0.5 * sign(x.mu))/x.sigma)
+            p = ccdf(Normal(), abs(x.mu - 0.5 * sign(x.mu))/x.sigma)
+            min(2 * p, 1)
```
pdeffebach, line 232:
```suggestion
            min(2 * p, 1.0)
```

src/mann_whitney.jl
```diff
@@ -139,10 +139,11 @@ function pvalue(x::ExactMannWhitneyUTest; tail=:both)
         # cannot account for ties
         if tail == :both
             if x.U < x.nx * x.ny / 2
-                2 * pwilcox(x.U, x.nx, x.ny, true)
+                p = pwilcox(x.U, x.nx, x.ny, true)
             else
-                2 * pwilcox(x.U - 1, x.nx, x.ny, false)
+                p =  pwilcox(x.U - 1, x.nx, x.ny, false)
             end
+            min(2 * p, 1.0)
```
devmotion, line 146: I'm curious, why a `Float64` literal here?

pdeffebach: Probably no good reason. Just did it because I know `p` will be a `Float64`

pdeffebach:
Ready for a review @andreasnoack @nalimilan

src/mann_whitney.jl
```diff
@@ -139,10 +139,11 @@ function pvalue(x::ExactMannWhitneyUTest; tail=:both)
         # cannot account for ties
         if tail == :both
             if x.U < x.nx * x.ny / 2
-                2 * pwilcox(x.U, x.nx, x.ny, true)
+                p = pwilcox(x.U, x.nx, x.ny, true)
             else
-                2 * pwilcox(x.U - 1, x.nx, x.ny, false)
+                p =  pwilcox(x.U - 1, x.nx, x.ny, false)
```
nalimilan, line 144:
```suggestion
                p = pwilcox(x.U - 1, x.nx, x.ny, false)
```

test/mann_whitney.jl
```diff
@@ -53,4 +53,22 @@ end
 @testset "Issue #39" begin
 	@test abs(pvalue(ExactMannWhitneyUTest(Float32[1:10;], Float32[2:11;])) - 0.5096) <= 1e-4
 end
+
+@testset "Issue #126" begin
+	A = [ 1.34937, 1.75722,0.276514, 1.04546 , 1.69085 , 0.738085, 2.36313 ]
+	B = [ 2.62325 , 1.16533 , 1.1327  ,  0.728714]
```
nalimilan, line 59:
```suggestion
	A = [1.34937, 1.75722,0.276514, 1.04546, 1.69085, 0.738085, 2.36313 ]
	B = [2.62325, 1.16533, 1.1327,  0.728714]
```

test/mann_whitney.jl
```diff
@@ -53,4 +53,22 @@ end
 @testset "Issue #39" begin
 	@test abs(pvalue(ExactMannWhitneyUTest(Float32[1:10;], Float32[2:11;])) - 0.5096) <= 1e-4
 end
+
+@testset "Issue #126" begin
```
nalimilan, line 57: Say what this issue is about?

nalimilan (commented):
Thanks. It's kind of weird that we have to clamp the value like that, but that's less surprising if other software do that. Anyway `PValue` throws an error for values above 1 and we probably shouldn't change that.

Do we need the same fix for other Mann-Whitney tests?

test/mann_whitney.jl
pdeffebach: Added.

pdeffebach:
I fixed all the ocurrences of `pvalue` in that file. `MannWhitneyUTest` dispatches to either the exact or approximate depending on some condition. 

I'm sure there are more cases of bad p-values but we can tackle those more slowly in a slew of smaller PRs as they are discovered.

nalimilan (approved):

src/mann_whitney.jl
```diff
@@ -139,10 +139,11 @@ function pvalue(x::ExactMannWhitneyUTest; tail=:both)
         # cannot account for ties
         if tail == :both
             if x.U < x.nx * x.ny / 2
-                2 * pwilcox(x.U, x.nx, x.ny, true)
+                p = pwilcox(x.U, x.nx, x.ny, true)
             else
-                2 * pwilcox(x.U - 1, x.nx, x.ny, false)
+                p = pwilcox(x.U - 1, x.nx, x.ny, false)
             end
```
devmotion, line 145: Maybe you could write
```julia
            p = if x.U < x.nx * x.ny / 2
                pwilcox(x.U, x.nx, x.ny, true)
            else
                pwilcox(x.U - 1, x.nx, x.ny, false)
            end
```
instead (can't make a suggestion on github it seems)? I think it is a bit more readable but I don't think it makes a difference for the compiler (maybe it does?), so it's just a matter of taste I guess.

src/mann_whitney.jl
devmotion: BTW below in the other branch `min(1, 2 * ...)` is used, i.e. no `Float64` literal.

src/mann_whitney.jl
```diff
@@ -227,7 +228,8 @@ function pvalue(x::ApproximateMannWhitneyUTest; tail=:both)
         1
     else
         if tail == :both
-            2 * ccdf(Normal(), abs(x.mu - 0.5 * sign(x.mu))/x.sigma)
+            p = ccdf(Normal(), abs(x.mu - 0.5 * sign(x.mu))/x.sigma)
+            min(2 * p, 1.0)
```
devmotion, line 232: A comment related to this function (and your comment that a `Float64` will be returned) but unrelated to this change: this function is not type stable since it returns `1` in the other branch above.

test/mann_whitney.jl
```diff
@@ -53,4 +53,22 @@ end
 @testset "Issue #39" begin
 	@test abs(pvalue(ExactMannWhitneyUTest(Float32[1:10;], Float32[2:11;])) - 0.5096) <= 1e-4
 end
+
+@testset "Issue #126 pvalues above 1" begin
+	A = [1.34937, 1.75722,0.276514, 1.04546, 1.69085, 0.738085, 2.36313 ]
```
devmotion, line 58:
```suggestion
	A = [1.34937, 1.75722, 0.276514, 1.04546, 1.69085, 0.738085, 2.36313]
```

test/mann_whitney.jl
```diff
@@ -53,4 +53,22 @@ end
...
+	B = [2.62325, 1.16533, 1.1327,  0.728714]
```
devmotion, line 59:
```suggestion
	B = [2.62325, 1.16533, 1.1327, 0.728714]
```

test/mann_whitney.jl
```diff
@@ -53,4 +53,22 @@ end
...
+
+	m = ExactMannWhitneyUTest(A,B)
+	p = pvalue(m; tail = :both)
+	@test p == 1
```
devmotion, line 63: I guess you could just test
```suggestion
	@test p <= 1
```
It does not see relevant that it is actually 1.

test/mann_whitney.jl
```diff
@@ -53,4 +53,22 @@ end
...
+
+	A = [12,10,7,6,3,1]
+	B = [11,9,8,5,4,2]
+
+	m = MannWhitneyUTest(A,B)
+	p = pvalue(m; tail = :both)
+
+	@test p == 1
+end
+
```
devmotion, line 73:
```suggestion
	@test p <= 1
end
```

pdeffebach:
Bumping this @andreasnoack. Can I get a review / merge?

andreasnoack:
@devmotion Have your comments been addressed?

test/mann_whitney.jl
devmotion: Let's leave it. If floating point issues cause test errors we can change it to `<= 1`.

test/mann_whitney.jl
```diff
@@ -53,4 +53,22 @@ end
...
+	B = [2.62325, 1.16533, 1.1327, 0.728714]
+
+	m = ExactMannWhitneyUTest(A,B)
+	p = pvalue(m; tail = :both)
+	@test p == 1
+
+	A = [12,10,7,6,3,1]
+	B = [11,9,8,5,4,2]
+
+	m = MannWhitneyUTest(A,B)
```
devmotion, line 68: This will also be an `ExactMannWhitneyUTest`, so the changes for the p-value and type stability in the `ApproximateWhitneyUTest` are not tested.

I wanted to suggest changing this to
```julia
	m = ApproximateMannWhitneyUTest(A,B)
```
but unfortunately the p-value is already 1 for this example so it can only be used for testing type stability of `pvalue`.

devmotion (changes requested):
It seems the type instability was fixed :+1: I think two things should be improved:
- Can you add a test for the type stability of `pvalue(::ApproximateMannWhitneyTest)`? It can be provoked on the master branch with basically any standard example and you can just change `pvalue` to `@inferred(pvalue(...))`.
- Can you add a test for `pvalue(::ApproximateMannWhitneyUTest; tail=:both) <= 1`? Currently both tests just check the `ExactMannWhitneyUTest`. Unfortunately, the second example already works on the master branch for the approximate version so one needs different data to provoke a p-value > 1.

test/mann_whitney.jl
pdeffebach: I can't find inputs to get the p-value above `1` here. But I did fix an inference issue, which is nice.

pdeffebach:
> Can you add a test for the type stability of pvalue(::ApproximateMannWhitneyTest)? It can be provoked on the master branch with basically any standard example and you can just change pvalue to @inferred(pvalue(...)).

Added `@inferred` where possible, including fixing a bad inference in the `ApproximateMannWhitneyU` test. 

> Can you add a test for pvalue(::ApproximateMannWhitneyUTest; tail=:both) <= 1? Currently both tests just check the ExactMannWhitneyUTest. Unfortunately, the second example already works on the master branch for the approximate version so one needs different data to provoke a p-value > 1.

Added a test, which is really just testing inference.

pdeffebach:
Thanks, let me know if you have more comments.

devmotion (approved):
Looks good to me now, thank you @pdeffebach for taking into account all comments!

devmotion:
Seems Julia 1.7 broke the tests, probably due to the RNG changes. In general, equality checks of random results with fixed seeds seem very brittle and not reliable since different Julia versions are not guaranteed to yield the same random numbers (as happened now) - IMO ideally they should be replaced with some more stable tests. But of course this is not something you should have to deal with in this PR...

pdeffebach:
@andreasnoack can this be merged? Test failures unrelated.

andreasnoack:
Let's get tests green before merging. I've opened https://github.com/JuliaStats/HypothesisTests.jl/pull/252 as a minimal PR to make tests green again.

pdeffebach:
Will we merge #252 and then rebase?

pdeffebach:
Now that #252 is merged, what is the next course of action?

devmotion:
Update the PR so that tests pass, i.e., either merge the master branch or rebase the PR :slightly_smiling_face:

pdeffebach:
Okay I rebased

Then I changed the base branch back and forth on the github UI so the changes look good. I think we should run the workflow.

nalimilan (approved):
@devmotion Merge?

devmotion (approved):

devmotion:
Thank you @pdeffebach :slightly_smiling_face:

PR #247, opened 2021-10-24
Mention other names for OneSampleTTest

### LINKED ISSUE #246: A paired t-test is not necessarily a one sample test
According to the source, paired *t*-test is an instance of a `OneSampleTTest` :

https://github.com/JuliaStats/HypothesisTests.jl/blob/dd35ae0f1c08b3ae3558e1dada9fc57e41eff1aa/src/t.jl#L100-L113

From [Wikipedia](https://en.wikipedia.org/wiki/Student's_t-test#Dependent_t-test_for_paired_samples): "This test is used when the **samples** are dependent [emphasis mine]". Yes, indeed a paired t-test can also be used for "only one **sample** that has been tested twice (repeated measures)". 

So, implementation wise I don't see a problem, but the docs suggests that a paired t-test is always on one sample and that is wrong.

rikhuijzer:
Closes #246

@mschauer, given your comments in the linked issue, I've been thinking about a compromise. Maybe we can add a note listing some alternative names to indicate to readers that the name `OneSampleTTest` is also just a name and shouldn't be taken too literally.

### DIFF
--- a/src/t.jl
+++ b/src/t.jl
@@ -107,7 +107,7 @@ alternative hypothesis that the distribution does not have mean `μ0`.
 Implements: [`pvalue`](@ref), [`confint`](@ref)
     
 !!! note
-    This test is also known as a dependent samples t-test, dependent means t-test or correlated pairs t-test.
+    This test is also known as a t-test for paired or dependent samples,  see [paired difference test](https://en.wikipedia.org/wiki/Paired_difference_test) on Wikipedia.
 """
 function OneSampleTTest(x::AbstractVector{T}, y::AbstractVector{S}, μ0::Real=0) where {T<:Real, S<:Real}
     check_same_length(x, y)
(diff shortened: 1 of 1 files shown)


nalimilan (approved):
Sounds reasonable.

src/t.jl
```diff
@@ -105,6 +105,9 @@ values in vectors `x` and `y` come from a distribution with mean `μ0` against t
 alternative hypothesis that the distribution does not have mean `μ0`.
 
 Implements: [`pvalue`](@ref), [`confint`](@ref)
+    
+!!! note
+    This test is also known as a dependent samples t-test, dependent means t-test or correlated pairs t-test.
```
mschauer, line 110:
```suggestion
    This test is also known as a dependent samples t-test.
```

Correlated pairs t-test has 732 hits against "dependent samples t-test" which has hundreds of thousands, and I don't know what a "dependent mean" could be.

nalimilan (approved):

PR #252, opened 2021-12-28
Use StableRNG in more tests to avoid breakage in Julia 1.7

andreasnoack:
Also, introduce an `rng` keyword argument in `ApproximatePermutationTest` to allow passing an `AbstractRNG` object. This avoid the reliance on the global RNG. I decided to use a keyword argument instead of using the first argument as we do for the `rand` methods. I think it's easier to use a keyword argument for functions that take many argument. Please comment if you disagree.

### DIFF
--- a/src/permutation.jl
+++ b/src/permutation.jl
@@ -18,8 +18,11 @@ end
 Perform a permutation test (a.k.a. randomization test) of the null hypothesis
 that `f(x)` is equal to `f(y)`.  All possible permutations are sampled.
 """
-function ExactPermutationTest(x::AbstractVector{R}, y::AbstractVector{S},
-                              f::Function) where {R<:Real,S<:Real}
+function ExactPermutationTest(
+    x::AbstractVector{R},
+    y::AbstractVector{S},
+    f::Function
+) where {R<:Real,S<:Real}
     xy, rx, ry = ptstats(x,y)
     P = permutations(xy)
     samples = [f(view(p,rx)) - f(view(p,ry)) for p in P]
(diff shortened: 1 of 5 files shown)


src/permutation.jl
```diff
@@ -33,10 +33,15 @@ Perform a permutation test (a.k.a. randomization test) of the null hypothesis
...
 """
-function ApproximatePermutationTest(x::AbstractVector{R}, y::AbstractVector{S},
-                                    f::Function, n::Int) where {R<:Real,S<:Real}
+function ApproximatePermutationTest(
+    x::AbstractVector{R},
+    y::AbstractVector{S},
+    f::Function,
+    n::Int;
+    rng::AbstractRNG=Random.default_rng()
+) where {R<:Real,S<:Real}
```
nalimilan, line 42: Can you keep the same style as in the rest of the file? I'm not sure how to indent the last line though.
```suggestion
function ApproximatePermutationTest(x::AbstractVector{R},
                                    y::AbstractVector{S},
                                    f::Function,
                                    n::Int;
                                    rng::AbstractRNG=Random.default_rng()) where
                                    {R<:Real,S<:Real}
```

devmotion (approved):

nalimilan (commented):
Thanks! I'm almost tempted to put `rng` as the first positional argument like in Base and Random -- but that's indeed a lot of arguments... Do we have examples of other functions that use an `rng` keyword argument?

src/permutation.jl
andreasnoack: I've made the indentation in this file consistent. The old way of doing it wastes a lot of horizontal space.

andreasnoack:
I've changed the argument from a keyword to a positional one. We can consider adding the keyword argument in a separate PR if needed. I added a test for it and it revealed that we'd have to bump the minimum Julia requirement to 1.3 since `Random.default_rng` wasn't introduced with that version. With a new long term support release out, I don't think it's worth adding version conditional definitions.

src/permutation.jl
nalimilan: Wait, none of the other files in the project (nor in JuliaStats packages AFAIK) use this style, so better stick to the previous one. If we decide to change styles we should discuss this first.

andreasnoack: The new style was discussed and adopted six years ago in https://github.com/JuliaStats/GLM.jl/pull/126 :trollface: I propose that we gradually move towards that style as we touch the files for other reasons. It really is a nicer formatting for functions with many arguments.

nalimilan: Styles are the worst thing to discuss so I'd rather stick to what we have now, unless we apply an autoformatter to all our codebase so that we no longer have to deal with these issues.

pdeffebach:
Is there any way I can help to get this PR merged?

devmotion:
What's holding back this PR? It looks good to me and it would be nice to fix the test failures :slightly_smiling_face:

nalimilan:
Only the debate about style. I vote for deferring changes in style conventions and discuss this elsewhere (style should be consistent across JuliaStats anyway).

andreasnoack:
Feel free to adjust the alignment and be done with it. I'll just point out that, in practice, we'll be stuck with the white-space-wasteful style with the proposed approach since nobody is going to go through all the JuliaStats repos and adjust them in one go.

nalimilan:
> Feel free to adjust the alignment and be done with it. I'll just point out that, in practice, we'll be stuck with the white-space-wasteful style with the proposed approach since nobody is going to go through all the JuliaStats repos and adjust them in one go.

OK, done. To be clear, I'm not asking that we update all repos at the same time, just that 1) we discuss this in advance with all JuliaStats people first, and 2) that we switch whole repos using an automated formatter such as JuliaFormatter.jl rather than changing things manually, so that we can automate this for future PRs.

PR #254, opened 2022-01-04
Better docstring for VarianceTests

KronosTheLate:
This applies to both Equal and Unequal, T and Z tests. E.g. EqualVarianceTTest.

I have added the third positional argument mu0, as well as it default value, in the docstring.
I have also added a reference to VarianceTTest, because I does not hurt, and that was what I was looking for, so it would have been very helpful to me.

### DIFF
--- a/src/t.jl
+++ b/src/t.jl
@@ -160,11 +160,13 @@ end
 
 """
     EqualVarianceTTest(x::AbstractVector{T<:Real}, y::AbstractVector{T<:Real})
+    EqualVarianceTTest(x::AbstractVector{T<:Real}, y::AbstractVector{T<:Real}, μ0::Real)
 
 Perform a two-sample t-test of the null hypothesis that `x` and `y` come from distributions
 with equal means and variances against the alternative hypothesis that the distributions
-have different means but equal variances.
+have different means but equal variances. If μ0 is not set, it defaults to 0.
 
+For testing if two datasets have equal variance, see [`VarianceFTest`](@ref).
 Implements: [`pvalue`](@ref), [`confint`](@ref)
 """
 function EqualVarianceTTest(x::AbstractVector{T}, y::AbstractVector{S}, μ0::Real=0) where {T<:Real,S<:Real}
@@ -204,6 +206,7 @@ equation:
         \\frac{(k_i s_i^2)^2}{ν_i}}
 ```
 
+For testing if two datasets have equal variance, see [`VarianceFTest`](@ref).
 Implements: [`pvalue`](@ref), [`confint`](@ref)
 """
 function UnequalVarianceTTest(x::AbstractVector{T}, y::AbstractVector{S}, μ0::Real=0) where {T<:Real,S<:Real}
--- a/src/t.jl
+++ b/src/t.jl
@@ -166,6 +166,7 @@ with equal variances and mean difference μ0 (defaults to 0) against the alterna
 that the distributions have equal variances and a mean difference different from μ0.
 
 See also:  [`VarianceFTest`](@ref) to test whether two datasets have equal variance.
+    
 Implements: [`pvalue`](@ref), [`confint`](@ref)
 """
 function EqualVarianceTTest(x::AbstractVector{T}, y::AbstractVector{S}, μ0::Real=0) where {T<:Real,S<:Real}
@@ -206,6 +207,7 @@ equation:
 ```
 
 See also:  [`VarianceFTest`](@ref) to test whether two datasets have equal variance.
+    
 Implements: [`pvalue`](@ref), [`confint`](@ref)
 """
 function UnequalVarianceTTest(x::AbstractVector{T}, y::AbstractVector{S}, μ0::Real=0) where {T<:Real,S<:Real}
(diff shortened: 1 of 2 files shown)


nalimilan:
Thanks, but the meaning of the argument should be mentioned too, otherwise it's impossible to use.

src/z.jl
```diff
@@ -131,12 +131,13 @@ testname(::EqualVarianceZTest) = "Two sample z-test (equal variance)"
...
 
 """
-    EqualVarianceZTest(x::AbstractVector{T<:Real}, y::AbstractVector{T<:Real})
+    EqualVarianceZTest(x::AbstractVector{T<:Real}, y::AbstractVector{T<:Real}, μ0::Real=0)
 
 Perform a two-sample z-test of the null hypothesis that `x` and `y` come from distributions
 with equal means and variances against the alternative hypothesis that the distributions
 have different means but equal variances.
 
+For testing if two datasets have equal variance, see [`VarianceFTest`](@ref).
```
nalimilan, line 140:
```suggestion
See also:  [`VarianceFTest`](@ref) to test whether two datasets have equal variance.

```
(Same below.)

KronosTheLate:
> Thanks, but the meaning of the argument should be mentioned too, otherwise it's impossible to use.

Good point. I have now changed all 4 docstrings to mention μ0, in a uniform and clear manner. Satisfied with my suggestion?

src/t.jl
```diff
@@ -159,12 +159,14 @@ function EqualVarianceTTest(nx::Int, ny::Int, mx::Real, my::Real, vx::Real, vy::
...
 
 """
-    EqualVarianceTTest(x::AbstractVector{T<:Real}, y::AbstractVector{T<:Real})
+    EqualVarianceTTest(x::AbstractVector{T<:Real}, y::AbstractVector{T<:Real}, μ0::Real=0)
 
 Perform a two-sample t-test of the null hypothesis that `x` and `y` come from distributions
-with equal means and variances against the alternative hypothesis that the distributions
-have different means but equal variances.
+with equal variances and mean difference μ0 (defaults to 0) against the alternative hypothesis
+that the distributions have equal variances and a mean difference different from μ0.
```
nalimilan, line 166: It seems important to mention the order of the subtraction:
```suggestion
with equal variances and a difference between their means equal to `μ0`
(i.e. `mean(X) - mean(Y) == μ0`) against the alternative hypothesis that
the distributions have equal variances and a difference between their means
different from `μ0`. By default  `μ0 = 0`, that is the null hypothesis
is that means are equal.
```

mschauer:
I would separate model assumptions, e.g. equal variance from hypotheses (eq. equal mean.) Something like:

> For samples from two populations perform a two sample `t`-test for the null-hypothesis that the means of the two populations differ by $μ_diff = E[Y] - E [X]$ under the assumption that both populations have equal, but unknown variances.

Can we also rename `μ0` to  `μ_diff` or `Δμ`?

KronosTheLate:
> I would separate model assumptions, e.g. equal variance from hypotheses (eq. equal mean.) Something like:
> 
> > For samples from two populations perform a two sample `t`-test for the null-hypothesis that the means of the two populations differ by $μ_diff = E[Y] - E [X]$ under the assumption that both populations have equal, but unknown variances.
> 
> Can we also rename `μ0` to `μ_diff` or `Δμ`?

I think that both you suggestions are great. Do you agree that `$μ_diff = E[Y] - E [X]$` would be more readable as `$μ_diff = mean(x) - mean(y)$`? I think that it is understandable even if you don't know standard notation, and that since it is actual code, it is as clear as can be. Note that the data arguments are (x, y), and a quick test tells me that the point estimate is mean(x)-mean(y), so I have swapped the order.

On the matter of `μ_diff` ovs`Δμ`, I would personally think that `Δμ` looks the best. And since we are already using Unicode for mu, adding the delta should not do much harm.

Does `μ0` occur in other places with the meaning "mean difference"? If so, this change should be enforced everywhere.

nalimilan:
Yes the E[X] notation may be more appropriate.

`μ0` is used as an argument to `OneSampleTTest`, but there it's really the mean, so using a different notation is justified. It's also used as the name of an (internal) field of some other tests, but again referring to a mean. So overall it doesn't seem you'd have to rename anything else.

src/t.jl
KronosTheLate: Would you be okay with rewording from "
By default  `μ0 = 0`, that is the null hypothesis is that means are equal."
to
"By default  the null hypothesis is that the means are equal, meaning `μ0 = 0`"

nalimilan: Yeah sure.

KronosTheLate:
So mu0 should become Deltamu (on phone, no unicode). If you see this comment, please vote yes or no by comment and reason, so that the potential rename is a real concensus.

nalimilan:
Maybe call the argument `Δμ0`, for consistency with the one-sample tests where `μ0` is used to refer to "mean under H0"?

KronosTheLate:
Sounds good by me. Are you okay with making the change?

nalimilan (approved):

Issue #263, opened 2022-03-03
Poor inference for `pvalue`


nalimilan:
The inferred return type of `pvalue` is quite imprecise or even weird for some tests. This appears to be a regression from some time ago (as it didn't happen on an old state I had), but the problem doesn't go away when using old versions so I suspect it's related to a change in a dependency.
```julia
julia> @code_warntype pvalue(SignedRankTest([1], [1]))
MethodInstance for HypothesisTests.pvalue(::ExactSignedRankTest{Int64})
  from pvalue(x::ExactSignedRankTest; tail) in HypothesisTests at /home/milan/.julia/dev/HypothesisTests/src/wilcoxon.jl:135
Arguments
  #self#::Core.Const(HypothesisTests.pvalue)
  x::ExactSignedRankTest{Int64}
Body::Union{Nothing, Float64, Int64}
1 ─ %1 = HypothesisTests.:(var"#pvalue#33")(:both, #self#, x)::Union{Nothing, Float64, Int64}
└──      return %1

julia> @code_warntype pvalue(EqualVarianceTTest([1], [1]))
MethodInstance for HypothesisTests.pvalue(::EqualVarianceTTest)
  from pvalue(x::HypothesisTests.TTest; tail) in HypothesisTests at /home/milan/.julia/dev/HypothesisTests/src/t.jl:31
Arguments
  #self#::Core.Const(HypothesisTests.pvalue)
  x::EqualVarianceTTest
Body::Any
1 ─ %1 = HypothesisTests.:(var"#pvalue#29")(:both, #self#, x)::Any
└──      return %1

# But
julia> @code_warntype pvalue(MannWhitneyUTest([1], [1]))
MethodInstance for HypothesisTests.pvalue(::ExactMannWhitneyUTest{Float64})
  from pvalue(x::ExactMannWhitneyUTest; tail) in HypothesisTests at /home/milan/.julia/dev/HypothesisTests/src/mann_whitney.jl:136
Arguments
  #self#::Core.Const(HypothesisTests.pvalue)
  x::ExactMannWhitneyUTest{Float64}
Body::Float64
1 ─ %1 = HypothesisTests.:(var"#pvalue#27")(:both, #self#, x)::Float64
└──      return %1
```

This is one of the causes of https://github.com/JuliaStats/StatsBase.jl/issues/771, though the error is due to a bug in StatsBase.

nalimilan:
[...]

PR #264, opened 2022-03-04
Fix some type instabilities of `pvalue` + `confint`

devmotion:
This PR fixes a part of https://github.com/JuliaStats/HypothesisTests.jl/issues/263.

### DIFF
--- a/src/HypothesisTests.jl
+++ b/src/HypothesisTests.jl
@@ -73,7 +73,8 @@ function pvalue(dist::ContinuousUnivariateDistribution, x::Number; tail=:both)
     check_tail(tail)
 
     if tail == :both
-        2 * min(cdf(dist, x), ccdf(dist, x))
+        p = 2 * min(cdf(dist, x), ccdf(dist, x))
+        min(p, oneunit(p)) # if P(X = x) > 0, then possibly p > 1
     elseif tail == :left
         cdf(dist, x)
     else # tail == :right
@@ -85,7 +86,8 @@ function pvalue(dist::DiscreteUnivariateDistribution, x::Number; tail=:both)
     check_tail(tail)
 
     if tail == :both
-        2 * min(ccdf(dist, x-1), cdf(dist, x))
+        p = 2 * min(ccdf(dist, x-1), cdf(dist, x))
+        min(p, oneunit(p)) # if P(X = x) > 0, then possibly p > 1
     elseif tail == :left
         cdf(dist, x)
     else # tail == :right
--- a/src/HypothesisTests.jl
+++ b/src/HypothesisTests.jl
@@ -73,7 +73,7 @@ function pvalue(dist::ContinuousUnivariateDistribution, x::Number; tail=:both)
     check_tail(tail)
 
     if tail == :both
-        min(2 * min(cdf(dist, x), ccdf(dist, x)), 1)
+        2 * min(cdf(dist, x), ccdf(dist, x))
     elseif tail == :left
         cdf(dist, x)
     else # tail == :right
@@ -85,7 +85,7 @@ function pvalue(dist::DiscreteUnivariateDistribution, x::Number; tail=:both)
     check_tail(tail)
 
     if tail == :both
-        min(2 * min(ccdf(dist, x-1), cdf(dist, x)), 1)
+        2 * min(ccdf(dist, x-1), cdf(dist, x))
     elseif tail == :left
         cdf(dist, x)
     else # tail == :right
--- a/src/HypothesisTests.jl
+++ b/src/HypothesisTests.jl
@@ -69,31 +69,39 @@ If `tail` is `:both` (default), then the p-value for the two-sided test is retur
 function pvalue end
 
 # Basic function for finding a p-value given a distribution and tail
-pvalue(dist::ContinuousUnivariateDistribution, x::Number; tail=:both) =
-    if tail == :both
-        min(2 * min(cdf(dist, x), ccdf(dist, x)), 1.0)
-    elseif tail == :left
-        cdf(dist, x)
-    elseif tail == :right
-        ccdf(dist, x)
-    else
-        throw(ArgumentError("tail=$(tail) is invalid"))
-    end
+function pvalue(dist::ContinuousUnivariateDistribution, x::Number; tail=:both)
+    check_tail(tail)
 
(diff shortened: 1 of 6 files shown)


src/wilcoxon.jl
```diff
@@ -250,8 +253,8 @@ function calculate_ci(x::AbstractVector, level::Real=0.95; tail=:both)
     n = length(x)
     m = div(n * (n + 1), 2)
     k_range = 1:div(m, 2)
-    l = [1 - 2 * psignrank(i, n, true) for i in k_range]
-    k = argmin(abs.(l .- c))
+    ks = [abs(1 - 2 * psignrank(i, n, true) - c) for i in k_range]
+    k = argmin(ks)
```
devmotion, line 257:
```suggestion
    l = [1 - 2 * psignrank(i, n, true) for i in k_range]
    k = argmin(abs.(l .- c))
```

src/HypothesisTests.jl
```diff
@@ -69,34 +69,42 @@ If `tail` is `:both` (default), then the p-value for the two-sided test is retur
 function pvalue end
 
 # Basic function for finding a p-value given a distribution and tail
-pvalue(dist::ContinuousUnivariateDistribution, x::Number; tail=:both) =
+function pvalue(dist::ContinuousUnivariateDistribution, x::Number; tail=:both)
+    check_tail(tail)
+
     if tail == :both
-        min(2 * min(cdf(dist, x), ccdf(dist, x)), 1.0)
+        min(2 * min(cdf(dist, x), ccdf(dist, x)), 1)
```
nalimilan, line 76: Why not keep this one? I would actually have done the contrary. :-D

nalimilan (commented):
Great, thanks!

src/HypothesisTests.jl
devmotion: Because in this definition we don't restrict, and hence don't know about, the types of `x` and possible parameters of the distribution. So `cdf` and `ccdf` might not return a `Float64` but eg a `Float32`. And therefore the `Float64` literal can cause a type instability.

devmotion: Hm I thought `min` promotes the arguments but it seems it doesn't? https://github.com/JuliaLang/julia/blob/bf534986350a991e4a1b29126de0342ffd76205e/base/operators.jl#L506

Then we should explicitly promote or maybe just use `oneunit`.

nalimilan: Yeah that was my concern. Maybe we should just add `::Float64` to all p-value methods? I don't see under which circumstances one would prefer getting a p-value in another type.

devmotion: I think the better approach is to promote here. For all standard arguments it will be `Float64`, so it won't cause any surprising return types.

Restricting everything to `Float64` seems unnecessary (most methods don't allow such general arguments and hence there's no such freedom as in this method here) and a bit unidiomatic - why restrict it if it works fine? More importantly though, I learnt in Distributions how problematic for e.g. GPU support and AD a `Float64`-only approach and how long it takes to (somewhat) fix it. So on purpose, both in StatsFuns and Distributions neither arguments nor return values of `cdf` and `ccdf` (which are called here) are restricted to `Float64`. Thus even though I don't have a more concrete example at hand apart from abstract GPU and AD applications, I think it's better to just fix the line here.

devmotion: On a second thought, is `min(..., 1)` actually necessary here? We should always have `min(cdf, ccdf) <= 1//2` but maybe there were some numerical issues that were fixed by the additional truncation to 1?

nalimilan: OK, promoting should be fine (it's just a bit longer to write). It seems safer to keep `min`, as we have seen several numerical issues with p-values. But `clamp` is probably more convenient as it converts the result.

devmotion: I've seen these issues but I still wonder if  here in this method values above 1 should just be considered an upstream bug. And if hence it is safe to assume that values are valid given that `cdf` and `ccdf` are implemented correctly.

devmotion: Basically similar to how we don't clamp values in the other branches because we assume upstream packages ensure that they are between 0 and 1.

nalimilan: Makes sense.

devmotion: But, of course, another alternative would be to clamp them to the unit interval in all branches - I just think it shouldn't be necessary if there are no bugs upstream.

I guess I could be convinced otherwise if there's an example where we get invalid results even though there are no upstream bugs.

nalimilan: Yeah I agree that when the upstream function is mathematically supposed to returned values in a given interval, it's more appropriate to avoid clamping so that we don't hide bugs. Cases where we need to clamp are when doing computations ourselves or when calling BLAS, as it allows for some roundoff error (and even in these cases, it would probably be safer to only allow for some tolerance when clamping, as we recently discussed elsewhere).

src/mann_whitney.jl
```diff
@@ -145,21 +147,17 @@ function pvalue(x::ExactMannWhitneyUTest; tail=:both)
...
-        elseif tail == :right
+        else # tail == :right
             pwilcox(x.U - 1, x.nx, x.ny, false)
-        else
-            throw(ArgumentError("tail=$(tail) is invalid"))
         end
     else
         # Compute exact p-value by enumerating possible ranks in the tied data
         if tail == :both
             min(1, 2 * minimum(mwuenumerate(x)))
```
nalimilan, line 156: Need to fix this one too.

devmotion: It seems due to the definition of `mwuenumerate` and `x`, `mwuenumerate(x)` is always a `Float64`. So we can use `1.0`.

nalimilan (approved):

src/HypothesisTests.jl
devmotion: It only applies to continuous distributions though, for discrete ones the values might not sum to 1 but something larger since it's P(X <= x) and P(X >= x). I guess generally the discrete case is somewhat broken since `ccdf(d, x - 1)` is just a heuristic. Ideally we want a function that returns P(X >= x) directly instead of working with P(X > x), possibly using an approach such as https://github.com/JuliaStats/Distributions.jl/issues/1483.

devmotion: So I'll revert the discrete case.

Issue #268, opened 2022-04-04
confint take either alpha everywhere or level everywhere


schlichtanders:
I just came across HypothesisTest.jl and was very surprised to see that the central function `confint` has not a unified interface.

Some use `level` others use `alpha` (=1-level) as the coverage parameter. I am very unsure why such inconsistency is the case in a well known package like HypothesisTest.jl, probably I am missing something.

I would suggest to unify this interface, to support alpha and level everywhere and compute the respective other value as needed.

nalimilan:
Good catch. `alpha` has been deprecated a long time ago, but for some reason it's still used by binomial and power divergence tests. Would you make a pull request to fix this?

PR #270, opened 2022-04-08
Remove `alpha` from docstrings

devmotion:
Fixes https://github.com/JuliaStats/HypothesisTests.jl/issues/268. It seems the implementation only supports `level` but the docstrings mentioned `alpha`.


nalimilan (approved):
Ah, great, I was afraid that the API was inconsistent, but the only places where we use `alpha` are indeed internal (or deprecated).

Issue #242, opened 2021-08-02
OneWayANOVATest seems to differ with different group sizes


andevellicus:
I'm a bit of a stats noob so I'm not sure if this is expected behavior or not, but when using GLM's ftest to do ANOVA calculations, and HypothesisTests OneWayANOVATest, I get different results with different sample sizes (see below)

MWE:

```
using GLM, StatsKit
# This set of data results in the same F-score and p-value
#=
g1 = [51, 87, 50, 48, 79, 61, 53, 54]
g2 = [82, 91, 92, 80, 52, 79, 73, 74]
g3 = [79, 84, 74, 98, 63, 83, 85, 58]
g4 = [85, 80, 65, 71, 67, 51, 63, 93]
=#

# This set of data results in a different F-score and p-value
g1 = [51, 87, 50, 48, 79, 61, 53]
g2 = [82, 91, 92, 80, 52, 79, 73, 74]
g3 = [79, 84, 74, 98, 63, 83, 85, 58]
g4 = [85, 80, 65, 71, 67, 51]

groups = [g1, g2, g3, g4]
println(OneWayANOVATest(groups...))

df1 = DataFrame(method = 0, scores=g1)
df2 = DataFrame(method = 1, scores=g2)
df3 = DataFrame(method = 2, scores=g3)
df4 = DataFrame(method = 3, scores=g4)

df = vcat(df1, df2, df3, df4)
df.method = categorical(df.method)

nullModel = lm(@formula(scores ~ 1), df)
methodModel = lm(@formula(scores ~ method), df)

println(nullModel)
println(methodModel)
ftest(nullModel.model, methodModel.model)
```

wildart:
It's a bug. The overall sample mean computed as mean of means, which is true only if samples of the equal size.

https://github.com/JuliaStats/HypothesisTests.jl/blob/dd35ae0f1c08b3ae3558e1dada9fc57e41eff1aa/src/var_equality.jl#L62-L63

PR #273, opened 2022-05-28
Fix total mean calculation in ANOVA

wildart:
Fix for #242

### DIFF
--- a/src/var_equality.jl
+++ b/src/var_equality.jl
@@ -60,7 +60,7 @@ end
 function anova(scores::AbstractVector{<:Real}...)
     Nᵢ = [length(g) for g in scores]
     Z̄ᵢ = mean.(scores)
-    Z̄ = mean(Z̄ᵢ)
+    Z̄ = sum(Iterators.flatten(scores))/sum(Nᵢ)
     SStᵢ = Nᵢ .* (Z̄ᵢ .- Z̄).^2
     SSeᵢ = sum.( (z .- z̄).^2 for (z, z̄) in zip(scores, Z̄ᵢ) )
     (Nᵢ, SStᵢ, SSeᵢ)
--- a/test/var_equality.jl
+++ b/test/var_equality.jl
@@ -31,7 +31,8 @@ using DelimitedFiles
     t = OneWayANOVATest(groups...)
     @test nobs(t) == [7, 8, 8, 6]
     @test dof(t) == (3,25)
-    @test pvalue(t) ≈ 0.072 atol=1e-3
+    @test pvalue(t) ≈ 0.07276 atol=1e-6
+    @test HypothesisTests.teststatistic(t) ≈ 2.62311 atol=1e-6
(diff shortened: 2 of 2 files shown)


src/var_equality.jl
```diff
@@ -60,7 +60,7 @@ end
 function anova(scores::AbstractVector{<:Real}...)
     Nᵢ = [length(g) for g in scores]
     Z̄ᵢ = mean.(scores)
-    Z̄ = mean(Z̄ᵢ)
+    Z̄ = sum(Iterators.flatten(scores))/sum(Nᵢ)
```
devmotion, line 63: Alternatively, one could use
```suggestion
    Z̄ = dot(Z̄ᵢ, Nᵢ) / sum(Nᵢ)
```
In a quick benchmark this seemed to be similarly fast, and usually even marginally faster.

wildart: You'd probably need a very large dataset to see the difference.

devmotion: Actually a tiny dataset is sufficient. Of course, the difference is very small but it seemed to be consistent:
```julia

julia> using Statistics, LinearAlgebra, BenchmarkTools

julia> function f(scores::AbstractVector{<:Real}...)
           Nᵢ = [length(g) for g in scores]
           Z̄ᵢ = mean.(scores)
           Z̄ = sum(Iterators.flatten(scores)) / sum(Nᵢ)
           return Nᵢ, Z̄ᵢ, Z̄
       end
f (generic function with 1 method)

julia> function g(scores::AbstractVector{<:Real}...)
           Nᵢ = [length(g) for g in scores]
           Z̄ᵢ = mean.(scores)
           Z̄ = dot(Z̄ᵢ, Nᵢ) / sum(Nᵢ)
           return Nᵢ, Z̄ᵢ, Z̄
       end
g (generic function with 1 method)

julia> scores = map(n -> rand(n), (3, 9, 12));

julia> @btime f($(scores...));
  33.893 ns (1 allocation: 64 bytes)

julia> @btime g($(scores...));
  33.687 ns (1 allocation: 64 bytes)

julia> scores = map(n -> rand(n), (3, 9, 12, 134));

julia> @btime f($(scores...));
  33.944 ns (1 allocation: 64 bytes)

julia> @btime g($(scores...));
  33.681 ns (1 allocation: 64 bytes)

julia> scores = map(n -> rand(n), (3, 9, 12, 134, 12, 4134, 1231, 122, 12, 1, 23, 58));

julia> @btime f($(scores...));
  34.070 ns (1 allocation: 64 bytes)

julia> @btime g($(scores...));
  33.781 ns (1 allocation: 64 bytes)
```

wildart: It looks like something else from Z evaluation dominates in the function. But nanoseconds 😏.

alyst:
It would be nice to merge the PR (maybe with `dot()` suggestion).

PR #282, opened 2023-01-03
Enable the two sided KG Test to deal with Ties.

FHell:
Performance is comparable, fewer allocations.

### DIFF
--- a/src/kolmogorov_smirnov.jl
+++ b/src/kolmogorov_smirnov.jl
@@ -163,9 +163,6 @@ Implements: [`pvalue`](@ref)
 """
 function ApproximateTwoSampleKSTest(x::AbstractVector{T}, y::AbstractVector{S}) where {T<:Real, S<:Real}
     n_x, n_y = length(x), length(y)
-    if n_x+n_y > length(unique([x; y]))
-        @warn("This test is inaccurate with ties")
-    end
 
     ApproximateTwoSampleKSTest(ksstats(x, y)...)
 end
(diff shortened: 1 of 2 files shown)


src/kolmogorov_smirnov.jl
```diff
@@ -194,11 +191,30 @@ end
...
-    δ = max(δp, δn)
-    (n_x, n_y, δ, δp, δn)
-end
+    all_values = [x; y]
+    sort_idx = sortperm(all_values)
+    δ_y = 1. / n_y
+    δ_x = 1. / n_x
+    δ = 0.
+    δp = 0.
+    δn = 0.
```
ararslan, line 200:
```suggestion
    δ_y = 1 / n_y
    δ_x = 1 / n_x
    δ = δp = δn = zero(δ_y)
```
It's better to avoid using `Float64` literals whenever they aren't actually necessary to avoid unnecessary promotion. Generally `length` will return an `Int64` and `inv(::Int64)` returns `Float64` so in this particular case it's more of an academic concern than a practical one but might as well use good style.

src/kolmogorov_smirnov.jl
```diff
@@ -194,11 +191,30 @@ end
...
+
+    for i in 1:(n_x + n_y)
+        if sort_idx[i] > n_x
+            δ -= δ_y
+        else
+            δ += δ_x
+        end
+
+        # only update δp/δn if the value is about to change or at the last step.
+        if ( i == n_x + n_y ) || ( all_values[sort_idx[i]] != all_values[sort_idx[i+1]] )
```
ararslan, line 210:
```suggestion
        if i == n_x + n_y || all_values[sort_idx[i]] != all_values[sort_idx[i + 1]]
```

ararslan (approved):
This is great, thanks for the contribution!

PR #288, opened 2023-03-02
Add ddof to PowerDivergenceTest and ChisqTest to allow changes to degrees of freedom

### LINKED ISSUE #266: Custom degrees of freedom argument for the `PowerDivergenceTest`
Is it possible to add a custom parameter for the degrees of freedom in the `PowerDivergenceTest` function? This may be useful for the goodness of fit tests ($G^2, \chi^2$) as the current implementation does not account for the number of parameres of the reference distribution. I think that implementation proposed in [scipy.stats.power_divergence](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.power_divergence.html) seems be the right way to do so (`ddof` parameter). Now, the user needs to remember to correct p-value after the test.

sprague252:
Resolves #266 

This change adds a new keyword parameter `ddof` to `PowerDivergenceTest` and `ChisqTest` allowing users to change the number of degrees of freedom used for calculating p-values. This works the same as the `ddof` parameter in scipy.stats.chisquare. See the documentation for that function at 

https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.chisquare.html. 

The `ddof` value defaults to 0 (no changes to the degrees of freedom). A supplied value for `ddof` decreases the number of degrees of freedom by the value of `ddof`. This is important when the distribution probabilities are calculated using parameters from measured data (e.g., mean and sample standard deviation), reducing the number of degrees of freedom.

I updated the docstrings for `PowerDivergenceTest` and `ChisqTest` and added a test for `ChisqTest` with `ddof=2` to `test/power_divergence.jl`. (It passes.)

### DIFF
--- a/src/power_divergence.jl
+++ b/src/power_divergence.jl
@@ -306,7 +306,7 @@ Implements: [`pvalue`](@ref), [`confint(::PowerDivergenceTest)`](@ref)
   * Agresti, Alan. Categorical Data Analysis, 3rd Edition. Wiley, 2013.
 """
 function PowerDivergenceTest(x::AbstractMatrix{T}; lambda::U=1.0, theta0::Vector{U} =
-	ones(length(x))/length(x), ddof::Int64=0) where {T<:Integer,U<:AbstractFloat}
+	ones(length(x))/length(x), ddof::Integer=0) where {T<:Integer,U<:AbstractFloat}
 
     nrows, ncols = size(x)
     n = sum(x)
@@ -365,19 +365,19 @@ end
 
 #PDT
 function PowerDivergenceTest(x::AbstractVector{T}, y::AbstractVector{T},
-	levels::Levels{T}; lambda::U=1.0, ddof::Int64=0) where {T<:Integer,U<:AbstractFloat}
+	levels::Levels{T}; lambda::U=1.0, ddof::Integer=0) where {T<:Integer,U<:AbstractFloat}
     d = counts(x, y, levels)
     PowerDivergenceTest(d, lambda=lambda, ddof=ddof)
 end
 
 function PowerDivergenceTest(x::AbstractVector{T}, y::AbstractVector{T}, k::T;
-	lambda::U=1.0, ddof::Int64=0) where {T<:Integer,U<:AbstractFloat}
+	lambda::U=1.0, ddof::Integer=0) where {T<:Integer,U<:AbstractFloat}
     d = counts(x, y, k)
     PowerDivergenceTest(d, lambda=lambda, ddof=ddof)
 end
 
 PowerDivergenceTest(x::AbstractVector{T}; lambda::U=1.0, theta0::Vector{U} =
-	ones(length(x))/length(x), ddof::Int64=0) where {T<:Integer,U<:AbstractFloat} =
+	ones(length(x))/length(x), ddof::Integer=0) where {T<:Integer,U<:AbstractFloat} =
 	PowerDivergenceTest(reshape(x, length(x), 1), lambda=lambda, theta0=theta0,
 	ddof=ddof)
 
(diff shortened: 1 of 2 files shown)


src/power_divergence.jl
```diff
@@ -337,18 +340,18 @@ end
 #convenience functions
 
 #PDT
-function PowerDivergenceTest(x::AbstractVector{T}, y::AbstractVector{T}, levels::Levels{T}; lambda::U=1.0) where {T<:Integer,U<:AbstractFloat}
+function PowerDivergenceTest(x::AbstractVector{T}, y::AbstractVector{T}, levels::Levels{T}; lambda::U=1.0, ddof::Int64=0) where {T<:Integer,U<:AbstractFloat}
```
nalimilan, line 343: Can you add line breaks when the line goes beyond the 92-char limit? For example:
```suggestion
function PowerDivergenceTest(x::AbstractVector{T}, y::AbstractVector{T}, levels::Levels{T};
                             lambda::U=1.0, ddof::Int64=0) where
                             {T<:Integer,U<:AbstractFloat}
```
(I know the file didn't really follow this rule but let's not make things worse;)

test/power_divergence.jl
```diff
@@ -198,4 +198,10 @@ MultinomialLRTest(x,y,(1:3,1:3))
 d = [113997 1031298
      334453 37471]
 PowerDivergenceTest(d)
+
+# Test ddof for the ChisqTest
```
nalimilan, line 202: Also test it for `PowerDivergenceTest`.

src/power_divergence.jl
```diff
@@ -237,14 +237,17 @@ end
...
 
 Perform a Power Divergence test.
 
 If `y` is not given and `x` is a matrix with one row or column, or `x` is a vector, then
 a goodness-of-fit test is performed (`x` is treated as a one-dimensional contingency
 table). In this case, the hypothesis tested is whether the population probabilities equal
-those in `theta0`, or are all equal if `theta0` is not given.
+those in `theta0`, or are all equal if `theta0` is not given. The `ddof` parameter is the 
+"delta degrees of freedom" adjustment to the number of degrees of freedom used for 
+calculation of p-values. The number of degrees of freedom is decreased by `ddof`.
```
nalimilan, line 249: Please put this after "Note that the entries...", as otherwise it breaks the logic of the "if `y`"/"if `x`".

Also, we should mention what degrees of freedom are used by default, and it would make sense to explain when it's useful to provide `ddof` (i.e. when estimating parameters).

src/power_divergence.jl
```diff
@@ -284,7 +287,7 @@ Implements: [`pvalue`](@ref), [`confint(::PowerDivergenceTest)`](@ref)
 
   * Agresti, Alan. Categorical Data Analysis, 3rd Edition. Wiley, 2013.
 """
-function PowerDivergenceTest(x::AbstractMatrix{T}; lambda::U=1.0, theta0::Vector{U} = ones(length(x))/length(x)) where {T<:Integer,U<:AbstractFloat}
+function PowerDivergenceTest(x::AbstractMatrix{T}; lambda::U=1.0, theta0::Vector{U} = ones(length(x))/length(x), ddof::Int64=0) where {T<:Integer,U<:AbstractFloat}
```
nalimilan, line 290:
```suggestion
function PowerDivergenceTest(x::AbstractMatrix{T}; lambda::U=1.0, theta0::Vector{U} = ones(length(x))/length(x), ddof::Integer=0) where {T<:Integer,U<:AbstractFloat}
```
Same elsewhere.

nalimilan (commented):
Thanks!

src/power_divergence.jl
sprague252: I will be happy to fix the line breaks. I will push them later today.

src/power_divergence.jl
sprague252: Great suggestion. I wonder is we should also allow `ddof` to be a Float?

nalimilan: Probably not at this stage, unless we can find a use case for it.

src/power_divergence.jl
sprague252: I fixed all the long lines in `power_divergence.jl`. It took me a couple of commits because I broke some lines before `where` statements, but I fixed all of them (eventually).

src/power_divergence.jl
sprague252: I changed `doff` from `Int64` to `Integer` in all function definitions.

test/power_divergence.jl
sprague252: Done.

PR #296, opened 2023-05-05
Bound binomial CI

### LINKED ISSUE #295: Lower and upper limits of CI for proportion
Should lower limit be `min(0, val)` and upper `max(val, 1)`?

for example:
```
confint(HypothesisTests.BinomialTest(0, 5), method = :agresti_coull)
```

Result
```
(**-**0.05457239484968546, 0.4890548596328611)
```

palday:
closes #295

### DIFF
--- a/src/binomial.jl
+++ b/src/binomial.jl
@@ -105,7 +105,7 @@ of the following methods. Possible values for `method` are:
 function StatsBase.confint(x::BinomialTest; level::Float64=0.95, tail=:both, method=:clopper_pearson)
     check_level(level)
 
-    if tail == :left
+    lower, upper = if tail == :left
         (0.0, StatsBase.confint(x, level=1-(1-level)*2, method=method)[2])
     elseif tail == :right
         (StatsBase.confint(x, level=1-(1-level)*2, method=method)[1], 1.0)
@@ -128,6 +128,11 @@ function StatsBase.confint(x::BinomialTest; level::Float64=0.95, tail=:both, met
     else
         throw(ArgumentError("tail=$(tail) is invalid"))
     end
+
+    # make sure we stay in [0, 1]
+    lower = max(lower, zero(lower))
+    upper = min(upper, one(upper))
+    return (lower, upper)
 end
 
 # Clopper-Pearson interval (confidence interval by inversion)
--- a/src/binomial.jl
+++ b/src/binomial.jl
@@ -105,7 +105,7 @@ of the following methods. Possible values for `method` are:
 function StatsBase.confint(x::BinomialTest; level::Float64=0.95, tail=:both, method=:clopper_pearson)
     check_level(level)
 
-    lower, upper = if tail == :left
+    if tail == :left
         (0.0, StatsBase.confint(x, level=1-(1-level)*2, method=method)[2])
     elseif tail == :right
         (StatsBase.confint(x, level=1-(1-level)*2, method=method)[1], 1.0)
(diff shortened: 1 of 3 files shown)


src/binomial.jl
```diff
@@ -128,6 +128,11 @@ function StatsBase.confint(x::BinomialTest; level::Float64=0.95, tail=:both, met
     else
         throw(ArgumentError("tail=$(tail) is invalid"))
     end
+
+    # make sure we stay in [0, 1]
```
devmotion, line 132: Most CI methods already guarantee this property, it seems. Wouldn't it be better to directly fix the relevant `ci_...` methods below (AFAICT only Wald, Agresti-Coull, and Wilson (?) might be problematic)?

ararslan: > Wouldn't it be better to directly fix the relevant `ci_...` methods below

Why would that be better? They'll all need to be in the unit interval anyway so it seems more straightforward and future-proof to just guarantee it for everything.

src/binomial.jl
```diff
@@ -154,7 +158,11 @@ function ci_agresti_coull(x::BinomialTest, alpha::Float64=0.05)
     n = x.n + q^2
     μ = (x.x + q^2/2)/n
     σ = sqrt(μ*(1-μ)/n)
-    (μ-q*σ, μ+q*σ)
+    lower, upper = (μ-q*σ, μ+q*σ)
+    # make sure we stay in [0, 1]
+    lower = max(lower, zero(lower))
+    upper = min(upper, one(upper))
+    return (lower, upper)
```
ararslan, line 165:
```suggestion
    # make sure we stay in [0, 1]
    (max(μ-q*σ, 0), min(μ+q*σ, 1))
```

src/binomial.jl
```diff
@@ -166,7 +174,11 @@ function ci_wilson(x::BinomialTest, alpha::Float64=0.05)
     μ /= denominator
     σ = sqrt(p*(1-p)/x.n + q^2/(4x.n^2))
     σ /= denominator
-    (μ-q*σ, μ+q*σ)
+    lower, upper = (μ-q*σ, μ+q*σ)
+    # make sure we stay in [0, 1]
+    lower = max(lower, zero(lower))
+    upper = min(upper, one(upper))
+    return (lower, upper)
```
ararslan, line 181:
```suggestion
    # make sure we stay in [0, 1]
    (max(μ-q*σ, 0), min(μ+q*σ, 1))
```

ararslan (commented):
IMO what you had originally that applied the clamping in only one place was preferable but I have some simplifying suggestions for what's currently implemented. They apply analogously (but only once) if you revert to the original approach.

src/binomial.jl
devmotion: The motivation for the suggestion is two-fold: It avoids unnecessary (even though cheap) computations when they are not needed and it ensures that there are no surprisingly different results when calling `ci_wald` etc. directly.

test/binomial.jl
```diff
@@ -59,6 +59,14 @@ using HypothesisTests: default_tail
...
     @test pvalue(t) ≈ 0.7320646825464584
     show(IOBuffer(), t)
+
+    # from issue #295
+    # without clamping: (-0.05457239484968546, 0.4890548596328611)
+    @test_ci_approx confint(BinomialTest(0, 5), method=:agresti_coull) (0.0, 0.4890548596328611)
+    # without clamping: (-0.15060901623063327, 0.5506090162306333)
+    @test_ci_approx confint(BinomialTest(1, 5), method=:wald) (0.0, 0.5506090162306333)
+    # without clamping: (-2.7755575615628914e-17, 0.2775327998628899)
+    @test_ci_approx confint(BinomialTest(0, 10), method=:wilson) (0.0, 0.2775327998628899)
```
devmotion, line 69: Can you also add tests in which the upper bound is adjusted? In all tests the lower bound seems to be problematic.

palday: done

devmotion (approved):
Looks good to me.

PR #297, opened 2023-05-08
Use `StatsAPI.pvalue` + `StatsAPI.HypothesisTest`

### LINKED ISSUE #290: Please extend `pvalue` from StatsAPI.jl
See https://github.com/JuliaStats/StatsAPI.jl/commit/1efca7dd5178004ec02b9b7de4f389bb1265c20a

We also want to have `pvalue` in TimeseriesSurrogates.jl, which also does hypothesis testing, but HypothesisTests.jl is too heavy of a dependency to take into.

devmotion:
Fixes #290.

The PR requires https://github.com/JuliaStats/Distributions.jl/pull/1719. Keeping these definitions in HypothesisTests would be type piracy when `pvalue` is not owned by HypothesisTests anymore. Alternatively, the definitions for `UnivariateDistribution`s could be moved to an internal `_pvalue` function but IMO they seem more generally useful, so I moved them to Distributions.


Project.toml
```diff
@@ -1,22 +1,25 @@
 name = "HypothesisTests"
 uuid = "09f84164-cd44-5f33-b23f-e6b0d136a0d5"
-version = "0.10.13"
+version = "0.11.0"
```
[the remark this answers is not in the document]
ararslan: Ah, got it. Thanks!

src/power_divergence.jl
```diff
@@ -67,18 +67,18 @@ one of the following methods. Possible values for `method` are:
   * Gold, R. Z. Tests Auxiliary to ``χ^2`` Tests in a Markov Chain. Annals of
     Mathematical Statistics, 30:56-74, 1963.
 """
-function StatsBase.confint(x::PowerDivergenceTest; level::Float64=0.95,
-                           tail::Symbol=:both, method::Symbol=:auto, correct::Bool=true,
-                           bootstrap_iters::Int64=10000, GC::Bool=true)
+function StatsAPI.confint(x::PowerDivergenceTest; level::Float64=0.95,
+                 tail::Symbol=:both, method::Symbol=:auto, correct::Bool=true,
+                 bootstrap_iters::Int64=10000, GC::Bool=true)
```
ararslan, line 72:
```suggestion
function StatsAPI.confint(x::PowerDivergenceTest; level::Float64=0.95,
                          tail::Symbol=:both, method::Symbol=:auto, correct::Bool=true,
                          bootstrap_iters::Int64=10000, GC::Bool=true)
```
Looks like alignment got messed up somehow

test/bartlett.jl
```diff
@@ -1,8 +1,9 @@
 using HypothesisTests
 using Test
-using StatsBase
 using DelimitedFiles
 
+using StatsAPI: dof, nobs
```
ararslan, line 5: This change isn't actually necessary since StatsBase reexports those

devmotion: I guess I changed it while simplifying other imports and uses of StatsBase since no functionality or definition of StatsBase is needed here.

src/bartlett.jl
```diff
@@ -45,12 +45,12 @@ function BartlettTest(X::AbstractMatrix, Y::AbstractMatrix)
     return BartlettTest(L′, p, nx, ny)
 end
 
-StatsBase.nobs(B::BartlettTest) = (B.nx, B.ny)
-StatsBase.dof(B::BartlettTest) = div(B.p * (B.p + 1), 2)
+StatsAPI.nobs(B::BartlettTest) = (B.nx, B.ny)
```
ararslan, line 48: If we're defining methods for these for some tests, we should reexport them

devmotion: I exported them in https://github.com/JuliaStats/HypothesisTests.jl/pull/297/commits/d83c2fd2da5b29676c7a4d7c984d0f044f2f860a. Thereby I could also address your comment in https://github.com/JuliaStats/HypothesisTests.jl/pull/297#discussion_r1187655952: The imports are removed completely.

ararslan: image

src/HypothesisTests.jl
```diff
@@ -25,76 +25,23 @@
 module HypothesisTests
 
 using Statistics, Random, LinearAlgebra
-using Distributions, Roots, StatsBase
```
ararslan, line 28: Why change this?

devmotion: Initially it seemed as if we could get rid of the StatsBase dependency completely. It turned out that this is not the case but the five functions loaded below are needed + `StatsBase.PValue` for printing. Nevertheless, as with the imports of Combinatorics and Rmath in the lines below, explicitly importing these five functions seemed a bit clearer.

devmotion:
Bump :slightly_smiling_face:

src/HypothesisTests.jl
ararslan: I don't personally find listing out every symbol to be particularly clarifying or useful but I know not everyone shares that opinion. :) In this case I'd prefer not to change the StatsBase bit as part of this PR as it seems like an unrelated change but I don't feel too strongly.

devmotion: I don't have a completely strong opinion on this matter either. I wasn't 100% sure what changes you had in mind exactly, but I reverted this line and some related changes below. Is it better now?

ararslan (approved):

src/HypothesisTests.jl
ararslan: It looks ✨ a m a z i n g ✨, apologies for the pedantry and for the delay, this got lost amidst a bunch of other notifications

Project.toml
```diff
@@ -1,22 +1,25 @@
...
 Rmath = "79098fc4-a85e-5d69-aa6a-4863f24498fa"
 Roots = "f2b01f46-fcfa-551c-844a-d8ac1e96c665"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+StatsAPI = "82ae8749-77ed-4fe6-ae5f-f523153014b0"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 
 [compat]
 Combinatorics = "1"
-Distributions = "0.19, 0.20, 0.21, 0.22, 0.23, 0.24, 0.25"
+Distributions = "0.25.90"
```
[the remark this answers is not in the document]
devmotion: Yes, this PR requires https://github.com/JuliaStats/Distributions.jl/pull/1719 (as mentioned in the OP an alternative would be to commit type piracy and keep the methods of `pvalue` for `UnivariateDistribution`s), and this is the first version of Distributions that includes it.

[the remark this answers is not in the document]
ararslan: Ah, of course you addressed this in the OP, should have reread that too!

PR #302, opened 2023-07-02
Use `allunique`

devmotion:
This PR contains a minor optimization: Instead of computing `unique(x)` and comparing its length with the length of `x`, it just checks `allunique(x)`.


nalimilan:
Thanks!

PR #308, opened 2023-08-18
Master

VincentPoupart:
Hi! This my first pull request ever!

I just corrected some iterators and one comparaison sign.

Vincent


src/HypothesisTests.jl
```diff
@@ -110,7 +110,7 @@ function show_params(io::IO, test::T, ident="") where T<:HypothesisTest
         lengths = [length(string(T.names[i])) for i in fieldidx]
         maxlen = maximum(lengths)
 
-        for i = 1:length(fieldidx)
+        for i in eachindex(fieldidx)
```
devmotion, line 113: One should be a bit careful here since `i` is used both for indexing into `fieldidx` and `maxlen-lengths`. Here, I think it would be easier to avoid indexing altogether and write
```suggestion
        for (fieldidx_i, maxlen-lengths_i) in zip(fieldidx, maxlen-lengths)
```
and use these variables in the body of the for-loop.

src/anderson_darling.jl
```diff
@@ -220,7 +220,7 @@ function pvalueasym(x::KSampleADTest)
     n = length(PV)
     Tm = zeros(n)
     for i in 1:n
-        A = [ sqm[i]^p for i = 1:length(sqm), p = 0:3 ] # fit in cubic
+        A = [ sqm[i]^p for i in eachindex(sqm), p = 0:3 ] # fit in cubic
```
devmotion, line 223: Maybe just
```suggestion
        A = [ sqm_i^p for sqm_i in sqm, p = 0:3 ] # fit in cubic
```
?

src/HypothesisTests.jl
ararslan: I don't think this suggestion is correct: `maxlen-lengths` will error since the former is a scalar and the latter is an array, and `maxlen-lengths_i` is not a valid identifier. It's likely this method doesn't have tests, which is why CI didn't complain about it.

ararslan: > One should be a bit careful here since `i` is used both for indexing into `fieldidx` and `maxlen-lengths`.

That's true in general, but in this case by construction the arrays are plain `Vector`s with the same length, so indexing both with `i` should be safe.

PR #158, opened 2019-05-17
Rename some tests for consistency

nalimilan:
MultinomialLRT and the three *HotellingT2 were the only tests whose names did not end with "Test".

PR #318, opened 2024-06-04
Fix typo in `HotellingT2Test` name

nalimilan:
Introduced by #158.

### DIFF
--- a/src/hotelling.jl
+++ b/src/hotelling.jl
@@ -2,12 +2,12 @@
 
 export OneSampleHotellingT2Test, EqualCovHotellingT2Test, UnequalCovHotellingT2Test
 
-abstract type HotellingT2TestTest <: HypothesisTest end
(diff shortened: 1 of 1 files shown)


ararslan:
Thank you! Do you think it'd be worth adding a deprecation for the old name? Though it looks like no registered packages are referencing it[^1] so it should be safe to just change.

[^1]: https://juliahub.com/ui/Search?type=symbols&q=HotellingT2TestTest

ararslan (approved):

nalimilan:
Given how ridiculous the old name is, I'd say we can assume nobody uses it or it would have been reported. :-)

ararslan:
You mean you've never used Hotelling's $T^2$ test test? It's twice as powerful as Hotelling's $T^2$ test, but not nearly as powerful as Hotelling's $T^2$ test test test.

PR #330, opened 2025-02-09
remove psignrank from Rmath

ArnoStrouwen:
This PR attempts to reimplement `psignrank` in pure Julia, to get rid of any GNU license issues with the R version @andreasnoack .
The Rmath version returns the exact correct decimal number, while my version has some slight floating point error.
I'm not sure why this is because, the Rmath version also containts floating point operations:
https://github.com/JuliaStats/Rmath-julia/blob/6f2d37ff112914d65559bc3e0035b325c11cf361/src/signrank.c#L161
```jl
julia> @btime HypothesisTests.psignrank.(0:sum(0:n), n, true)
  1.030 μs (38 allocations: 5.41 KiB)
11-element Vector{Float64}:
 0.0625
 0.12500000000000003
 0.18750000000000003
 0.3125
 0.4375
 0.5625000000000001
 0.6875000000000001
 0.8125000000000001
 0.8749999999999999
 0.9375000000000001
 1.0
 
 julia> @btime Rmath.psignrank.(0:sum(0:n), n, true)
  1.950 μs (5 allocations: 256 bytes)
11-element Vector{Float64}:
 0.0625
 0.125
 0.1875
 0.3125
 0.4375
 0.5625
 0.6875
 0.8125
 0.875
 0.9375
 1.0


julia> @btime HypothesisTests.psignrank.(0:sum(0:n), n, false)
  1.130 μs (38 allocations: 5.41 KiB)
11-element Vector{Float64}:
 1.0
 0.9375000000000001
 0.8749999999999999
 0.8125000000000001
 0.6875000000000001
 0.5625000000000001
 0.4375
 0.3125
 0.18750000000000003
 0.12500000000000003
 0.0625


julia> @btime Rmath.psignrank.((0:sum(0:n)).-1, n, false)
  1.960 μs (6 allocations: 288 bytes)
11-element Vector{Float64}:
 1.0
 0.9375
 0.875
 0.8125
 0.6875
 0.5625
 0.4375
 0.3125
 0.1875
 0.125
 0.0625
```

### DIFF
--- a/src/wilcoxon.jl
+++ b/src/wilcoxon.jl
@@ -85,8 +85,8 @@ Perform a Wilcoxon exact signed rank U test of the null hypothesis that the dist
 `x` (or the difference `x - y` if `y` is provided) has zero median against the alternative
 hypothesis that the median is non-zero.
 
-When there are no tied ranks, the exact p-value is computed using the `psignrank` function
-from the `Rmath` package. In the presence of tied ranks, a p-value is computed by exhaustive
+When there are no tied ranks, the exact p-value is computed using dynamic programming.
+In the presence of tied ranks, a p-value is computed by exhaustive
 enumeration of permutations, which can be very slow for even moderately sized data sets.
 
 Implements: [`pvalue`](@ref), [`confint`](@ref)
@@ -144,12 +144,12 @@ function StatsAPI.pvalue(x::ExactSignedRankTest; tail=:both)
             if x.W <= n * (n + 1)/4
                 2 * psignrank(x.W, n, true)
             else
-                2 * psignrank(x.W - 1, n, false)
+                2 * psignrank(x.W, n, false)
             end
         elseif tail == :left
             psignrank(x.W, n, true)
         else
-            psignrank(x.W - 1, n, false)
+            psignrank(x.W, n, false)
         end
     else
         # Compute exact p-value by enumerating all possible ranks in the tied data
@@ -161,6 +161,27 @@ function StatsAPI.pvalue(x::ExactSignedRankTest; tail=:both)
             last(signedrankenumerate(x))
         end
     end
+end
+
+function psignrank(W::Union{Float64, Int}, n::Int, lower_tail::Bool)
+    W = round(Int, W)
+    max_W =  (n * (n + 1)) >> 1
+    if !lower_tail
+        return psignrank(max_W - W, n, true)
+    end
+    if W == max_W
+        return 1.0
+    elseif W > max_W >> 1
+        return 1 - psignrank(max_W - W - 1, n, true)
+    end
+    DP = zeros(Int, W + 1)
+    DP[W+1] = 1
+    for j in 1:n
+        for i in 1:(W+1-j)
+            DP[i] += DP[i + j]
+        end
+    end
+    return sum(ldexp.(float.(DP), -n))
 end
 
 StatsAPI.confint(x::ExactSignedRankTest; level::Real=0.95, tail=:both) = calculate_ci(x.vals, level, tail=tail)
--- a/src/wilcoxon.jl
+++ b/src/wilcoxon.jl
@@ -181,7 +181,7 @@ function psignrank(W::Union{Float64, Int}, n::Int, lower_tail::Bool)
             DP[i] += DP[i + j]
         end
     end
-    return sum(ldexp.(float.(DP), -n))
+    return sum(Base.Fix2(ldexp, -n) ∘ float, DP)
 end
 
 StatsAPI.confint(x::ExactSignedRankTest; level::Real=0.95, tail=:both) = calculate_ci(x.vals, level, tail=tail)
--- a/src/HypothesisTests.jl
+++ b/src/HypothesisTests.jl
@@ -27,7 +27,7 @@ module HypothesisTests
 using Statistics, Random, LinearAlgebra
 using Distributions, Roots, StatsBase
 using Combinatorics: combinations, permutations
-using Rmath: pwilcox, psignrank
+using Rmath: pwilcox
 using Printf: @printf
 
 import StatsAPI
--- a/test/wilcoxon.jl
+++ b/test/wilcoxon.jl
@@ -1,5 +1,6 @@
 using HypothesisTests, Test
 using HypothesisTests: default_tail
+using Rmath
 
 @testset "Wilcoxon" begin
 @testset "Basic exact test" begin
@@ -72,5 +73,29 @@ end
     @test isapprox(@inferred(confint(ApproximateSignedRankTest(x)))[2], 15.5, atol=1e-4)
     @test isapprox(@inferred(confint(SignedRankTest(x); tail=:left))[1], 4.45, atol=1e-4)
     @test isapprox(@inferred(confint(SignedRankTest(x); tail=:right))[2], 14.45, atol=1e-4)
+end
+
+@testset "Comparison psignrank with Rmath" begin
+    n = 4
+    for W in 0.0:sum(0:n)
+        @test isapprox(@inferred(HypothesisTests.psignrank(W, n, true)),
+                                 Rmath.psignrank.(W, n, true),
+                                 atol=1e-10)
+    end
+    for W in 0.0:sum(0:n)
+        @test isapprox(@inferred(HypothesisTests.psignrank(W, n, false)),
+                                 Rmath.psignrank(W-1, n, false),
+                                 atol=1e-10)
+    end
+    for W in 0:sum(0:n)
+        @test isapprox(@inferred(HypothesisTests.psignrank(W, n, true)),
+                                 Rmath.psignrank.(W, n, true),
+                                 atol=1e-10)
+    end
+    for W in 0:sum(0:n)
+        @test isapprox(@inferred(HypothesisTests.psignrank(W, n, false)),
+                                 Rmath.psignrank(W-1, n, false),
+                                 atol=1e-10)
+    end
 end
 end


ArnoStrouwen:
Julia 1.3 complains about the use of `begin`, or we can up the CI bound to something more modern?

src/wilcoxon.jl
```diff
@@ -163,6 +163,27 @@ function StatsAPI.pvalue(x::ExactSignedRankTest; tail=:both)
...
+    DP[begin] = one(int_type)
+    for j in one(int_type):n
+        for i in max_W:-one(int_type):one(int_type)
+            if i - j >= one(int_type)
+                DP[i] += DP[i - j]
+            end
+        end
+    end
+    if lower_tail
+        return exp(log(sum(@view DP[begin:W+one(int_type)])) - n*log(two)) # exp log to avoid 2^n
```
devmotion, line 181: Isn't this just

```suggestion
        return ldexp(sum(@view DP[1:(W + 1)])), -n)
```

src/wilcoxon.jl
```diff
@@ -163,6 +163,27 @@ function StatsAPI.pvalue(x::ExactSignedRankTest; tail=:both)
...
+    else
+        return exp(log(sum(@view DP[W+one(int_type):end])) - n*log(two))
```
devmotion, line 183:
```suggestion
        return ldexp(sum(@view DP[(W + 1):end]), -n)
```

src/wilcoxon.jl
```diff
@@ -163,6 +163,27 @@ function StatsAPI.pvalue(x::ExactSignedRankTest; tail=:both)
     end
 end
 
+function psignrank(W, n, lower_tail)
```
devmotion, line 166: Might be good to add a docstring and constrain the arguments?

```suggestion
function psignrank(W::T, n::T, lower_tail::Bool) where {T<:Integer}
```
(promotion could be done separately if necessary)

src/wilcoxon.jl
```diff
@@ -163,6 +163,27 @@ function StatsAPI.pvalue(x::ExactSignedRankTest; tail=:both)
...
+    int_type = typeof(n)
+    W = int_type(W)
+    two = one(int_type) + one(int_type)
+    max_W =  n * (n + one(int_type)) ÷ two + one(int_type)
+    DP = zeros(int_type(n), max_W)
+    DP[begin] = one(int_type)
```
devmotion, line 172: Since `DP` is a `Vector`, you could just use
```suggestion
    DP[1] = one(int_type)
```

src/wilcoxon.jl
```diff
@@ -163,6 +163,27 @@ function StatsAPI.pvalue(x::ExactSignedRankTest; tail=:both)
     end
 end
 
+function psignrank(W, n, lower_tail)
+    int_type = typeof(n)
+    W = int_type(W)
+    two = one(int_type) + one(int_type)
+    max_W =  n * (n + one(int_type)) ÷ two + one(int_type)
```
devmotion, line 170: Is it actually possible to construct a `Vector` with non-`Int` length? I assume this could just always be an `Int` as it doesn't have any ilpact on the calculated values.

src/wilcoxon.jl
```diff
@@ -163,6 +163,27 @@ function StatsAPI.pvalue(x::ExactSignedRankTest; tail=:both)
...
+    DP = zeros(int_type(n), max_W)
+    DP[begin] = one(int_type)
+    for j in one(int_type):n
+        for i in max_W:-one(int_type):one(int_type)
+            if i - j >= one(int_type)
+                DP[i] += DP[i - j]
+            end
+        end
```
devmotion, line 178: This could be simplified to

```suggestion
        for i in max_W:-1:(j + 1)
            DP[i] += DP[i - j]
        end
```

devmotion (commented):
I wonder if the function should be added to StatsFuns, to make it easier and to reduce dependencies in case you want to use it in some other package and since it's currently the main package with Julia implementations of Rmath functions.

src/wilcoxon.jl
devmotion: Given that the test only allows for `n::Int` maybe at least initially we could just support `W::Int` and `n::Int`.

ArnoStrouwen:
> I wonder if the function should be added to StatsFuns, to make it easier and to reduce dependencies in case you want to use it in some other package and since it's currently the main package with Julia implementations of Rmath functions.

It should really go to Distributions, since this is calculating the null distribution, but I'm too lazy to implement all the other functions. I would leave it here for now and create an issue for upstreaming proper Wilcoxon related distributions.

src/wilcoxon.jl
ArnoStrouwen: I agree that W only makes sense as integer in the exact test with no ties. However, the caller of `psignrank` seems to provide a float, I assume to accommodate for approximate test/ties. 

We could just round the float to the nearest integer before we call `psignrank`.

src/wilcoxon.jl
ArnoStrouwen: I am not familiar with `ldexp`.

src/wilcoxon.jl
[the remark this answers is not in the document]
ArnoStrouwen: Yes, there are further optimizations to make here.
Also, for upper tail, you can transform it to a lower tail because the distribution is symmetric.
If you calculate a lower tail with more than 50% chance, you can again use symmetry.

src/wilcoxon.jl
devmotion: `ldexp(x, n)` returns `x * 2^n`: https://docs.julialang.org/en/v1/base/math/#Base.Math.ldexp

ArnoStrouwen: Ok, that seems like the calculation we need.
However:
```jl
julia> ldexp(5,2)
ERROR: MethodError: no method matching ldexp(::Int64, ::Int64)

julia> ldexp(5.0,2)
20.0  
```

devmotion: We could use
```suggestion
return ldexp(float(sum(@view DP[(W + 1):end])), -n)
```

ArnoStrouwen:
Not bad:
```jl
julia> n = 4
       @btime HypothesisTests.psignrank.(0:sum(0:n), n, true)
  401.493 ns (25 allocations: 1.08 KiB)
11-element Vector{Float64}:
 0.0625
 0.125
 0.1875
 0.3125
 0.4375
 0.5625
 0.6875
 0.8125
 0.875
 0.9375
 1.0

julia> @btime HypothesisTests.psignrank.(0:sum(0:n), n, false)
  422.111 ns (25 allocations: 1.08 KiB)
11-element Vector{Float64}:
 1.0
 0.9375
 0.875
 0.8125
 0.6875
 0.5625
 0.4375
 0.3125
 0.1875
 0.125
 0.0625
```

src/wilcoxon.jl
```diff
@@ -142,14 +142,14 @@ function StatsAPI.pvalue(x::ExactSignedRankTest; tail=:both)
         # Compute exact p-value using method from Rmath, which is fast but cannot account for ties
         if tail == :both
             if x.W <= n * (n + 1)/4
-                2 * psignrank(x.W, n, true)
+                2 * psignrank(typeof(n)(x.W), n, true)
```
devmotion, line 145: Since `n` is always an `Int` here, this just means
```suggestion
                2 * psignrank(Int(x.W), n, true)
```
it seems? But AFAICT that would error if `x.W` is non-integer (it's stored as a `Float64` it seems).

What's the current Rmath implementation actually doing for non-integer values of `W`? Probably we just have to round, either here or at the top of the function body of `psignrank`. Since it has to be done in multiple places it might be easier to add a `round(Int, ...)`/`floor(Int, ...)/...` statement to `psignrank`.

src/wilcoxon.jl
```diff
@@ -163,6 +163,26 @@ function StatsAPI.pvalue(x::ExactSignedRankTest; tail=:both)
     end
 end
 
+function psignrank(W::T, n::T, lower_tail::Bool) where {T<:Integer}
```
devmotion, line 166: IMO we should just keep it simple for now, in HypothesisTests we only need
```suggestion
function psignrank(W::Float64, n::Int, lower_tail::Bool)
```
or
```suggestion
function psignrank(W::Int, n::Int, lower_tail::Bool)
```
it seems, depending on whether we round above or inside of `psignrank`.

src/wilcoxon.jl
```diff
@@ -163,6 +163,26 @@ function StatsAPI.pvalue(x::ExactSignedRankTest; tail=:both)
...
+    max_W =  n * (n + one(T)) ÷ (one(T) + one(T))
```
devmotion, line 167: `>>` should work even with generic integer types (but IMO we shouldn't be worried about them at this point anyway and they are not tested at all):
```suggestion
    max_W =  (n * (n + 1)) >> 1
```

src/wilcoxon.jl
```diff
@@ -163,6 +163,26 @@ function StatsAPI.pvalue(x::ExactSignedRankTest; tail=:both)
...
+    if !lower_tail
+        return psignrank(max_W - W, n, !lower_tail)
+    end
+    if W == max_W
+        return 1.0
```
devmotion, line 172: It seems this already assumes we only work with `Int` and `Float64`? Otherwise it would cause a type instability.

src/wilcoxon.jl
```diff
@@ -163,6 +163,26 @@ function StatsAPI.pvalue(x::ExactSignedRankTest; tail=:both)
...
+    elseif W > max_W ÷ (one(T) + one(T))
+        return 1.0 - psignrank(W + one(T), n, !lower_tail)
+    end
+    DP = zeros(T, W + 1)
```
devmotion, line 176:
```suggestion
    DP = zeros(Int, W + 1)
```

src/wilcoxon.jl
```diff
@@ -163,6 +163,26 @@ function StatsAPI.pvalue(x::ExactSignedRankTest; tail=:both)
...
+    DP[W+1] = one(T)
```
devmotion, line 177:
```suggestion
    DP[W+1] = 1
```

src/wilcoxon.jl
```diff
@@ -163,6 +163,26 @@ function StatsAPI.pvalue(x::ExactSignedRankTest; tail=:both)
...
+    for j in 1:n
+        for i in 1:W+1-j
```
devmotion, line 179:
```suggestion
        for i in 1:(W+1-j)
```

src/wilcoxon.jl
```diff
@@ -163,6 +163,26 @@ function StatsAPI.pvalue(x::ExactSignedRankTest; tail=:both)
...
 end
 
+function psignrank(W::T, n::T, lower_tail::Bool) where {T<:Integer}
+    max_W =  n * (n + one(T)) ÷ (one(T) + one(T))
+    if !lower_tail
+        return psignrank(max_W - W, n, !lower_tail)
+    end
+    if W == max_W
+        return 1.0
+    elseif W > max_W ÷ (one(T) + one(T))
```
devmotion, line 173:
```suggestion
    elseif W > (max_W >> 1)
```

devmotion:
Can you also add tests that compare the `psignrank` implementation to `Rmath.psignrank`?

devmotion:
> It should really go to Distributions, since this is calculating the null distribution,

I'm not so sure about this argument. StatsFuns contains distribution functions, densities etc. that are used by Distributions (e.g. for Normal, Beta, Gamma, and student t distribution) but can also be used by other packages independently of Distributions.

src/wilcoxon.jl
ArnoStrouwen: If we are going to assume Int64 and Float64, I would prefer to just `÷2` to keep it readable for statisticians.

devmotion: `>>` is already used on the master branch a few lines above: https://github.com/JuliaStats/HypothesisTests.jl/blob/1cbeccd8e31bc26f6d0f44301243b23eba558b05/src/wilcoxon.jl#L127

Probably doesn't matter in this line though given my suggestion below (which would get rid of it here). Generally, it seems the generated code is slightly simpler with `>>` (not too surprising I assume) but they're both very simple:
```julia
julia> f(x) = x >> 1
f (generic function with 1 method)

julia> g(x) = x ÷ 2
g (generic function with 1 method)

julia> @code_native f(4)
        .section        __TEXT,__text,regular,pure_instructions
        .build_version macos, 15, 0
        .globl  _julia_f_6454                   ; -- Begin function julia_f_6454
        .p2align        2
_julia_f_6454:                          ; @julia_f_6454
; Function Signature: f(Int64)
; ┌ @ REPL[9]:1 within `f`
; %bb.0:                                ; %top
; │ @ REPL[9] within `f`
        ;DEBUG_VALUE: f:x <- $x0
        ;DEBUG_VALUE: f:x <- $x0
; │ @ REPL[9]:1 within `f`
; │┌ @ int.jl:534 within `>>` @ int.jl:527
        asr     x0, x0, #1
; ││ @ int.jl:534 within `>>`
; ││┌ @ essentials.jl:796 within `ifelse`
        ret
; └└└
                                        ; -- End function
.subsections_via_symbols

julia> @code_native g(4)
        .section        __TEXT,__text,regular,pure_instructions
        .build_version macos, 15, 0
        .globl  _julia_g_6610                   ; -- Begin function julia_g_6610
        .p2align        2
_julia_g_6610:                          ; @julia_g_6610
; Function Signature: g(Int64)
; ┌ @ REPL[10]:1 within `g`
; %bb.0:                                ; %pass
; │ @ REPL[10] within `g`
        ;DEBUG_VALUE: g:x <- $x0
        ;DEBUG_VALUE: g:x <- $x0
; │ @ REPL[10]:1 within `g`
; │┌ @ int.jl:295 within `div`
        cmp     x0, #0
        cinc    x8, x0, lt
        asr     x0, x8, #1
        ret
; └└
                                        ; -- End function
.subsections_via_symbols
```

src/wilcoxon.jl
```diff
@@ -163,6 +163,27 @@ function StatsAPI.pvalue(x::ExactSignedRankTest; tail=:both)
...
+        return 1 - psignrank(max_W - W - 1, n, true)
+    end
+    DP = zeros(Int, W + 1)
+    DP[W+1] = 1
+    for j in 1:n
+        for i in 1:(W+1-j)
+            DP[i] += DP[i + j]
+        end
+    end
+    return sum(ldexp.(float.(DP), -n))
```
devmotion, line 184: We can avoid broadcasting and reduce allocations here:
```suggestion
    return sum(Base.Fix2(ldexp, -n) ∘ float, DP)
```

ArnoStrouwen: Map reduce also gets rid of the allocation, not sure what the exact difference is.
```suggestion
    mapreduce(x->ldexp(float(x), -n), +, DP)
```

PR #331, opened 2025-02-16
remove pwilcox from Rmath

ArnoStrouwen:
Similar to https://github.com/JuliaStats/HypothesisTests.jl/pull/330 but with a way trickier recursion.

There are still a few parts that have to be optimized.
For example, there are `sum(1:k)-1` useless rows at the top.
I think it should be possible to do the recursion, such that these rows are recovered,
but could not figure out the indexing.
Capture
Once that is done, the allocations in the final part of the function are also easier to address.

Finally, I think there are also optimizations possible if `nx `or `ny` is larger.

### DIFF
--- a/src/prototype.jl
+++ b/src/prototype.jl
@@ -1,0 +1,44 @@
+using HypothesisTests
+using BenchmarkTools
+
+nx = 3
+ny = 4
+
+U = 9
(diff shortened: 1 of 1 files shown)


ArnoStrouwen:
Superseded by https://github.com/JuliaStats/HypothesisTests.jl/pull/334.

PR #332, opened 2025-03-07
Allow vector of groups as an input for variance equality tests

alyst:
Currently, one-way ANOVA and other variance equality tests require that each group of values is passed as a separate argument (`OneWayAnova([1, 2], [1, 2, 3])`).
When the user has a vector of groups (`groups = [[1, 2], [1, 2, 3]]`), it is therefore required to splat it when passing to the tests (`OneWayAnova(groups...)`).
That adds unnecessary overhead to the Julia compiler, especially when the number of groups is large -- the execution becomes extremely slow.

The PR changes the var-equality tests to accept the vector of groups as the input.
The "splatting" variants of the functions are converted to call the vector-based ones.

The PR also includes the modified version of the #273 fix -- but I will also welcome if that PR is merged first.

### DIFF
--- a/src/var_equality.jl
+++ b/src/var_equality.jl
@@ -57,8 +57,8 @@ function Base.show(io::IOContext, t::VarianceEqualityTest)
     end
 end
 
-function anova(scores::AbstractVector{<:Real}...)
-    Nᵢ = [length(g) for g in scores]
+function anova(scores::AbstractVector{<:AbstractVector})
+    Nᵢ = length.(scores)
     Z̄ᵢ = mean.(scores)
     Z̄ = dot(Z̄ᵢ, Nᵢ) / sum(Nᵢ)
     SStᵢ = Nᵢ .* abs2.(Z̄ᵢ .- Z̄)
@@ -66,7 +66,10 @@ function anova(scores::AbstractVector{<:Real}...)
     (Nᵢ, SStᵢ, SSeᵢ)
 end
 
+anova(scores::AbstractVector{<:Real}...) = anova([scores...])
+
 """
+    OneWayANOVATest(groups::AbstractVector{AbstractVector{<:Real}})
     OneWayANOVATest(groups::AbstractVector{<:Real}...)
 
 Perform one-way analysis of variance test of the hypothesis that that the `groups`
@@ -86,14 +89,17 @@ Implements: [`pvalue`](@ref)
   * [One-way analysis of variance on Wikipedia
     ](https://en.wikipedia.org/wiki/One-way_analysis_of_variance)
 """
-function OneWayANOVATest(groups::AbstractVector{<:Real}...)
-    Nᵢ, SStᵢ, SSeᵢ = anova(groups...)
+function OneWayANOVATest(groups::AbstractVector{<:AbstractVector})
+    Nᵢ, SStᵢ, SSeᵢ = anova(groups)
     k = length(Nᵢ)
     VarianceEqualityTest{FDist}(Nᵢ, SStᵢ, SSeᵢ, k-1, sum(Nᵢ)-k,
         ("One-way analysis of variance (ANOVA) test", "Means", "F"))
 end
 
+OneWayANOVATest(groups::AbstractVector{<:Real}...) = OneWayANOVATest([groups...])
+
 """
+    LeveneTest(groups::AbstractVector{AbstractVector{<:Real}}; scorediff=abs, statistic=mean)
     LeveneTest(groups::AbstractVector{<:Real}...; scorediff=abs, statistic=mean)
 
 Perform Levene's test of the hypothesis that that the `groups` variances are equal.
@@ -134,17 +140,21 @@ Implements: [`pvalue`](@ref)
   * [Levene's test on Wikipedia
     ](https://en.wikipedia.org/wiki/Levene%27s_test)
 """
-function LeveneTest(groups::AbstractVector{<:Real}...; scorediff=abs, statistic=mean)
+function LeveneTest(groups::AbstractVector{<:AbstractVector}; scorediff=abs, statistic=mean)
     # calculate scores
     Zᵢⱼ = [scorediff.(g .- statistic(g)) for g in groups]
     # anova
-    Nᵢ, SStᵢ, SSeᵢ = anova(Zᵢⱼ...)
+    Nᵢ, SStᵢ, SSeᵢ = anova(Zᵢⱼ)
     k = length(Nᵢ)
     VarianceEqualityTest{FDist}(Nᵢ, SStᵢ, SSeᵢ, k-1, sum(Nᵢ)-k,
         ("Levene's test", "Variances", "W"))
 end
 
+LeveneTest(groups::AbstractVector{<:Real}...; kwargs...) =
+    LeveneTest([groups...]; kwargs...)
+
 """
+    BrownForsytheTest(groups::AbstractVector{AbstractVector{<:Real}})
     BrownForsytheTest(groups::AbstractVector{<:Real}...)
 
 The Brown–Forsythe test is a statistical test for the equality of `groups` variances.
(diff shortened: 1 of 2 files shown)


src/var_equality.jl
```diff
@@ -57,16 +57,19 @@ function Base.show(io::IOContext, t::VarianceEqualityTest)
     end
 end
 
-function anova(scores::AbstractVector{<:Real}...)
-    Nᵢ = [length(g) for g in scores]
+function anova(scores::AbstractVector{<:AbstractVector})
+    Nᵢ = length.(scores)
```
devmotion, line 61: No need for broadcasting here:
```suggestion
    Nᵢ = map(length, scores)
```

src/var_equality.jl
```diff
@@ -57,16 +57,19 @@ function Base.show(io::IOContext, t::VarianceEqualityTest)
     end
 end
 
-function anova(scores::AbstractVector{<:Real}...)
-    Nᵢ = [length(g) for g in scores]
+function anova(scores::AbstractVector{<:AbstractVector})
```
devmotion, line 60: It seems a (minor but still) regression to allocate a vector in the multi-arg version. Maybe we could just allow iterators here, something like
```suggestion
function anova(scores)
    if isempty(scores) || !all(x -> x isa AbstractVector{<:Real}, scores)
        throw(ArgumentError("`anova` requires a non-empty collection of vectors of `Real` numbers"))
    end
```

src/var_equality.jl
```diff
@@ -57,16 +57,19 @@ function Base.show(io::IOContext, t::VarianceEqualityTest)
...
+    Nᵢ = length.(scores)
     Z̄ᵢ = mean.(scores)
-    Z̄ = mean(Z̄ᵢ)
-    SStᵢ = Nᵢ .* (Z̄ᵢ .- Z̄).^2
-    SSeᵢ = sum.( (z .- z̄).^2 for (z, z̄) in zip(scores, Z̄ᵢ) )
+    Z̄ = dot(Z̄ᵢ, Nᵢ) / sum(Nᵢ)
+    SStᵢ = Nᵢ .* abs2.(Z̄ᵢ .- Z̄)
+    SSeᵢ = [sum(z -> abs2(z - z̄ᵢ), scoresᵢ) for (scoresᵢ, z̄ᵢ) in zip(scores, Z̄ᵢ)]
```
devmotion, line 65: Could use a `map` here as well? This would also be better for tuple or static array input.
```suggestion
    SSeᵢ = map((scoresᵢ, z̄ᵢ) -> sum(z -> abs2(z - z̄ᵢ), scoresᵢ), scores, Z̄ᵢ)
```

src/var_equality.jl
```diff
@@ -57,16 +57,19 @@ function Base.show(io::IOContext, t::VarianceEqualityTest)
...
     (Nᵢ, SStᵢ, SSeᵢ)
 end
 
+anova(scores::AbstractVector{<:Real}...) = anova([scores...])
```
devmotion, line 69:
```suggestion
anova(scores::AbstractVector{<:Real}...) = anova(scores)
```

src/var_equality.jl
alyst: Actually, `Nᵢ` has to be a vector -- that's what `VarianceEqualityTest` struct expects

src/var_equality.jl
alyst: Same thing as for `Nᵢ` -- `SSeᵢ` has to be a vector

src/var_equality.jl
alyst: Thank you for the review.
I dropped the `scores`/`groups` type constraints, as you suggested, and tried to avoid vector allocations, where possible.
Considering the other code tweaks, this PR should reduce the allocations when constructing `VarianceEqualityTest` objects.

alyst:
bump :)

test/var_equality.jl
```diff
@@ -1,37 +1,69 @@
...
 
-    show(IOContext(IOBuffer(), :table => true), t)
-    show(IOBuffer(), t)
+        # test splatting version
+        t2 = OneWayANOVATest(groups...)
+        @test nobs(t2) == nobs(t)
+        @test dof(t2) == dof(t)
+        @test pvalue(t2) == pvalue(t)
+        @test HypothesisTests.teststatistic(t2) == HypothesisTests.teststatistic(t)
+        @test sprint(show, t2) == sprint(show, t)
```
devmotion, line 32: This is just
```suggestion
        @test repr(t2) == repr(t)
```

test/var_equality.jl
```diff
@@ -1,37 +1,69 @@
...
-    @test dof(t) == (3,25)
-    @test pvalue(t) ≈ 0.072 atol=1e-3
-    @test occursin("reject h_0", sprint(show, t))
+        # test tuple version
+        t3 = OneWayANOVATest((groups...))
+        @test nobs(t3) == nobs(t)
+        @test dof(t3) == dof(t)
+        @test pvalue(t3) == pvalue(t)
+        @test HypothesisTests.teststatistic(t3) == HypothesisTests.teststatistic(t)
+        @test sprint(show, t3) == sprint(show, t)
```
devmotion, line 40:
```suggestion
        @test repr(t3) == repr(t)
```

test/var_equality.jl
```diff
@@ -1,37 +1,69 @@
...
+
+        # test generator version
+        t4 = OneWayANOVATest(g for g in groups)
+        @test nobs(t4) == nobs(t)
+        @test dof(t4) == dof(t)
+        @test pvalue(t4) == pvalue(t)
+        @test HypothesisTests.teststatistic(t4) == HypothesisTests.teststatistic(t)
+        @test sprint(show, t4) == sprint(show, t)
```
devmotion, line 48:
```suggestion
        @test repr(t4) == repr(t)
```

test/var_equality.jl
```diff
@@ -40,38 +72,63 @@ using DelimitedFiles
...
+        @test nobs(l) == fill(8, 4)
+        @test dof(l) == (3,28)
+        @test pvalue(l) ≈ 0.90357 atol=1e-4
+
+        l2 = LeveneTest(groups...; statistic=mean)
+        @test nobs(l2) == nobs(l)
+        @test dof(l2) == dof(l)
+        @test pvalue(l2) == pvalue(l)
+        @test HypothesisTests.teststatistic(l2) == HypothesisTests.teststatistic(l)
+        @test sprint(show, l2) == sprint(show, l)
```
devmotion, line 87:
```suggestion
        @test repr(l2) == repr(l)
```

test/var_equality.jl
```diff
@@ -40,38 +72,63 @@ using DelimitedFiles
...
-    @test pvalue(t) ≈ 0.9878 atol=1e-4
-    @test HypothesisTests.teststatistic(t) ≈ 0.1311 atol=1e-5
-    @test occursin("reject h_0", sprint(show, t))
+        # test splatting version
+        t2 = FlignerKilleenTest(groups...)
+        @test nobs(t2) == nobs(t)
+        @test dof(t2) == dof(t)
+        @test pvalue(t2) == pvalue(t)
+        @test HypothesisTests.teststatistic(t2) == HypothesisTests.teststatistic(t)
+        @test sprint(show, t2) == sprint(show, t)
```
devmotion, line 111:
```suggestion
        @test repr(t2) == repr(t)
```

test/var_equality.jl
```diff
@@ -40,38 +72,63 @@ using DelimitedFiles
...
-    @test dof(l) == (9,90)
-    @test HypothesisTests.teststatistic(l) ≈ 1.705910 atol=1e-5
-    @test pvalue(l) ≈ 0.0991 atol=1e-4
-    @test occursin("reject h_0", sprint(show, l))
+        l2 = BrownForsytheTest(eachcol(samples)...)
+        @test nobs(l2) == nobs(l)
+        @test dof(l2) == dof(l)
+        @test HypothesisTests.teststatistic(l2) == HypothesisTests.teststatistic(l)
+        @test pvalue(l2) == pvalue(l)
+        @test sprint(show, l2) == sprint(show, l)
```
devmotion, line 132:
```suggestion
        @test repr(l2) == repr(l)
```

alyst:
@devmotion Thank you, I've applied your suggestions

devmotion (approved):

PR #334, opened 2025-04-01
Replace Rmath with StatsFuns

devmotion:
Closes #331. Closes #330.

PR #335, opened 2025-04-02
Add tests of left-tail + right-tail p-values for exact Mann-Whitney U and signed rank tests

devmotion:
This PR increases test coverage: Currently, left-tail and right-tail p-values of these two hypothesis tests are not tested. This came up in #334. I moved the coverage changes to a separate PR to ensure that #334 does not introduce any regressions.

~~Since I was unable to reproduce even the existing two-sided p-values with R, the reference values added in this PR are based on HypothesisTests#master. Clearly they should be validated with some other software before merging this PR. As an example, the existing tests check~~
```julia
    @test abs(@inferred(pvalue(ExactMannWhitneyUTest([1:10;], [2.1:2:21;]))) - 0.0232) <= 1e-4
```
~~but with R I get completely different p-values:~~
```R
> wilcox.test(c(1:10), c(2.1:2:21), alternative="two.sided", exact=TRUE, correct=FALSE)


        Wilcoxon rank sum exact test

data:  c(1:10) and c(2.1:2:21)
W = 36, p-value = 0.00568
alternative hypothesis: true location shift is not equal to 0

> wilcox.test(c(1:10), c(2.1:2:21), alternative="two.sided", exact=TRUE, correct=TRUE)


        Wilcoxon rank sum exact test

data:  c(1:10) and c(2.1:2:21)
W = 36, p-value = 0.00568
alternative hypothesis: true location shift is not equal to 0
```
I assume I'm doing something wrong in R?! @ArnoStrouwen do you know how to validate the p-values with R?

Edit: Of course, it was a stupid mistake on my side.

```R
> c(2.1:2:21)

 [1]  2.1  3.1  4.1  5.1  6.1  7.1  8.1  9.1 10.1 11.1 12.1 13.1 14.1 15.1 16.1
[16] 17.1 18.1 19.1 20.1
> seq(2.1, 21, by=2)

 [1]  2.1  4.1  6.1  8.1 10.1 12.1 14.1 16.1 18.1 20.1

> wilcox.test(1:10, seq(2.1, 21, by=2), alternative="two.sided", exact=TRUE, correct=FALSE)


        Wilcoxon rank sum exact test

data:  1:10 and seq(2.1, 21, by = 2)
W = 20, p-value = 0.02323
alternative hypothesis: true location shift is not equal to 0
```

### DIFF
--- a/test/mann_whitney.jl
+++ b/test/mann_whitney.jl
@@ -44,7 +44,7 @@ end
     @test abs(@inferred(pvalue(ExactMannWhitneyUTest([1:10;], [1:10;]); tail = :left)) - 0.5296) <= 1e-4
     @test abs(@inferred(pvalue(ExactMannWhitneyUTest([1:10;], [2:11;]); tail = :left)) - 0.2548) <= 1e-4
     @test abs(@inferred(pvalue(ExactMannWhitneyUTest([2:11;], [1:10;]); tail = :left)) - 0.7634) <= 1e-4
-    @test abs(@inferred(pvalue(ExactMannWhitneyUTest([1:10;], [1:5; ones(5)]); tail = :left)) - 0.9978) <= 1e-4
+    @test abs(@inferred(pvalue(ExactMannWhitneyUTest([1:10;], [1:5; ones(5)]); tail = :left)) - 0.9979) <= 1e-4
     @test abs(@inferred(pvalue(ExactMannWhitneyUTest([1:5; ones(5)], [1:10;]); tail = :left)) - 0.0028) <= 1e-4
 
     # Right tail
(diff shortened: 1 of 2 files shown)


test/mann_whitney.jl
devmotion, line 1: The p-values of `ExactMannWhitneyUTest` are checked against R using `coin::wilcox_test` with `distribution="exact"` and `alternative="two.sided"` (two-sided), `alternative="less"` (left tail) and `alternative="greater"` (right tail). This uncovered bugs of one-sided p-values in tests with ties and samples of unequal length.

test/wilcoxon.jl
devmotion, line 1: The p-values of `ExactSignedRankTest` are checked against R using `coin::wilcoxsign_test` with `distribution="exact"` and `alternative="two.sided"` (two-sided), `alternative="less"` (left tail) and `alternative="greater"` (right tail).

test/mann_whitney.jl
ArnoStrouwen: Not very familiar with `coin`, but the base R tests has a paired keyword. Do you not need to specify this here?

devmotion: `wilcox.test` does not support exact test with ties (or at least it errored in the cases I tried).

devmotion: For the cases that are supported it gave the same p-values as `coin::wilcox_test`.

devmotion: But the bug is clear even without knowing the correct p-value: #337

ArnoStrouwen (approved):
I've found the same values with the coin package.
I also checked the tests without ties using base R and found the same.

andreasnoack (approved):

PR #338, opened 2025-04-03
Fix one-sided p-values of exact Mann-Whitney-U test with ties

### LINKED ISSUE #337: One-sided p-values of `ExactMannWhitneyUTest` with ties + unequal lengths of samples broken
I noticed this issue when adding tests for these one-sided p-values in #335 (currently they are not tested). In R, I get

```R
> data <- data.frame(x = factor(c(rep("a", 10), rep("b", 12))), y = c(1:10, seq(2, 24, by=2)))

> coin::pvalue(coin::wilcox_test(y ~ x, data, distribution="exact", alternative="two.sided"))

[1] 0.01188904
> coin::pvalue(coin::wilcox_test(y ~ x, data, distribution="exact", alternative="less"))

[1] 0.006001738
> coin::pvalue(coin::wilcox_test(y ~ x, data, distribution="exact", alternative="greater"))

[1] 0.9949199
```

and (with groups swapped)

```R
> data <- data.frame(x = factor(c(rep("a", 12), rep("b", 10))), y = c(seq(2, 24, by=2), 1:10))

> coin::pvalue(coin::wilcox_test(y ~ x, data, distribution="exact", alternative="two.sided"))

[1] 0.01188904
> coin::pvalue(coin::wilcox_test(y ~ x, data, distribution="exact", alternative="less"))

[1] 0.9949199
> coin::pvalue(coin::wilcox_test(y ~ x, data, distribution="exact", alternative="greater"))

[1] 0.006001738
```

Note in particular that the symmetry is as expected.

However, with HypothesisTests I get

```julia
julia> test = ExactMannWhitneyUTest(1:10, 2:2:24);

julia> pvalue(test; tail=:both)
0.011774603105872456

julia> pvalue(test; tail=:left)
0.9948055040934298

julia> pvalue(test; tail=:right)
0.005887301552936228

julia> test = ExactMannWhitneyUTest(2:2:24, 1:10);

julia> pvalue(test; tail=:both)
0.011774603105872456

julia> pvalue(test; tail=:left)
0.9948055040934298

julia> pvalue(test; tail=:right)
0.005887301552936228
```

That is,
1. one-sided p-values are wrong in the case `length(x) < length(y)` (symmetry property is clearly not satisfied),
2. possibly the "right-tail" p-values are slightly off (0.005887... in HypothesisTests vs 0.006001738... in R).

devmotion:
Fixes #337.

### DIFF
--- a/src/mann_whitney.jl
+++ b/src/mann_whitney.jl
@@ -57,17 +57,19 @@ function MannWhitneyUTest(x::AbstractVector{S}, y::AbstractVector{T}) where {S<:
 end
 
 # Get U, ranks, and tie adjustment for Mann-Whitney U test
-function mwustats(x::AbstractVector{S}, y::AbstractVector{T}) where {S<:Real,T<:Real}
+# U is the sum of adjusted ranks in the first sample minus the minimal sum of ranks (ie sum(1:length(x))
+function mwustats(x::AbstractVector{<:Real}, y::AbstractVector{<:Real})
+    ranks, tieadj = tiedrank_adj([x; y])
     nx = length(x)
     ny = length(y)
     if nx <= ny
-        (ranks, tieadj) = tiedrank_adj([x; y])
-        U = sum(@view ranks[1:nx]) - nx*(nx+1)/2
+        U = sum(@view(ranks[begin:(begin + (nx-1))])) - nx*(nx+1)/2
     else
-        (ranks, tieadj) = tiedrank_adj([y; x])
-        U = nx*ny - sum(@view ranks[1:ny]) + ny*(ny+1)/2
+        # Sum of adjusted ranks of first and second sample sums to (nx + ny)*(nx + ny + 1)/2, hence
+        # U = (nx + ny)*(nx + ny + 1)/2 - sum(ranks_y) - nx*(nx + 1)/2
+        U = ny*(2 * nx + ny + 1)/2 - sum(@view(ranks[(begin + nx):end]))
     end
-    (U, ranks, tieadj, nx, ny, median(x)-median(y))
+    return (U, ranks, tieadj, nx, ny, median(x)-median(y))
 end
 
 
@@ -110,7 +112,7 @@ function show_params(io::IO, x::ExactMannWhitneyUTest, ident)
     println(io)
     println(io, ident, "Mann-Whitney-U statistic:             ", x.U)
     print(io, ident, "rank sums:                            ")
-    show(io, [sum(@view x.ranks[1:x.nx]), sum(@view x.ranks[x.nx+1:end])])
+    show(io, [sum(@view(x.ranks[begin:(begin + (x.nx - 1))])), sum(@view(x.ranks[(begin + x.nx):end]))])
     println(io)
     println(io, ident, "adjustment for ties:                  ", x.tie_adjustment)
 end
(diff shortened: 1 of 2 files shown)


test/mann_whitney.jl
```diff
@@ -3,27 +3,144 @@ using HypothesisTests: default_tail
...
+        number of observations in each group: [12, 10]
+        Mann-Whitney-U statistic:             22.5
+        rank sums:                            [175.5, 77.5]
+        adjustment for ties:                  30.0
+    """
+
+    # Two-sided
+    for kwargs in ((), (; tail = :both))
+        @test abs(@inferred(pvalue(ExactMannWhitneyUTest([1:10;], [2:2:24;]); kwargs...)) - 0.0120) <= 1e-4
+        @test abs(@inferred(pvalue(ExactMannWhitneyUTest([2:2:24;], [1:10;]); kwargs...)) - 0.0120) <= 1e-4
```
devmotion, line 134: I'm not sure why but there's a slight discrepancy here, the values reported by R (#337) are `0.01188904`.

ArnoStrouwen:
This also contains changes for the approximate test.
If that one is also broken, it might be a good idea to add a test that fails on the old version and passes on the new?

src/mann_whitney.jl
```diff
@@ -110,26 +112,39 @@ function show_params(io::IO, x::ExactMannWhitneyUTest, ident)
...
     print(io, ident, "rank sums:                            ")
-    show(io, [sum(@view x.ranks[1:x.nx]), sum(@view x.ranks[x.nx+1:end])])
+    show(io, [sum(@view(x.ranks[begin:(begin + (x.nx - 1))])), sum(@view(x.ranks[(begin + x.nx):end]))])
     println(io)
     println(io, ident, "adjustment for ties:                  ", x.tie_adjustment)
 end
 
 # Enumerate all possible Mann-Whitney U results for a given vector,
 # determining left-and right-tailed p values
 function mwuenumerate(x::ExactMannWhitneyUTest)
```
ArnoStrouwen, line 122: I haven't gone through the math of this hypothesis test when you can't assume that ties occur with probability zero (which you can when the underlying distributions are continuous).

This enumeration seems to assume that under the H0 each iteration in `combinations(x.ranks, n)` is equally likely to occur? I'm not convinced that is true, but I'll need to do a simulation study to confirm.

It will take me a few days to get familiar enough to review this thoroughly.

devmotion:
I haven't checked the approximate test thoroughly but the PR does not contain any bugfixes for it.

src/mann_whitney.jl
devmotion: There's actually no special assumptions involved here and no conceptual change. The main problem with master is just that it messed up calculations for the case nx < ny and arguably the code was a bit confusing and undocumented. I tried to make it a bit clearer what was going on.

This function is the standard way of deriving the distribution of U (and thereby its tail probabilities) for the given observations under the null hypothesis by counting and comparing all possible values.

ArnoStrouwen: I agree that this is the standard way, and that it makes sense to match the results from R in this PR.
I just have an inkling the standard way is wrong, which I'll investigate further.
My suspicion is that there is no nice distribution for U under the null when the probability of ties is nonzero.

Anyway, that doesn't necessarily need to hold up this PR. I can review it for matching the behavior of R.

devmotion: It will probably take a few days until I can return to this (I have some important deadlines to meet in the next few days and coming weeks), but depending on how large the discrepancies were it could be interesting to do a similar check with a) your own p-value computations or (probably easier and more relevant) b) the p-value calculations in this PR here. As I mentioned above, I saw a few minor discrepancies with the p-values reported by `wilcox_test`, so it might be good to compare with more than just `coin` (IIRC the docs also mentioned that p-values are typically adjusted, so maybe there's also a few more options to consider).

ArnoStrouwen: The discrepancies are rather large.

ArnoStrouwen: Ok, I figured out why the simulation study gives different results:
```
julia> x = [-3,-2,-2,-2,-1,0,1,1,2,3];
julia> esr = ExactSignedRankTest(x);
julia> esr.ranks
9-element Vector{Float64}:
 8.5
 5.5
 5.5
 5.5
 2.0
 2.0
 2.0
 5.5
 8.5
julia> esr.W
18.0
```
The full enumeration calculates which subsets of `esr.ranks` are more extreme than `esr.W`, relative to the total number of subsets.
This correctly assumes that each of the `2^n` variations of `x` with the signs of the `n` elements possibly switched are equally likely to occur under the null hypothesis. However, it also conditions on this particular tie-pattern occurring. If you were to repeat the experiment, you could have very well ended up with more -3/3 in the sample, which is not accounted for in the enumeration, but is present in my simulation study.

I'm not sure if conditioning on the observed tie-pattern leads to valid p-values. I'll have to think more about this. For continuous data, you do know the tie-pattern with probability one, namely that there are no ties, so this issue does not occur.

Also, I'm not a fan of dropping the zero observations, because you can't figure out to assign them a + or - sign, if you have a dataset with 99 zeros and a one, then there is strong evidence that the median is zero, but you don't notice this if you just drop them.

andreasnoack:
What remains to be done here?

devmotion:
It has to be approved 🙂 @ArnoStrouwen expressed concerns regarding the common way of computing p-values in the case of ties but I anticipate that I won't be able to look into that in more detail before the end of the week. The PR fixes the obvious bug in #337 (the broken symmetry property is wrong regardless of what p-values should be returned) by fixing an incorrect branch on master (not changing the general idea/algorithm of how to compute the p-value).

test/mann_whitney.jl
ArnoStrouwen: Can you make an issue about this so it doesn't get lost?
Can't find an immediate source of the discrepancy.

ArnoStrouwen (approved):
I approved, since my concern is a meta-concern about the appropriateness of the test, and not with the computation itself.

andreasnoack (approved):

PR #347, opened 2025-08-17
Create docs site favicon

abhro:


abhro:
Admittedly, this is a very minor improvement. It mostly helps bookmarks and open tabs in web browsers and immediately creates the association that it's one of the JuliaStats pages when reading documentation. I've also found that in other organizations, the org logo is used as the default fallback unless the repository itself has a different logo. As for the shared access to the organization logo for all repos, there probably is a way to facilitate that, but I'm not skillful enough to have found/developed it 😅

devmotion (approved):

PR #348, opened 2025-08-17
Use `[sources]` in docs/Project.toml to manage self dependency

abhro:
Documentation for Pkg.jl feature is available at
https://pkgdocs.julialang.org/v1/toml-files/#The-[sources]-section

This also simplifies local development of the docs project since one can just run `]instantiate` in the docs environment without having to do a `dev .` in the root directory.

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -1,5 +1,9 @@
 [deps]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+HypothesisTests = "09f84164-cd44-5f33-b23f-e6b0d136a0d5"
 
 [compat]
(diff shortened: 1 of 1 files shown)


docs/Project.toml
```diff
@@ -1,5 +1,9 @@
 [deps]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+HypothesisTests = "09f84164-cd44-5f33-b23f-e6b0d136a0d5"
 
 [compat]
 Documenter = "1"
```
devmotion, line 6: Can you add a compat entry to avoid an unnecessary CompatHelper PR?

Issue #364, opened 2026-07-28
Allow choosing between the exact and approximate methods for the Wilcoxon rank tests


yoninazarathy:
`SignedRankTest` and `MannWhitneyUTest` decide between the exact and approximate methods internally, from sample size and ties, with no way for a caller to override it.

```julia
SignedRankTest(x; method = :auto)      # :auto | :exact | :approximate | callable
MannWhitneyUTest(x, y; method = :auto)
```

`:auto` reproduces today's dispatch, so nothing changes for existing callers.

## Why

The choice currently depends on HypothesisTests' internal thresholds, so a version bump can change which branch a dataset takes and therefore the reported numbers. A downstream package that needs a reproducible analysis cannot pin it.

The reverse direction matters too: forcing `:approximate` avoids the exact branch where it is expensive (#7's exhaustive enumeration under ties) or numerically unusable at large `n` (JuliaStats/StatsFuns.jl#219, #220).

## Callable form

`method::Union{Symbol,Function}` also covers policies that are not one of the two constants. When callable, it would be handed the decision inputs and return `:exact` or `:approximate`:

```julia
method = s -> (s.n_nonzero ≤ 50 && !s.ties) ? :exact : :approximate

# s :: (; n, n_nonzero, ties, tie_adjustment)   [signed-rank]
# s :: (; nx, ny, ties, tie_adjustment)         [rank-sum]
```

All available where the branch is taken: `signedrankstats` returns `(W, ranks, signs, tie_adjustment, n, median)` and `mwustats` returns `(U, ranks, tie_adjustment, nx, ny, median)`, so `n_nonzero` is `length(ranks)` and `ties` is `tie_adjustment != 0`.

## Prior art

`method` is already the house keyword for selecting among computations — `confint(::BinomialTest; method = :clopper_pearson)` offers seven, and `FisherExactTest` takes it on both `pvalue` and `confint`. #211 requests the same pattern for `BinomialTest`'s p-value.

## Breaking

No — purely additive, `:auto` by default. Forcing `:exact` where it does not apply should error rather than silently downgrade.

Related: #211 · #361 (the test type currently does not affect the interval at all, which this would make meaningful) · #7 · JuliaStats/StatsFuns.jl#219, #220

Julia 1.12.6 · HypothesisTests 0.11.8

nalimilan:
Would you make a pull request to add a `method` argument?

Issue #361, opened 2026-07-28
`calculate_ci` uses the exact null distribution for `Approximate*` tests, and selects the wrong order statistic from n = 73


yoninazarathy:
`calculate_ci` computes the interval from the exact null distribution regardless of which test was selected, regardless of ties, and regardless of whether the underlying routine is usable at that sample size.

On HypothesisTests 0.11.8 with StatsFuns 2.2.0.

## 1. The approximate test does not give an approximate interval

`StatsAPI.confint` is defined identically on both signed-rank types — [L166](https://github.com/JuliaStats/HypothesisTests.jl/blob/d2e20b5c69559f665bbc7fa10a23fc643a9540f9/src/wilcoxon.jl#L166), [L241](https://github.com/JuliaStats/HypothesisTests.jl/blob/d2e20b5c69559f665bbc7fa10a23fc643a9540f9/src/wilcoxon.jl#L241) — and both call `calculate_ci(x.vals, level)`, which calls `signrankcdf` unconditionally.

```julia
julia> using HypothesisTests

julia> d = [0.0,0,0,0.5,0.5,1,-0.5,-1,1.5,-1.5,0.5,0,1,-0.5,2,0,0.5,-1,1,0.5];

julia> confint(HypothesisTests.ExactSignedRankTest(d); level=0.9)
(0.0, 0.5)

julia> confint(HypothesisTests.ApproximateSignedRankTest(d); level=0.9)
(0.0, 0.5)
```

So there is no normal-approximation interval in the package. Two consequences: `tie_adjustment` never reaches the interval, though ties are guaranteed for discrete measurements; and [`k = argmin(abs.(l .- c))`](https://github.com/JuliaStats/HypothesisTests.jl/blob/d2e20b5c69559f665bbc7fa10a23fc643a9540f9/src/wilcoxon.jl#L257) picks the *nearest* attainable level, which may be below nominal, where `SignTest` inverts through `quantile(Binomial(n, 0.5), (1-level)/2)` and is conservative.

## 2. The selected order statistic is wrong from n = 73

[...]

Issue #365, opened 2026-07-29
SignedRankTest rejects Float32, Float16, BigFloat and Rational input


yoninazarathy:
## Summary

`SignedRankTest` fails to construct for any element type other than `Float64` and the integers. `MannWhitneyUTest` and `SignTest` accept all of them.

```julia
julia> using HypothesisTests

julia> x = Float32[3, -1, 4, -1, 5, 9, -2, 6, 5, 3];

julia> SignedRankTest(x)
ERROR: MethodError: no method matching ExactSignedRankTest(::Vector{Float32}, ::Float64, ::Vector{Float64}, ::BitVector, ::Float64, ::Int64, ::Float32)
```

| eltype | `SignedRankTest` | `MannWhitneyUTest` | `SignTest` |
|---|---|---|---|
| `Float64` | ok | ok | ok |
| `Int`, `Int32` | ok | ok | ok |
| `Float32` | **MethodError** | ok | ok |
| `Float16` | **MethodError** | ok | ok |
| `BigFloat` | **MethodError** | ok | ok |
| `Rational{Int}` | **MethodError** | ok | ok |

`ExactSignedRankTest` and `ApproximateSignedRankTest` fail the same way when constructed directly, as does the two-sample form `SignedRankTest(x, y)`.

## Cause

Both signed rank structs declare the median field as a concrete `Float64` — [wilcoxon.jl#L79](https://github.com/JuliaStats/HypothesisTests.jl/blob/d2e20b5c69559f665bbc7fa10a23fc643a9540f9/src/wilcoxon.jl#L79) and [#L178](https://github.com/JuliaStats/HypothesisTests.jl/blob/d2e20b5c69559f665bbc7fa10a23fc643a9540f9/src/wilcoxon.jl#L178) — while [`signedrankstats`](https://github.com/JuliaStats/HypothesisTests.jl/blob/d2e20b5c69559f665bbc7fa10a23fc643a9540f9/src/wilcoxon.jl#L67) returns `median(x)`, which is `Float32` for `Float32` input. The struct is parametric, so the generated outer constructor requires the declared field types exactly and the call does not match.

Integer input works by coincidence: `median` of an integer vector is already `Float64`.

## Precedent

[...]

Issue #363, opened 2026-07-28
Wilcoxon rank tests report a point estimate that is not the one their confidence interval is built around


yoninazarathy:
The four Wilcoxon rank test types report a point estimate labelled `"Location parameter (pseudomedian)"` that is not a pseudomedian, and not the quantity their interval is built around. It is the sample median, or the difference of sample medians.

```julia
julia> using HypothesisTests, Statistics

julia> x = [1.0, 2, 3, 4, 5, 6, 7, 8, 20, 30];

julia> HypothesisTests.population_param_of_interest(SignedRankTest(x))[3]
5.5                       # == median(x)

julia> median([(x[i]+x[j])/2 for i in eachindex(x) for j in i:length(x)])
6.0                       # the pseudomedian, which confint is built from

julia> a = [1.0,2,3,4,5]; b = [10.0,11,12,40,50];

julia> HypothesisTests.population_param_of_interest(MannWhitneyUTest(a,b))[3]
-9.0                      # == median(a) - median(b)

julia> median([ai - bj for ai in a, bj in b])
-10.0                     # the two-sample Hodges–Lehmann estimate
```

R returns the pseudomedian on the same two samples:

```r
> wilcox.test(c(1,2,3,4,5,6,7,8,20,30), conf.int = TRUE)$estimate
(pseudo)median
             6
> wilcox.test(c(1,2,3,4,5), c(10,11,12,40,50), conf.int = TRUE)$estimate
difference in location
                   -10
```

## Context

Every other location test reports the estimate its interval is built on — `BinomialTest` (`x.x/x.n`), `SignTest` (`median(x.data)`), both t-tests and both z-tests (`x.xbar`). The four rank tests return `x.median` under a label promising something else.

`SignTest` is the closest comparison: also distribution-free, also inverting a discrete test to order-statistic endpoints ([binomial.jl#L235-L250](https://github.com/JuliaStats/HypothesisTests.jl/blob/d2e20b5c69559f665bbc7fa10a23fc643a9540f9/src/binomial.jl#L235-L250)), and it reports the quantity its interval brackets.

[...]

PR #366, opened 2026-07-29
Improve and fix the nonparametric rank tests

yoninazarathy:
> **Draft.** The code is finished and green, but four of the changes below alter numbers that users currently see, and one is a judgement call between two defensible conventions (#362). I would rather have a maintainer's steer on those before this is treated as ready to merge. Individual commits are self-contained, so pieces can be taken or dropped.

Addresses #361, #362, #363, #364, #365, and the Mann–Whitney half of #8.

## Summary

Confidence intervals from all four Wilcoxon rank test types — exact and approximate, signed rank and Mann–Whitney — now match R's `wilcox.test`. So does the point estimate, which becomes the Hodges–Lehmann estimator that R reports on its exact route.

This changes numbers users currently see. Measured against released 0.11.8 over 66 samples, the interval moved on 43 of them and the point estimate on all 66. The **What actually breaks** section below gives the breakdown and the migration path.

## What was wrong

- **`MannWhitneyUTest` had no `confint`** — the only location test in the package without one.
- **For the signed rank tests, `calculate_ci` ignored which of them it was called on.** It is the only interval in this family, shared by `ExactSignedRankTest` and `ApproximateSignedRankTest`, and it inverted the exact null distribution either way — whether or not there were ties, and whether or not `signrankcdf` was numerically usable at that sample size.
- **The exact/approximate choice could not be pinned.** It followed internal thresholds, so a version bump could silently move a given dataset from one branch to the other, and with it the reported numbers.
- **The reported point estimate was not the one the interval was built around.** The label said pseudomedian; the value was the sample median, or the difference of sample medians.

## What changes

| issue | change | breaking |
|---|---|---|
| #364 | `method = :auto \| :exact \| :approximate \| callable` on both tests | no |
| #8 | Mann–Whitney structs keep their samples; `confint` and `hodgeslehmann` | structs only |
| #361 | the interval now depends on the test type | **yes** |
| #361 | level selection conservative rather than nearest | **yes** |
| #362 | zero differences dropped for the interval, as they already are for the statistic | **yes** |
| #363 | the reported point estimate is the Hodges–Lehmann estimate | **yes** |
| #365 | signed rank accepts element types other than `Float64` | structs only |

*structs only*: all four structs gain a second type parameter — the Mann–Whitney pair also gain `x` and `y` fields. The fields are appended, so the existing ones keep their positions, but the six-argument positional constructor is gone and `ExactMannWhitneyUTest{Float64}` is no longer a concrete type. Nothing in the package or its test suite constructed these positionally.

### The interval now depends on the test

`StatsAPI.confint` was defined identically on both signed-rank types and called `signrankcdf` either way, so selecting `ApproximateSignedRankTest` did not produce an approximate interval — the package had no normal-approximation interval at all. `Exact*` now inverts the exact null distribution; `Approximate*` inverts the normal approximation, using the tie-corrected standard deviation it already stores. Ties therefore affect the interval, which they previously did not.

### Conservative level selection

`argmin(abs.(l .- c))` could settle on an attainable level *below* nominal. Taking instead the largest index that still attains the level reproduces R's `qsignrank`-based rule on **all 67** sample sizes `n = 5:71`, at every level tried — 0.80, 0.90, 0.95 and 0.99. The old rule agreed with R on only 30, 35, 32 and 27 of those 67 respectively.

### Zeros

`signedrankstats` drops zero differences before ranking, so the statistic and p-value describe the non-zero differences — but the interval was formed from the Walsh averages of *all* of them (210 against 120, on a 20-point sample with five zeros). The p-value and the interval therefore described different samples. Both now describe the non-zero differences, as R does.

### The point estimate

The label said "pseudomedian"; the value was `median(x)`, or `median(x) - median(y)`. The pseudomedian — the Hodges–Lehmann estimator, the median of the contrast set the interval is read off — is a different quantity, and is now what is reported. It is also exposed directly as `hodgeslehmann`.

## Evidence: agreement with R

Released 0.11.8, this branch, and R's `wilcox.test`, on the same samples.

Some rows below match R digit for digit and others only to four or five places. That is R switching construction, not a disagreement:

- On its **exact** route R reads two order statistics off the sorted contrast set, exactly as this branch does, so the numbers are identical — the `(3.3, 15.5)`, `(-11.1, -0.1)` and `(-0.95525, 0.30358)` rows.
- On its **approximate** route R does not use order statistics at all. It solves numerically, with `uniroot`, for the shift at which the statistic crosses its critical value. Its endpoints therefore land near an order statistic without being one: `3.0500354` where this branch returns `3.05`.

### Confidence intervals

| case | 0.11.8 | this branch | R |
|---|---|---|---|
| signed rank, n=15, no ties — exact | `(3.3, 15.5)` | `(3.3, 15.5)` | `(3.3, 15.5)` |
| signed rank, n=15, no ties — approximate | `(3.3, 15.5)` ¹ | `(3.05, 15.5)` | `(3.05004, 15.50001)` |
| signed rank, n=15, ties | `(-4.6, -0.1)` | `(-4.6, -0.1)` | `(-4.60008, -0.09997)` |
| signed rank, n=20, five zeros, level 0.9 | `(0.0, 0.5)` | `(-0.25, 0.75)` | `(-0.24999, 0.75005)` |
| signed rank, n=60, no ties — approximate | `(0.3155, 0.847)` | `(0.315, 0.8495)` | `(0.31492, 0.84954)` |
| Mann–Whitney, 10 v 10, no ties — exact | *none* ² | `(-11.1, -0.1)` | `(-11.1, -0.1)` |
| Mann–Whitney, 10 v 12, ties | *none* ² | `(-14.0, -1.0)` | `(-13.99997, -1.00003)` |
| Mann–Whitney, 20 v 25, no ties — exact | *none* ² | `(-0.95525, 0.30358)` | `(-0.95525, 0.30358)` |
| Mann–Whitney, 9 v 9, ties — approximate | *none* ² | `(-0.37, 1.183)` | `(-0.36998, 1.18297)` |
| Mann–Whitney, 40 v 45, no ties — approximate | *none* ² | `(-1.10474, -0.12834)` | `(-1.10473, -0.12837)` |

¹ The approximate test returned the exact interval.
² `MannWhitneyUTest` had no `confint` method.

### Point estimates

| case | 0.11.8 | this branch | R |
|---|---|---|---|
| signed rank, n=15, no ties | `10.1` | `9.675` | `9.675` |
| signed rank, n=15, ties | `-0.1` | `-2.1` | `-2.1` |
| signed rank, n=20, five zeros | `0.25` | `0.5` | `0.49999849` |
| signed rank, n=60, no ties | `0.5595` | `0.56775` | `0.56795` ³ |
| Mann–Whitney, 10 v 10, no ties | `-5.6` | `-5.6` ⁴ | `-5.6` |
| Mann–Whitney, 9 v 9, ties | `0.62` | `0.56` | `0.56005616` |

³ R's default at this size is its approximate route, which root-finds the estimate instead of taking the median of the contrast set. R's exact route gives `0.56775`, matching this branch exactly.
⁴ Unchanged here because this sample is symmetric enough that the sample median and the Hodges–Lehmann estimate coincide. That is why the existing `show` tests did not catch the change — it is not the general case, as the Mann–Whitney row below it shows.

**Comparing like with like.** R and this package dispatch differently. R takes the exact route when each sample is under 50 *and* there are no ties; this package when the total is at most 50 (Mann–Whitney) or the non-zero count is at most 50 (signed rank), and additionally at 15 or fewer non-zero values even with ties. The interval rows above therefore set R's `exact` argument explicitly, to compare against the same branch the test took, rather than relying on either default.

**Two caveats.**

Wherever R uses its approximate route it root-finds the point estimate rather than taking the median of the contrast set, so its `estimate` drifts slightly from the Hodges–Lehmann value — `9.71184` against `9.675` for the n=15 sample at `exact = FALSE`. This branch always reports the estimator itself, which is why it agrees exactly with R wherever R took its exact route.

Exact Mann–Whitney intervals are unavailable past `MAX_EXACT_WILCOX_BINOMIAL`. At 40 v 45, `binomial(85, 40)` is about `2.9e24`, so `wilcoxcdf` cannot be used at all and the branch refuses. R has no such limit, because Rmath's `pwilcox` does not accumulate counts in `Int`. This package's automatic rule selects the approximate test well below that size, so only a forced `method = :exact` reaches it. This is JuliaStats/StatsFuns.jl#220 reaching user-visible behaviour.

## What actually breaks

**Measured.** 66 deterministic signed-rank samples — `n` from 8 to 60, a third of them rounded to force ties, a third carrying zero differences — run under released 0.11.8 and under this branch, comparing `confint` and the reported point estimate:

| data | test | cases | interval moved | estimate moved |
|---|---|---|---|---|
| no ties, no zeros | exact | 18 | 6 | 18 |
| no ties, no zeros | approximate | 4 | 4 | 4 |
| ties | exact | 17 | 6 | 17 |
| ties | approximate | 5 | 5 | 5 |
| zeros present | exact | 22 | 22 | 22 |
| **total** | | **66** | **43** | **66** |

The point estimate moves on every sample: the Hodges–Lehmann estimate coincides with the sample median only for exactly symmetric data. Intervals move on all approximate tests, on every sample containing a zero difference, and on about a third of exact tests through the conservative level selection.

That case mix is deliberately weighted toward the affected paths, so 43/66 measures how often it moves when it can move, not a rate over typical usage. Most users of these four tests will see different numbers.

**In the existing test suite, exactly one assertion changes value.** `test/wilcoxon.jl` pinned `confint` at `(3.3, 15.5)` for *both* the exact and the approximate test — which was the bug itself. The approximate lower bound is now `3.05`, against R's `3.05004`; the exact test still gives `3.3`.

That reflects the suite's coverage rather than the size of the change. There is one `confint` dataset in the whole suite (n=15, untied, no zeros), no signed-rank `show` test, and no Mann–Whitney `confint` to break. The three Mann–Whitney `show` tests that do change gain a confidence-interval line — new output rather than changed output — and their point estimates do not move, because their samples are symmetric enough that the two estimators agree.

**Migration.** `method = :exact` recovers the old index rule wherever it was valid, which is everywhere the automatic rule used to select the exact test. There is no opt-out for the zeros change or the point estimate; those are corrections rather than choices.

## Judgement calls

**Continuity correction on the approximate index.** For a lattice statistic `P(W ≤ j) ≈ Φ((j + ½ − μ)/σ)`, so the critical value is `ceil(μ − zσ − ½)`. Without the correction the index is anticonservative: the interval comes out narrower than the exact one on 44 of 66 signed-rank sizes at `level = 0.90`, and on 7 of 66 at `0.95`. With it, 10 of 66 and zero respectively. It is also what R applies by default.

**Ties on the exact branch.** Under ties the exact null distribution is no longer the untied one being inverted, so the achieved coverage is approximate. This branch keeps the classical construction, with a comment at both call sites recording the caveat. R instead declines and falls back to the normal approximation, which on every tied sample tested here lands on the same order statistics. Switching to R's behaviour is a small change if preferred.

**Guards, and where the constants come from.** The exact routes now refuse rather than return an invalid answer or run unbounded. The first two bounds are measured against `Rmath` over the whole support, not guessed; the third is a policy choice.

- `MAX_EXACT_SIGNRANK_N = 71` — `signrankcdf` is correct to 71 and wrong from 72, monotonically.
- `MAX_EXACT_WILCOX_BINOMIAL = 3_009_106_305_270_645_216` — `wilcoxcdf` is correct exactly while `binomial(nx + ny, nx)` is at or below this. Largest correct split is (30, 35); smallest wrong is (31, 34). The criterion was checked against 22 further splits, including highly unbalanced and large-`N` ones, with no mismatches.
- `MAX_EXACT_ENUMERATION_N = 25` — bounds the tied-data enumeration: `2^n` sign assignments for the signed rank test, `binomial(nx + ny, min(nx, ny))` rank assignments for Mann–Whitney. A judgement rather than a measurement; the automatic rule never enumerates beyond `2^15`.

None of the three is reachable through the automatic rule. They bite only on a forced `method = :exact`, or on direct construction of an `Exact*` type — the paths where the old code either returned a silently wrong answer or ran without bound. Upstream: JuliaStats/StatsFuns.jl#219 and JuliaStats/StatsFuns.jl#220. Neither has a fix in progress, so these are load-bearing indefinitely.

**A performance note.** `population_param_of_interest` now computes the Hodges–Lehmann estimate, which is `O(n²)` where the sample median was `O(n)`, so `show` is more expensive. `MAX_RANK_CONTRASTS` caps it at 10⁸ contrasts (`n ≈ 14_000`), past which it raises. At those sizes `confint` previously died with `signrank allocation error`, so this substitutes a clear error for a dead session without fixing the underlying cost; see #97. A selection-based estimator would remove the quadratic, which is #7.

If the point-estimate change is unwelcome, the label-only resolution offered in #363 is a two-line revert of `population_param_of_interest`, which also removes this cost.

## Structure

Seven commits, `+691 / −80` across seven files. New `src/rank_common.jl` (256 lines) holds what the two families share: method resolution, the guards, the two contrast sets, the two index rules. `src/wilcoxon.jl` and `src/mann_whitney.jl` mostly delegate to it.

## Tests

Wilcoxon 225 (was 62), Mann–Whitney 119 (was 77), rest of the suite unchanged. New coverage falls in two layers. Against R: the estimators, zero-difference consistency, and Mann–Whitney intervals on untied, tied and approximate samples. As properties, across sizes and both routes: the estimate lies inside its own interval, the interval cannot narrow as the level rises, the one-sided bounds are the two-sided endpoints with the far side open, `method` reaches the interval rather than only the returned type, the callable receives exactly the documented fields, element types are preserved rather than widened, and each guard raises at its exact boundary.

A pre-existing bug surfaced while testing and is fixed in the last commit: `SignedRankTest` rejected `Float32`, `Float16`, `BigFloat` and `Rational` input outright on released 0.11.8, because the structs declared `median::Float64` while `signedrankstats` returns `median(x)`. Filed as #365. That commit is independent of everything above and can be cherry-picked or dropped on its own.

Green across the full CI matrix — Julia `min-patch` (1.10), `1` and `pre`, on ubuntu, macOS and windows. All nine jobs report `Testing HypothesisTests tests passed`, with Wilcoxon 225/225 and Mann–Whitney 119/119 on every one, and the Documentation job builds clean.

Those runs were on a fork, where `codecov/codecov-action` cannot authenticate and the workflow's `fail_ci_if_error: true` reddens the job. It was the only failing step in any of the nine; `julia-runtest` succeeded in all of them. GitHub withholds secrets from fork-originated pull requests too, so the same step may well fail here — the Julia jobs themselves are the signal.


yoninazarathy:
JuliaStats/StatsFuns.jl#221 has since been merged and released as StatsFuns 2.2.1, fixing both JuliaStats/StatsFuns.jl#219 and JuliaStats/StatsFuns.jl#220 by folding the normaliser into the recursions so the counts are never formed. That changes the premise behind two of the guards here, so recording where it leaves this PR.

## Verified against 2.2.1

Worst absolute error over the **full** support, against `Rmath`:

| | StatsFuns 2.2.0 | StatsFuns 2.2.1 |
|---|---|---|
| `signrankcdf`, n = 72 | 0.23 | 6.7e-16 |
| `signrankcdf`, n = 200 | 0.377 | 2.9e-15 |
| `wilcoxcdf`, (31, 34) | 0.0105 | 6.3e-15 |
| `wilcoxcdf`, (33, 33) | 0.322 | 7.2e-15 |
| `wilcoxcdf`, (40, 40) | `OverflowError` | 2.2e-16 |
| `wilcoxcdf`, (50, 50) | `OverflowError` | 7.1e-15 |

(50, 50) is `binomial(100, 50)` ≈ 1.0e29, so the bound is gone rather than merely raised.

## What that means for this branch

`MAX_EXACT_SIGNRANK_N` and `MAX_EXACT_WILCOX_BINOMIAL` are correctness bounds for StatsFuns **≤ 2.2.0** and dead weight on 2.2.1. This package's `[compat]` is `StatsFuns = "1.4, 2"`, so both are still reachable and removing the guards outright would reintroduce silently invalid exact p-values and intervals for anyone resolving an older StatsFuns.

Three ways out, and the choice is a dependency-policy one rather than a statistical one, so I would rather have a maintainer pick than assume:

1. **Leave the guards as they are.** Correct on every supported StatsFuns, redundant on the newest. Costs nothing but two constants and a branch that never fires for current users.
2. **Raise `[compat]` to `StatsFuns = "2.2.1"` and delete them.** Cleanest code, and it restores the exact route at sample sizes that matter — a balanced two-sample design is currently pushed onto the normal approximation from about 33 per group. The cost is a hard floor on a release that is days old.
3. **Keep the guards but make them version-conditional.** More machinery than either bound is worth, in my view.

I have no stake in which; happy to push whichever you prefer. My mild preference is (1) for now and (2) whenever this package next raises its floor for unrelated reasons, since the guards are cheap and the floor is not.

`MAX_EXACT_ENUMERATION_N` is unaffected either way — it bounds the tied-data enumeration in this package, which StatsFuns has no part in.

One thing worth noting regardless of the choice: the two issues were only diagnosable because the failures were silent. Both routines returned plausible probabilities well before they threw. If it is of interest, the sharp frontier for the old behaviour was `binomial(nx + ny, nx)`, not `nx + ny` — largest correct split (30, 35), smallest wrong (31, 34) — which is why the constant above is a binomial rather than a sample size.

yoninazarathy:
Closing this in favour of three smaller PRs. @devmotion asked for #361 to be split into actionable pieces; this splits the implementation the same way, so that the review can separate what is additive from what changes numbers.

The split is by kind of change, not by size:

1. **[#367](https://github.com/JuliaStats/HypothesisTests.jl/pull/367) — additive.** Mann-Whitney `confint`, `hodgeslehmann` on all four rank types, a `method` keyword, non-`Float64` element types, and bounds on the two costs. Closes #364, #365 and the Mann-Whitney half of #8. Nothing this package returns today changes; where a fix is deferred, a test pins the current behaviour with a `PENDING` marker naming its issue.
2. **[Breaking signed rank fixes](https://github.com/yoninazarathy/HypothesisTests.jl/pull/2)** — the approximate test getting an approximate interval, zero differences no longer kept in an interval the statistic dropped, and the reported estimate becoming the Hodges-Lehmann one. Closes #361, #362 and #363. Every change here moves a returned value, and each flips one of the pinned assertions above, so its diff shows exactly what changes.
3. **[Mathematical specification appendix](https://github.com/yoninazarathy/HypothesisTests.jl/pull/3)** — documentation only, describing both rank procedures and the t-tests as they stand after (1) and (2).

2 and 3 are stacked on 1 and live on my fork, since a PR here cannot target a branch that does not exist in this repo. Both are complete and green; I will open them here as each predecessor merges. All three are drafts for now.

Everything in this PR is carried by one of the three. Nothing is dropped.

Issue #362, opened 2026-07-28
Zero differences are dropped for the signed-rank statistic but retained for its confidence interval


yoninazarathy:
`SignedRankTest` treats zero differences two different ways within the same object. [`signedrankstats`](https://github.com/JuliaStats/HypothesisTests.jl/blob/d2e20b5c69559f665bbc7fa10a23fc643a9540f9/src/wilcoxon.jl#L58-L59) discards them before ranking, so the statistic and p-value use the non-zero differences. [`calculate_ci`](https://github.com/JuliaStats/HypothesisTests.jl/blob/d2e20b5c69559f665bbc7fa10a23fc643a9540f9/src/wilcoxon.jl#L253-L254) takes `n = length(x)` on the stored `vals`, so the interval uses the Walsh averages of all of them.

## Demonstration

Dropping the zeros beforehand leaves the test bit-identical and yet moves the interval:

```julia
julia> using HypothesisTests

julia> d = [0.0,0,0,0.5,0.5,1,-0.5,-1,1.5,-1.5,0.5,0,1,-0.5,2,0,0.5,-1,1,0.5];  # five zeros

julia> nz = d[d .!= 0];

julia> SignedRankTest(d).W == SignedRankTest(nz).W
true

julia> pvalue(SignedRankTest(d)) == pvalue(SignedRankTest(nz))
true

julia> confint(SignedRankTest(d);  level = 0.9)
(0.0, 0.5)

julia> confint(SignedRankTest(nz); level = 0.9)
(-0.25, 0.75)
```

210 Walsh averages retaining the zeros against 120 dropping them; the implied estimate moves `0.25` → `0.5`. The p-value and the interval answer to different samples.

## Which convention

R's `wilcox.test` discards zeros for both, and lands on the zeros-dropped values:

```r
> d <- c(0,0,0,0.5,0.5,1,-0.5,-1,1.5,-1.5,0.5,0,1,-0.5,2,0,0.5,-1,1,0.5)
> w <- wilcox.test(d, conf.int = TRUE, conf.level = 0.9)
> w$estimate
(pseudo)median
     0.4999985
> w$conf.int
[1] -0.2499911  0.7500497
```

`0.4999985` is the Hodges–Lehmann estimate over the non-zero differences (`0.5`); over all of them it is `0.25`. The interval matches to ~1e-5, the tolerance of R's root-finding — R does not use plain order statistics under ties, so this corroborates rather than proves.

[...]

Issue #97, opened 2017-05-11
Wilcoxon crashes the REPL with enourmous data size


nico202:
length(Array(hc[:cycles])): 1000000

julia> ApproximateSignedRankTest(Array(hc[:cycles]), Array(rc[:cycles]))

Approximate Wilcoxon signed rank test
-------------------------------------
Population details:
    parameter of interest:   Location parameter (pseudomedian)
    value under h_0:         0
    point estimate:          -134.0
signrank allocation error

Is it possible to thraw the error *without* closing the session? Is frustrating to restart

andreasnoack:
I think it is difficult to control the behavior of Julia when you run out of memory. We might want to check if we can reduce the allocations of the function though.

PR #367, opened 2026-08-15
Mann-Whitney confidence intervals, Hodges-Lehmann estimates, and a method keyword

### LINKED ISSUE #361: `calculate_ci` uses the exact null distribution for `Approximate*` tests, and selects the wrong order statistic from n = 73

### LINKED ISSUE #364: Allow choosing between the exact and approximate methods for the Wilcoxon rank tests

### LINKED ISSUE #365: SignedRankTest rejects Float32, Float16, BigFloat and Rational input

yoninazarathy:
Closes #364, closes #365, and closes the Mann-Whitney half of #8.

**Additive**, with one exception stated in [What changes for existing callers](#what-changes-for-existing-callers): every number this package returns today, it still returns, except a signed rank `confint` on 224 or more observations, which is now refused rather than computed over tens of seconds, and one rounding corner noted there: a Mann-Whitney median reported from `Rational` or `BigFloat` samples now arrives converted to `Float64`. The signed rank interval and estimate, and the Mann-Whitney reported estimate, are deliberately left alone, since changing those is breaking. That is [the next PR](https://github.com/yoninazarathy/HypothesisTests.jl/pull/2).

This is the first of three replacing #366, which tried to do all of it in one change. #361 was split into actionable pieces on the issue side; this splits the implementation the same way.

**The specification these tests implement is rendered [here](https://yoninazarathy.github.io/HypothesisTests.jl/specs/rank_tests/), and is [PR 3](https://github.com/yoninazarathy/HypothesisTests.jl/pull/3).** It states the statistic, the null distribution, the estimator and the interval in terms that do not depend on this package, so the arithmetic below can be checked against something other than the diff. That is a preview build from my fork rather than official documentation. [What it previews](https://yoninazarathy.github.io/HypothesisTests.jl/specs/) is the appendix that PR 3 adds.

---

### `confint` and `hodgeslehmann` for the Mann-Whitney U tests

`MannWhitneyUTest` was the only location test in the package without a confidence interval, and it could not have had one: the structs kept only the pooled ranks, and the cross-group differences an interval is read off cannot be recovered from ranks. Both types now retain `x` and `y`.

`ExactMannWhitneyUTest` inverts the exact null distribution, conservatively. `ApproximateMannWhitneyUTest` inverts the normal approximation, with a continuity correction and the tie-corrected variance it already stores. Checked against R's `wilcox.test` with `exact` set explicitly, so that both implementations take the same route:

| case | this PR | R |
|---|---|---|
| 10 v 10, no ties, exact | `(-11.1, -0.1)` | `(-11.1, -0.1)` |
| 10 v 12, ties, exact | `(-14.0, -1.0)` | `(-13.99997, -1.00003)` |
| 9 v 9, ties, approximate | `(-0.37, 1.183)` | `(-0.36998, 1.18297)` |

These are the comparisons the test suite pins, with R's values in the comments. Rows agree digit for digit where R also reads order statistics. Where R takes its approximate route it solves for the endpoint numerically with `uniroot` instead, so its endpoints land near a pairwise estimate without being one. That is the difference in the second and third rows, not a disagreement.

`hodgeslehmann` is exported and defined on all four rank test types: the median of the Walsh averages for the signed rank tests, of the cross-group differences for the Mann-Whitney tests.

One consequence is worth stating plainly, since it is visible and frequent. On a signed rank sample containing zero differences the estimate drops them, as the statistic and the p-value already do and as R does, while `confint` does not. The two then describe different samples, and the estimate can fall outside the interval printed beside it. That is #362 rather than anything new here, and the next PR fixes it by dropping zeros from the interval too, but it happens on about 11% of samples that contain zeros, so it will be met. A test pins one: fourteen differences of which seven are zero give a p-value and an estimate over the seven non-zero ones, `3.0`, against an interval over all fourteen of `(0.0, 1.5)`. The interval of the sample the estimate actually describes is `(1.0, 3.0)`, and contains it.

What `show` **reports** as the point estimate is deliberately unchanged: it is still the difference of sample medians, which is a different quantity from the estimate the new interval is built around, and can fall outside it. Changing it moves a number users already see, so it belongs with #363 in [the next PR](https://github.com/yoninazarathy/HypothesisTests.jl/pull/2). A test pins the current value, `0.62`, against a Hodges-Lehmann estimate of `0.56`, with a `PENDING` marker until then.

The three affected `show` tests therefore keep their printed estimates unchanged and gain a confidence-interval line, which is new output rather than changed output.

### A `method` keyword (#364)

On `SignedRankTest` and `MannWhitneyUTest`: `:auto` (the existing thresholds), `:exact`, `:approximate`, or a callable receiving `(; n, n_nonzero, ties, tie_adjustment)` / `(; nx, ny, ties, tie_adjustment)`.

The thresholds are historical and have never been promised across releases, so an analysis that must reproduce had no way to pin the route it wanted. Now it does.

### Element types other than `Float64` (#365)

Both signed rank structs declared `median::Float64` while `signedrankstats` returns `median(x)`, so `SignedRankTest` rejected `Float32`, `Float16`, `BigFloat` and `Rational` outright. The seven-argument constructor now declares `median::Real` and the field converts, so every one of them constructs while the field itself stays `Float64`. Independent of everything else here.

### Bounds on the three costs (#97, part of #7)

`MAX_EXACT_ENUMERATION_N` bounds the tied-data enumerations, which were previously unbounded: `2^n` sign assignments, or `binomial(nx + ny, min(nx, ny))` rank assignments. `MAX_PAIRWISE_ESTIMATES` bounds the set of pairwise estimates, the Walsh averages or cross-group differences, which is quadratic in the sample size, and is a bound on memory. `MAX_EXACT_CI_ESTIMATES` bounds the *exact* interval, which spends time on top of that memory, a lattice recursion for every candidate endpoint it considers, and so sits far below the second: at the bound, a signed rank sample of 223, the endpoint scan takes about 9 s, and it grows steeply past it, 39 s at 300 and no finish at 400. Past it the exact interval is refused and the approximate one, where the test has one, is not.

All three raise rather than run, and all three raise `ComputationTooLarge`, which is its own type rather than an `ArgumentError` so that `show` can drop a refused interval line without also hiding real errors. The enumeration bound cannot be reached through `method = :auto`, which sends tied data to the normal approximation well before it. The other two can, since `confint` and `hodgeslehmann` form the set whichever route the p-value took, so `confint(SignedRankTest(randn(20_000)))` raises.

---

## `[compat] StatsFuns = "2.2.1"`

This is the one decision in the PR that is a judgement rather than a fix, and it is the question left open on #366. Our suggestion is to raise the lower bound, so that this package requires StatsFuns 2.2.1 or newer, rather than keep supporting the older versions and guard against their bug from inside this package.

`signrankcdf` and `wilcoxcdf` accumulated lattice counts in `Int` and returned silently plausible but invalid probabilities once those overflowed, from `n = 72` and once `binomial(nx + ny, nx)` exceeded about `3.0e18`, so the sharp frontier was a binomial rather than a sample size (largest valid split `(30, 35)`, smallest invalid `(31, 34)`). JuliaStats/StatsFuns.jl#219 and #220. Both were fixed in JuliaStats/StatsFuns.jl#221 by folding the normaliser into the recursions, released as StatsFuns 2.2.1. Worst absolute error against `Rmath` over the whole support, at `n = 200`: `0.4998` before, `2.9e-15` after. Before the fix the CDF is also non-monotone from `w = 771` and dips below zero, which is what makes the failure silent rather than loud.

Requiring 2.2.1 is what lets this PR add `method = :exact` without adding a correctness guard alongside it. Since `:exact` is new here, the alternative is not "do nothing". It is to carry two constants and two checks whose only purpose is to refuse on older StatsFuns. #366 did that. Requiring 2.2.1 instead deletes them, and restores the exact route at sizes where it previously could not be trusted: a balanced two-sample design was pushed onto the normal approximation from about 33 per group.

The cost is that installing this package would then require a StatsFuns released in July. We think that is the better side of the trade.

---

## What this deliberately does not do

#361, #362, #363 and #368. Those change numbers this package returns today, or which endpoint a one-sided interval returns, and they are [the next PR](https://github.com/yoninazarathy/HypothesisTests.jl/pull/2).

So that the change is visible rather than asserted, this PR pins the current, wrong behaviour with `PENDING` markers naming their issues, six markers over four behaviours: the approximate signed rank test still returning the exact interval, and the `method` choice therefore not yet reaching the interval (#361); the interval still keeping zero differences the statistic dropped (#362); and the reported estimate still being the sample median (#363). [The next PR](https://github.com/yoninazarathy/HypothesisTests.jl/pull/2) flips each one, so its diff shows exactly what moves.

---

## What changes for existing callers

No value changes into a different value, with one type-conversion corner. Every signed rank `pvalue` and `confint`, every Mann-Whitney `pvalue`, and every reported point estimate is identical to `master` wherever this branch returns one at all. The corner: a Mann-Whitney test built from `Rational` or `BigFloat` samples now stores and reports its median as `Float64`, so a value those types hold exactly and `Float64` cannot, a median of `1//3` say, arrives rounded. The signed rank tests have no such corner, since `master` refuses those element types outright (#365). The `calculate_ci` refactor here reuses the new `walsh_averages` and `ci_from_estimates` but leaves its selection rule untouched, and returns the same numbers.

Five things change without being a changed value:

| change | kind | why |
|---|---|---|
| `pvalue` on an `Exact*` test whose tied enumeration exceeds `MAX_EXACT_ENUMERATION_N` raises instead of enumerating | error where there was no result | on `master` this does not error, it hangs: 40 tied non-zero values did not return in five minutes |
| `confint` on a signed rank test past `MAX_PAIRWISE_ESTIMATES` raises up front | error where there was no result | on `master` this also hangs, in the selection scan: `confint(SignedRankTest(randn(20_000)))` did not return |
| `confint` on a signed rank sample of 224 or more raises | **error where `master` returned**, after a wait | `MAX_EXACT_CI_ESTIMATES`, and the one row here that refuses something `master` completes; see below |
| all four structs lose the type parameter they had; the six-argument positional Mann-Whitney constructor is gone | type level | every field is `Float64` now, so `ExactMannWhitneyUTest{Float64}` raises a `TypeError`; `confint` needs the samples, which cannot be recovered from ranks |
| `[compat] StatsFuns` requires 2.2.1 or newer | resolution | see above |

plus `hodgeslehmann` as a new export, which can only collide with a name.

The third row is worth reading twice, since it is the only place this PR withdraws something. `MAX_EXACT_CI_ESTIMATES = 25_000` is a bound on the `m` pairwise estimates, and `m = n(n+1)/2` passes it at `n = 224`, so a signed rank interval is refused from there up to where the memory bound takes over at `n = 1414`. On `master` those calls return: the endpoint scan costs `m/2` lattice recursions, which is 8.9 s at `n = 223`, 39 s at `n = 300`, and does not finish by `n = 400`. So the band this refuses is one where `master` is slow rather than broken, and on this branch there is no escape from it, since neither signed rank type has an approximate interval yet, which is #361.

The bound is sized against that scan, and the number is left alone here rather than tuned against code this branch does not contain. The next PR replaces the scan with a bisection, `log2(m/2)` recursions rather than `m/2`, gives the approximate type an interval of its own, and re-measures: the signed rank route sheds this bound entirely there, running to the memory bound in about 6 s, and the two-sample route keeps it at 90,000, the same wall-clock ceiling on the recursion that is now the expensive one. So this band exists only on this branch, and since the two PRs are one release, no caller meets it.

The type-level change is the one that stops this being a pure patch release. Every field on the four structs is `Float64`, so all four lose the parameter `master` gave them, and the Mann-Whitney pair gain `x` and `y` fields, so the six-argument positional constructor is gone. Fields are appended, so existing ones keep their positions, and nothing in the package or its test suite constructed these positionally.

Worth stating precisely, since it is a different break rather than a smaller one. `ExactMannWhitneyUTest{Float64}` is concrete on `master`; here it raises a `TypeError`, and the plain `ExactMannWhitneyUTest` is concrete instead, which it never was. A downstream annotation carrying the parameter has to drop it.

An earlier version of this description argued the type change was unavoidable, and that fixing the element type to `Vector{Float64}` would drop one parameter but lose #365. Both halves were wrong, and [these](https://github.com/JuliaStats/HypothesisTests.jl/pull/367#discussion_r3894215596) [comments](https://github.com/JuliaStats/HypothesisTests.jl/pull/367#discussion_r3897654704) are why it now reads as it does: fixing the element type drops every parameter rather than one, and it does not lose #365, which is about which samples construct at all rather than about which type a field carries. Whether all of this wants a minor or something else is your call.

**The version stays at 0.11.8 here, and the bump moves to PR 2.** `b378027` set it to 0.12.0 and `fa7f790` puts it back, for one reason: registration is not automatic on merge, but nothing stops an `@JuliaRegistrator register` on the merge commit, and with 0.12.0 in `Project.toml` that would succeed and ship the intermediate state. At 0.11.8, which is in General against tree `7040110b`, the same invocation is refused. So the plan of one release after both PRs is enforced by the tree rather than only stated in this description. `docs/Project.toml` follows it back, which leaves that file unchanged against `master`. If you would rather this PR be independently releasable, restoring the bump is a revert of that one commit and one line.

---

## One-sided intervals

The new Mann-Whitney `confint` follows the convention the signed rank tests use today, so that the four rank tests agree with each other. That convention is the opposite of the one the rest of this package and R use, which is #368; the PR stacked on this one flips all four together, which is where the breaking change belongs.

---

## Tests

Wilcoxon 249 (was 62), Mann-Whitney 271 (was 77). Common gains eight assertions pinning that a `ComputationTooLarge` refusal is the only thing `show` swallows, through stub `confint` methods that refuse, work, and fail two different ways. The only other test file touched is `test/show.jl`, for the reason at the end.

Against R, with the `wilcox.test` values recorded in the comments: the estimators on untied and tied data, and Mann-Whitney intervals on untied, tied and approximate samples. As properties: the estimate lies inside its own interval wherever the two describe the same sample, the `method` callable receives exactly the documented fields, element types other than `Float64` are accepted and converted rather than propagated, and all three bounds raise at their boundary.

Also as a property, and the sharpest check here: **both procedures reflect.** Swapping the two samples of a Mann-Whitney test negates the shift it is about, and negating the sample of a signed rank test negates the location it is about, so the estimate, both interval ends and the one-sided bound all flip while the p-value does not. It holds bitwise rather than approximately, since IEEE subtraction gives `xi - yj == -(yj - xi)` exactly and the index rules see only the sample sizes and the tie adjustment, so the tests assert `==` rather than a tolerance and an off-by-one in either index rule breaks them. Checked over 8000 random samples with ties and zeros before being pinned. The statistics are complementary rather than antisymmetric, `U + U' = nx*ny` and `W + W' = nz(nz+1)/2`, and are pinned that way. Two of the Mann-Whitney samples are separated deliberately: where `U` sits near its null mean the normal index rule cannot distinguish that mean from the centred statistic, and the reflection holds under either, so a testset built only from overlapping samples passes against that substitution.

That the exact route now works far past where the automatic rule abandons it, at `n = 100` for signed rank and `(40, 45)` for Mann-Whitney, where `binomial(85, 40)` is about `2.9e24`, is checked through the symmetry of the null distribution about its mean, which is the identity that failed on StatsFuns 2.2.0.

---

## One change that is not about the rank tests

It is here because CI was red without it. (Commit-wise the history carries two more that are not about the rank tests, `b378027` and `29d5739`, the 0.12.0 bump and its docs-compat companion, both since undone by `fa7f790` as described above; against `master` this is the only non-rank diff.)

`test/show.jl` compared the one-sample t-test summary against a literal string carrying `t-statistic: 2.100420126042015`. Statistics 1.11.4 returns `std(-5:10) == 4.760952285695234` where 1.11.1 returned `4.760952285695233`, so both that value and the empirical standard error move by one ulp and the comparison fails. This is `master`'s breakage rather than this branch's, and it is why the runs here went red on Julia 1 and Julia pre while min-patch, which resolves the older Statistics, stayed green. It matters here because the failure aborts `runtests.jl` at `show.jl`, before `test/wilcoxon.jl` is ever loaded, so none of the rank test suites had actually been running in CI. The two floats are interpolated rather than spelled out, so their presence, their full precision and their position are all still asserted and only their last bits are no longer pinned. Verified with a fresh resolve under Julia 1.12.7 and Statistics 1.11.4: the whole suite passes on `master` with just this change, and here. Happy to lift it into its own PR against `master` if you would rather have it separately.

---

## The other two

Both are written, green, and stacked on this branch. Since a fork PR cannot target another fork branch upstream, they are open on my fork so the incremental diffs are reviewable now, and I will open them here as each lands:

- **[PR 2](https://github.com/yoninazarathy/HypothesisTests.jl/pull/2)**: the breaking rank test fixes: #361, #362, #363 and #368.
- **[PR 3](https://github.com/yoninazarathy/HypothesisTests.jl/pull/3)**: documentation only: a mathematical specification of both rank procedures.


andreasnoack (commented):
**This entire review — this summary and every inline comment — was written by Claude Code (Opus 5), reviewing `e2cc1f9`. I've read through all of it before posting, and the reproductions and timings were checked in a scratch env resolved to StatsFuns 2.2.1.**

Starting with the good news: the full suite is green on the branch (Wilcoxon 228, Mann-Whitney 255), and the numerics hold up under scrutiny. The index rules match R's `qwilcox`/`qsignrank` construction (`k = qu - 1`, including R's `qu == 0 -> 1` special case), the `:auto` rules are byte-equivalent to the old thresholds, `last_true` terminates and is correct, and no inverted interval seems constructible (`k <= div(m,2)-1` is always achieved for `sigma > 0`, and `cdf(div(m,2)) >= 0.5 > alpha/2` blocks the exact rule).

So nothing below is a wrong number. Four look like real bugs, two are cost/limit calibration, and the rest are API decisions that are cheaper to settle now than after release. The first one is worth a look before anything else: the guard that #97 is meant to fix doesn't appear to bound what its comment says it bounds, so the hang is still reachable.

src/wilcoxon.jl
```diff
@@ -252,27 +306,15 @@ function calculate_ci(x::AbstractVector, level::Real=0.95; tail=:both)
     end
     n = length(x)
     m = div(n * (n + 1), 2)
+    # bound the set before the k scan below, which alone runs m/2 lattice
+    # recursions and would hang long before `walsh_averages` refuses
+    check_estimate_count(m, "Walsh averages")
```
andreasnoack, line 311: **The Walsh-average scan is still effectively unbounded — the hang #97 describes looks reachable.**

`check_estimate_count(m, "Walsh averages")` bounds `m`, the size of the materialised set, but the cost here is the *scan*: `div(m,2)` calls to `signrankcdf`, each a lattice DP. Measured `confint(SignedRankTest(randn(n)))`:

| n | time |
|---|---|
| 100 | 0.2 s |
| 150 | 1.7 s |
| 200 | 5.9 s |
| 400 | did not finish in 100 s |
| 800 | did not finish in 100 s |

`m <= 100_000_000` only trips at n ~ 14142, by which point the scan is astronomically past feasible. So `confint(SignedRankTest(randn(2000)))` still hangs indefinitely, and so does `show` of it.

The bound wants to be on the scan length (or the exact route needs to hand off to the normal rule above some n), not on the size of the estimate vector.

---
<sub>Written by Claude Code (Opus 5); reviewed by me before posting.</sub>

src/mann_whitney.jl
```diff
@@ -69,18 +88,22 @@ function mwustats(x::AbstractVector{<:Real}, y::AbstractVector{<:Real})
         # U = (nx + ny)*(nx + ny + 1)/2 - sum(ranks_y) - nx*(nx + 1)/2
         U = ny*(2 * nx + ny + 1)/2 - sum(@view(ranks[(begin + nx):end]))
     end
-    return (U, ranks, tieadj, nx, ny, median(x)-median(y))
+    R = promote_type(eltype(x), eltype(y))
+    return (U, ranks, tieadj, nx, ny, median(x)-median(y),
+            convert(Vector{R}, x), convert(Vector{R}, y))
```
andreasnoack, line 93: **Both MWU structs alias the caller's arrays.**

`convert(Vector{R}, x)` is a no-op when the eltype already matches, so `x` and `y` are stored by reference rather than copied. Every other field is a snapshot taken at construction, so these two are the only live views, and they're exactly the ones the new `confint`/`hodgeslehmann` read:

```julia
x = [...]; y = [...]
t = MannWhitneyUTest(x, y)
hodgeslehmann(t)   # -5.6
x[1] = 1000.0
hodgeslehmann(t)   # -4.1  -- while t.U, t.ranks, t.median are unchanged
```

`confint(t)` moves with it too, so a test object can report a p-value and an interval that describe different samples.

`Vector{R}(x)` converts *and* copies (unlike `convert`, which aliases on an eltype match), so it's a one-word fix.

---
<sub>Written by Claude Code (Opus 5); reviewed by me before posting.</sub>

src/rank_common.jl
```diff
@@ -0,0 +1,261 @@
...
+"""
+function normal_ci_index(m::Integer, mu::Real, sigma::Real, alpha::Real)
+    z = quantile(Normal(), 1 - alpha / 2)
+    return clamp(ceil(Int, mu - z * sigma - 0.5) - 1, 0, div(m, 2))
+end
+
+# `level`/`tail` to the two-sided alpha the index rules work in. A one-sided bound
+# at `level` is the corresponding endpoint of the two-sided interval at `2 level - 1`.
+function ci_alpha(level::Real, tail::Symbol)
+    check_level(level)
```
andreasnoack, line 226: **`check_level` is `Float64`-only, so the new `level::Real` signature throws on any other real.**

`ci_alpha` takes `level::Real`, but `check_level` is declared `check_level(level::Float64)`. Both of these are `MethodError: no method matching check_level(::Float32)` rather than working:

```julia
confint(MannWhitneyUTest(x, y); level = 0.9f0)
confint(MannWhitneyUTest(x, y); level = 9//10)
```

Every other `confint` in the package declares `level::Float64` and gets the conversion for free at the call boundary; the new MWU signature is the first `level::Real`, which is a nice widening but needs `check_level(float(level))` to go with it.

---
<sub>Written by Claude Code (Opus 5); reviewed by me before posting.</sub>

src/mann_whitney.jl
```diff
@@ -174,15 +204,30 @@ function StatsAPI.pvalue(x::ExactMannWhitneyUTest; tail=:both)
...
+hodgeslehmann(x::ExactMannWhitneyUTest) = median(cross_differences(x.x, x.y))
+
+# Under ties the achieved coverage is approximate rather than exact: the null
+# distribution inverted here is the untied one, where the tied data has the
+# conditional distribution `mwuenumerate` computes. The classical construction
+# inverts the untied one anyway; R declines an exact interval in that case.
+function StatsAPI.confint(x::ExactMannWhitneyUTest; level::Real=0.95, tail=:both)
+    alpha = ci_alpha(level, tail)
+    vals = cross_differences(x.x, x.y)
+    k = exact_ci_index(length(vals), alpha, u -> wilcoxcdf(x.nx, x.ny, u))
```
andreasnoack, line 216: **The exact MWU `confint` is unbounded in time, and `method = :exact` makes it reachable at any size.**

`exact_ci_index` runs ~log2(nx*ny/2) `wilcoxcdf` lattice DPs, each O(nx*ny*min(nx,ny)) — so roughly n^4 overall. Measured with `method = :exact`:

| nx v ny | time |
|---|---|
| 100 v 100 | 0.1 s |
| 200 v 200 | 1.9 s |
| 300 v 300 | 11.7 s |

600 v 600 is minutes and 1000 v 1000 effectively hangs — and `show` triggers it, so a REPL display is enough.

`MAX_EXACT_ENUMERATION_N` only bounds the *tied* enumeration; the untied route has no bound at all, and the docstring at src/rank_common.jl:66 actively tells the reader that `confint` is unaffected. That was defensible while the exact route was only reachable through the `:auto` sample-size rule, but `method = :exact` is new API that opts into it at any size.

---
<sub>Written by Claude Code (Opus 5); reviewed by me before posting.</sub>

src/rank_common.jl
```diff
@@ -0,0 +1,261 @@
...
+## PAIRWISE ESTIMATES
+
+"""
+Largest set of pairwise estimates (Walsh averages, or cross-group differences) that will be
+materialised. The estimators and intervals below are ``O(n^2)`` in memory; beyond
+this bound they refuse rather than exhaust the machine. Displaying such a test still
+works: `show` drops its confidence interval line rather than failing to print at all.
+See #7 and #97.
+"""
+const MAX_PAIRWISE_ESTIMATES = 100_000_000
```
andreasnoack, line 109: **`MAX_PAIRWISE_ESTIMATES = 1e8` is too loose for the "refuse rather than exhaust the machine" job.**

Displaying a Mann-Whitney test used to be free; it's now O(nx*ny) in memory. Measured `sprint(show, MannWhitneyUTest(randn(n), randn(n)))`:

| n v n | allocated |
|---|---|
| 2000 | 68 MiB |
| 5000 | 439 MiB |
| 8000 | 1.03 GiB |

At the limit itself (10000 v 10000) it's ~1.6 GiB — all of it spent on one display line whose printed point estimate is still the sample median (#363). `hodgeslehmann` is worse again, since `Statistics.median` copies, putting the peak at 2x the vector.

A bound two or three orders of magnitude lower would still be far above any sample where the quadratic set is a sensible thing to form.

---
<sub>Written by Claude Code (Opus 5); reviewed by me before posting.</sub>

src/rank_common.jl
```diff
@@ -0,0 +1,261 @@
...
+
+The signed rank test enumerates ``2^n`` sign assignments over its ``n`` non-zero
+observations, and refuses past `n = MAX_EXACT_ENUMERATION_N`. The Mann-Whitney test
+enumerates ``\\binom{n_x + n_y}{\\min(n_x, n_y)}`` rank assignments, so the bound is on that
+count rather than on the sample size: it refuses once there are more than
+`2^MAX_EXACT_ENUMERATION_N` of them. Either way the p-value raises rather than run for an
+unbounded time, and `method = :approximate` is the way on; `confint` is unaffected, since it
+inverts the tie-free distribution whichever route the p-value took. See #7.
+"""
+const MAX_EXACT_ENUMERATION_N = 25
```
andreasnoack, line 69: **25 is well below the pain threshold, and there's no override.**

At n = 25 the tied signed-rank enumeration takes 0.41 s, so this hard-errors on a range that master handles in under a second:

```julia
pvalue(ExactSignedRankTest(x))  # 26 tied non-zero differences -> ArgumentError
```

n = 26-30 costs 0.8-7 s on master. The MWU side has the same shape: 15 v 15 tied is `binomial(30,15)` = 155 M, ~19 s extrapolated from a measured 12 v 12, and is refused.

The only documented escape is `method = :approximate`, which answers a different question — so a user who wants the exact tied p-value at n = 27 now has no way to ask for it. Since this is a `const` with no keyword or scoped override, that's permanent. A substantially higher bound, or an opt-in override for callers willing to wait, would keep the guard useful without removing capability.

---
<sub>Written by Claude Code (Opus 5); reviewed by me before posting.</sub>

src/wilcoxon.jl
```diff
@@ -163,19 +186,38 @@ function StatsAPI.pvalue(x::ExactSignedRankTest; tail=:both)
...
+# The Walsh averages that the point estimate is read off.
+#
+# `signedrankstats` drops zero differences before ranking, so the statistic and the
+# p-value describe the non-zero differences, and the estimate must describe the same
+# sample. R's `wilcox.test` drops them likewise.
+#
+# `confint` does not yet use this: it still forms its Walsh averages from every
+# difference, zeros included, so the interval and the estimate can describe
+# different samples. See #362.
+function signedrank_pairwise_estimates(vals::AbstractVector{<:Real})
```
andreasnoack, line 198: **`hodgeslehmann` and `confint` describe different samples, and this new function is the half that could avoid it.**

The comment above is clear-eyed about the split — `hodgeslehmann` drops zeros, `confint` keeps them, tracked as #362. The consequence is that `lo <= hodgeslehmann(t) <= hi` fails on roughly 11% of samples containing zeros, and the PR's own test pins the contradiction: `hodgeslehmann == 3.0` against `confint == (0.0, 1.5)`.

The asymmetry worth noting is that `confint`'s behaviour is constrained by compatibility, but `hodgeslehmann` is brand-new API with no callers to preserve. Having it keep zeros now — matching `confint` — and switching both in the #362 follow-up means no released version ever contains a self-contradictory pair. As written, 0.11.x ships with an estimate that can sit outside its own interval.

---
<sub>Written by Claude Code (Opus 5); reviewed by me before posting.</sub>

src/rank_common.jl
```diff
@@ -0,0 +1,261 @@
...
+
+# `level`/`tail` to the two-sided alpha the index rules work in. A one-sided bound
+# at `level` is the corresponding endpoint of the two-sided interval at `2 level - 1`.
+function ci_alpha(level::Real, tail::Symbol)
+    check_level(level)
+    check_tail(tail)
+    return tail === :both ? 1 - level : 2 * (1 - level)
+end
+
+function ci_from_estimates(vals::AbstractVector, k::Integer, tail::Symbol)
```
andreasnoack, line 231: **This gives the new MWU `confint` the inverted one-sided convention, so it breaks again in the next release.**

`tail = :left` returns `(lo, Inf)` and `tail = :right` returns `(-Inf, hi)`, which is backwards relative to `t.jl`, `z.jl`, `binomial.jl`, and R. It's internally consistent with the signed-rank tests, so as a shared helper it's the right call for those — but the MWU `confint` methods are new public API, and #368 is going to flip them.

Worth deciding deliberately rather than by inheritance: a brand-new method born on the known-wrong side of an acknowledged bug costs a second breaking change for the same users.

---
<sub>Written by Claude Code (Opus 5); reviewed by me before posting.</sub>

src/HypothesisTests.jl
```diff
@@ -51,14 +51,30 @@ function check_tail(tail::Symbol)
     end
 end
 
+# The interval `show` prints beside the point estimate, or `nothing` where there is none
+# to print. A test whose `confint` refuses the sample it was given still has a name, a
+# statistic and a p-value worth displaying, so the offending line is dropped rather than
+# the whole display: the rank tests bound the pairwise estimates their intervals are read
+# off (see `MAX_PAIRWISE_ESTIMATES`), and past that bound a large `MannWhitneyUTest` would
+# otherwise be impossible to look at.
+function show_confint(test::HypothesisTest)
```
andreasnoack, line 60: **This swallows every `ArgumentError` from every `confint`, including third-party ones.**

The intent is clear — suppress the interval line when the estimate set is too large to form — but `show_confint` is generic over `HypothesisTest`, so it also catches genuine argument bugs in any downstream package's `confint`. Those now silently print a test with no interval line instead of surfacing a stack trace, which is a rough thing to debug.

Narrowing it would fix that: a dedicated exception type for the bound violation, or re-checking the estimate count here before calling `confint` at all.

---
<sub>Written by Claude Code (Opus 5); reviewed by me before posting.</sub>

src/rank_common.jl
```diff
@@ -0,0 +1,261 @@
...
+Their median is the one-sample Hodges-Lehmann estimator, and the distribution-free
+interval for the pseudomedian is a pair of their order statistics.
+"""
+function walsh_averages(x::AbstractVector{T}) where T<:Real
+    n = length(x)
+    m = div(n * (n + 1), 2)
+    check_estimate_count(m, "Walsh averages")
+    out = Vector{float(T)}(undef, m)
+    k = 0
+    @inbounds for i in 1:n, j in i:n
```
andreasnoack, line 132: **`@inbounds` over `1:n` rather than `eachindex` is a silent-corruption trap.**

The code this replaces iterated with `enumerate`, which is index-agnostic; `x[i] for i in 1:n` under `@inbounds` assumes 1-based indexing. It's latent today, since the only callers pass the `Vector` struct fields — but an `OffsetArray` reaching here would write out of bounds silently instead of erroring, and `walsh_averages` is now shared machinery that's easy to call from somewhere new.

`eachindex`-based loops keep the optimisation and remove the trap.

---
<sub>Written by Claude Code (Opus 5); reviewed by me before posting.</sub>

Project.toml
```diff
@@ -23,7 +23,7 @@ StableRNGs = "1"
 Statistics = "1"
 StatsAPI = "1.6"
 StatsBase = "0.27, 0.28, 0.29, 0.30, 0.31, 0.32, 0.33, 0.34"
-StatsFuns = "1.4, 2"
+StatsFuns = "2.2.1"
```
andreasnoack, line 26: **This needs at least a minor version bump.**

Two separate reasons the current `version = "0.11.8"` won't do:

- Dropping StatsFuns 1.x from `[compat]` is itself breaking for downstream resolution.
- The change is breaking at the type level: all four structs gain a type parameter, so `ExactMannWhitneyUTest{Float64}` stops being a concrete type and existing annotations stop matching; the 6-arg positional MWU constructor is gone; and `hodgeslehmann` is a new export that can collide.

Released as a patch, that breaks downstream type annotations and pins with no signal that anything changed.

---
<sub>Written by Claude Code (Opus 5); reviewed by me before posting.</sub>

src/wilcoxon.jl
yoninazarathy: Fixed in 153472c, and your timings reproduced closely where I measured them: n = 200 at
5.2 s against your 5.9, then n = 250 at 15.8 s and n = 300 at 39 s. I stopped there rather
than confirm your n = 400.

You offered two routes, a bound on the scan or a hand-off to the normal rule. It had to
be the bound: on this branch there is no normal-approximation signed rank interval to
hand off to, since `calculate_ci` is what both types call. `MAX_EXACT_CI_ESTIMATES`
bounds `m` at 25 000, which is a sample of 223 and about 9 s. `confint` past that raises
promptly, and `show` of it still prints, without the interval line.

Two things worth saying plainly about that number. It is deliberately far below
`MAX_PAIRWISE_ESTIMATES`, because the two bound different things: that one bounds the
array, this one bounds a lattice recursion per candidate endpoint. And it removes reach
that `master` has, since `master` answers slowly up to about 300 where this refuses past
223. It is hit by default rather than only under `method = :exact`, because #361 means
both signed rank types invert the exact distribution. That is only acceptable because
the stacked PR gives the approximate type a closed-form interval and both land in one
release, so no released version loses the capability. If you would rather these ship
separately, this bound is too aggressive and should be raised.

---
<sub>Written by Claude Code (Opus 5).</sub>

src/mann_whitney.jl
yoninazarathy: Fixed in 7bd541d, `Vector{R}(x)` as you suggested.

Chasing it turned out to be worth more than the one-word fix. The three signed rank
constructors had the same defect by a different route, handing `x` straight to a field
typed `Vector{T}` where the assignment's conversion is the same no-op, and that one
predates this stack: `master` already stored `vals` and read it in `confint`. So

```julia
x = [1.0, 2.0, 3.0, 4.0, 9.0, -2.0, 0.5]
s = ExactSignedRankTest(x)
pvalue(s)   # 0.09375        confint(s)  # (-0.5, 6.0)
x[1] = 500.0
pvalue(s)   # 0.09375        confint(s)  # (0.0, 252.0)
```

a frozen p-value printed beside a moving interval. Both are fixed, and the copy sits in
whichever constructor is last on each route so that no route copies twice and none can
skip it: `ApproximateSignedRankTest`'s seven-argument method takes its own, which also
covers a direct call to it, while `ExactSignedRankTest` has no method to hook, seven
arguments being its full field list, so its two call sites copy.

Then I checked the rest of the package. Four other tests keep a caller-supplied array,
and two do it observably: `DurbinWatsonTest` moves its p-value from 0.176 to 0.219 while
`DW` stays frozen, and `FisherTLinearAssociation` reports `rho_t` of essentially 1
beside a p-value of 0.2. `BoxPierceTest` and `LjungBoxTest` alias but never read, so
they are latent. That is a package-wide question about whether construction snapshots
its input, which has never been raised here, so it is #375 rather than something I
answered in this PR. A blanket sweep would be wrong for half of them: the two inert
ones probably want the field deleted rather than copied.

---
<sub>Written by Claude Code (Opus 5).</sub>

src/rank_common.jl
yoninazarathy: Fixed in 47af9f9. I widened `check_level` to `check_level(level::Real)` rather than
calling `check_level(float(level))` at the call site, so that the next `confint` to
declare `level::Real` gets it for free instead of rediscovering this. The two
comparisons hold for any real, and the seven methods that declare `Float64` still
convert at their own boundary, so nothing else moves.

Checked past the MethodError rather than just to it: `Float32` and `Rational{Int}`
levels now return the same endpoints as `Float64` on all four rank tests, exact and
approximate, and a level outside `(0.5, 1)` is still rejected whichever real type it
arrives as. The endpoints do not take the level's type, since they are order statistics
of the sample.

---
<sub>Written by Claude Code (Opus 5).</sub>

src/mann_whitney.jl
yoninazarathy: Fixed in 153472c by the same `MAX_EXACT_CI_ESTIMATES`, and your timings reproduced:
300v300 at 10.9 s against your 11.7, 400v400 at 35 s. The bound admits 158v158 at 0.67 s
and refuses 159v159; the approximate route is untouched and still answers at 1000v1000.

The check runs before `cross_differences` rather than after, so an oversized request is
refused rather than paid for first.

You also flagged the docstring that told the reader `confint` was unaffected. That is
fixed: it now says this bound does not reach `confint`, and points at the one that does.

One consequence to note: this single constant is sized by the signed rank scan, which is
about ten times more expensive than the two-sample bisection at the same `m`, so the
Mann-Whitney interval is bounded below what it could afford, 25 000 where 90 000 would
cost 11 s. I took that as acceptable because `:auto` never selects the exact route above
`nx + ny = 50`, so reaching the bound already requires a deliberate `method = :exact` on
a large sample, and `:approximate` is right there. Happy to split it into two constants
if you would rather not under-restrict the new method.

---
<sub>Written by Claude Code (Opus 5).</sub>

src/rank_common.jl
yoninazarathy: Agreed and fixed in 153472c: 1e8 to 1e6, which is about 16 MB rather than your measured
1.6 GiB, and still admits a signed rank sample of roughly 1400 and a design of 1000
against 1000. The docstring now says it bounds memory alone, which is what the
approximate routes spend, and points at the separate bound on the exact ones.

One incidental tidy: the `show`-past-the-bound test no longer needs 10 001 against 10 001
to exceed the bound, so it now uses 1001 against 1001. That does not show up as a faster
file, since the new exact-interval test below costs more than it saved, 3.4 s to 3.9 s
overall.

---
<sub>Written by Claude Code (Opus 5).</sub>

src/rank_common.jl
yoninazarathy: Raised to 30 in 153472c. Measured on this branch, with the bound lifted so the costs
could be seen:

| n | 2^n | seconds | | nx v ny | binomial | seconds |
|---|---|---|---|---|---|---|
| 26 | 6.7e7 | 0.8 | | 14 v 14 | 4.0e7 | 1.2 |
| 28 | 2.7e8 | 3.3 | | 15 v 15 | 1.6e8 | 4.8 |
| 30 | 1.1e9 | 14.2 | | 16 v 16 | 6.0e8 | 20.0 |
| 32 | 4.3e9 | 64.9 | | | | |

So 30 restores exactly the 26-to-30 range you named, and still refuses the doubling
past it. 32 would be 65 s, which is not a bound.

The second half of your comment I have **not** addressed: there is still no override, so
a caller willing to wait for a tied exact p-value at n = 31 has no way to ask. I left it
out because it is new API surface and I would rather you chose its shape than inherit
mine, but I do not think the argument against it is strong. If you want one, say whether
you would prefer a keyword on the test, a scoped `with_limits`-style block, or simply
making the constants non-const and settable, and I will add it here.

---
<sub>Written by Claude Code (Opus 5).</sub>

src/wilcoxon.jl
yoninazarathy: You are right that this is the half with no callers to preserve, and your reasoning holds
under the assumption that this PR is released on its own. I would rather change the
assumption than the estimator, because your suggestion asks a brand-new function to be
knowingly wrong, dropping the agreement with the statistic, the p-value and R, in order
to be consistent with a bug.

Your comment on `Project.toml` is what makes the alternative available. Once this is a
minor release rather than a patch, there is no version of the plan where users cross only
one boundary for this PR alone, so the additive framing has already lost its purpose. So
the proposal is to land both PRs and register 0.12.0 only after the second, which
satisfies the criterion you actually stated, that no *release* ships an estimate outside
its own interval, without asking the estimator to be wrong in the meantime.

For the record the contradiction is pinned, unchanged, in `test/wilcoxon.jl`:
`e = [0.0, 3, 2, 0, 0, 3, 0, -1, 0, 0, 3, 0, 3, 3]` gives `hodgeslehmann == 3.0` against
`confint == (0.0, 1.5)`, where the interval of the sample the estimate actually describes
is `(1.0, 3.0)`.

If you would rather not tie the two together, tell me and I will make `hodgeslehmann`
keep zeros as you suggest.

---
<sub>Written by Claude Code (Opus 5).</sub>

src/rank_common.jl
yoninazarathy: Same answer, and here partial adoption is not just less good but incoherent. Flipping
only the new Mann-Whitney methods would leave the four rank tests disagreeing with each
other, which is the property this PR's description gives as the reason for following the
existing convention in the first place. So the choice is all four now or all four before
release, not one now.

The stacked PR flips all four together, along with the pairing that makes it a real bug
rather than a naming one: on the same object `pvalue(t; tail = :left)` tests location
below the null while `confint(t; tail = :left)` currently returns the acceptance region of
the right-tailed test. Under the single-release plan no released version returns the wrong
endpoint from these methods.

Unchanged here, for the record: `confint(s; tail = :left)` returns `(0.5, Inf)` where
`confint(OneSampleTTest(x); tail = :left)` returns `(-Inf, 5.04)`.

---
<sub>Written by Claude Code (Opus 5).</sub>

src/HypothesisTests.jl
yoninazarathy: Fixed in 04955eb, by the first of the two routes you offered: `ComputationTooLarge` is
thrown by all three cost bounds and is the only thing `show_confint` catches. Re-checking
the count in `show_confint` would have meant that generic function knowing which tests
have quadratic estimate sets, which seemed worse.

Worth noting how thin the existing coverage was. `test/common.jl` did have a stub for
"a bug, not a refusal", but it raises through `error`, so it is an `ErrorException` and
was never caught by the old `ArgumentError` clause. The case you describe, a downstream
`ArgumentError`, had no test at all. It does now, and it propagates through both
`show_confint` and `show`.

Since these bounds are new in this PR, nothing released ever threw `ArgumentError` from
them, so the change of type breaks no existing caller.

---
<sub>Written by Claude Code (Opus 5).</sub>

src/rank_common.jl
yoninazarathy: Fixed in 9fb821a, though not the way you suggested, so it is worth saying why.

`walsh_averages` runs a triangular loop, `for i in 1:n, j in i:n`, where the inner range
depends on the outer index. There is no clean `eachindex` substitution for that: the
index-agnostic forms need `Iterators.drop` or positional slicing of an index range, and
the latter is wrong for a genuinely offset array anyway. Since the goal is to stop the
silent out-of-bounds read rather than to support offset arrays, and since every caller
passes a `Vector` struct field, `Base.require_one_based_indexing` states the assumption
and turns that read into an `ArgumentError`, with the loop and its speed untouched.

Its sibling needed nothing, which is worth recording: `cross_differences` iterates the
values, `for xi in x, yj in y`, and indexes only the output it just allocated, so it has
no such assumption. A test pins the asymmetry, that `cross_differences` accepts a
zero-based vector `walsh_averages` refuses. It builds a three-element stub rather than
taking a dependency on OffsetArrays for one assertion.

---
<sub>Written by Claude Code (Opus 5).</sub>

Project.toml
yoninazarathy: Agreed, and done in b378027: 0.12.0, for both of the reasons you give.

This comment is also what unblocked the two API decisions above. While the PR was
framed as patch-releasable, the only way to keep new API self-consistent was to make it
consistently wrong; once it is a minor release, fixing #362 and #368 before anything is
tagged is available instead. So this comment is upstream of the replies to
3843642330 and 3843642338.

One addition to your list, though it is not a break for anyone: `ComputationTooLarge`
replaces `ArgumentError` from the cost bounds, and since those bounds are new here, no
released version ever threw the old type.

---
<sub>Written by Claude Code (Opus 5).</sub>

yoninazarathy:
All eleven comments now have something behind them, six new commits above `e2cc1f9`, and
the suite is green at Wilcoxon 263 and Mann-Whitney 283 against 228 and 255 when you
reviewed.

Nine are fixed in code. Two, the zeros in `hodgeslehmann` and the one-sided convention,
are answered by a release plan rather than a change, and that is the one thing I need a
decision on.

The proposal: keep the two PRs separate, land both, and register 0.12.0 only after the
second. Your `Project.toml` comment is what makes this the better option. It satisfies
the criterion you stated, that no release contains an estimate outside its own interval
or an interval on the known-wrong side of #368, without asking brand-new API to be
knowingly wrong in the interim, and without collapsing the branches, which would
invalidate the anchoring of the review you have already done. If you would rather review
one diff, I will merge them and rewrite the description around a single release instead.

Two of the fixes reach further than asked, so flag anything you would rather I narrowed.
The aliasing fix covers the three signed rank constructors as well, where the defect
predates this stack, and the four other tests in the package that share it are #375
rather than this PR. And the interval bound covers the exact Mann-Whitney route as well
as the signed rank scan, since `method = :exact` opts into the same unbounded cost.

One thing left undone from your review, called out in its own reply: there is still no
opt-in override for the enumeration bound, only the higher value. Tell me what shape you
want and I will add it.

---
<sub>Written by Claude Code (Opus 5).</sub>

src/mann_whitney.jl
yoninazarathy: I am going to argue against my own fix here, and have reverted it in 1b42821; if you
are unconvinced, restoring it is a `git revert` of that one commit.

What changed my mind is the survey your comment prompted (#375). Five other tests in
this package keep a caller-supplied array, and every one of them keeps the array
itself, three observably so at `confint` or `pvalue` time. Copying inside the four
rank tests would have made them the only types with snapshot semantics, which decides
the package-wide question that issue asks silently, in the narrowest place, for some
tests and not others.

The copy was never the cost: one allocation under an n log n construction. The
objection is where the decision gets made. Not mutating an argument you have handed
to a constructor is the contract every other array-keeping test here already relies
on, and if #375 settles on snapshotting instead, restoring this is the same one-word
change, applied to every test at once rather than to four.

What your repro shows stays true, on this branch as on master for the signed rank
tests: mutate the input and the interval and the estimate move while the statistic
stays frozen. I have updated #375 so the rank tests are back in its scope rather than
listed as fixed here.

<sub>Written by Claude Code (Fable 5).</sub>

src/mann_whitney.jl
```diff
@@ -41,23 +41,42 @@ The Mann-Whitney U test is sometimes known as the Wilcoxon rank-sum test.
...
+    if resolve_rank_method(method, stats, auto_mwu_method(stats)) === :exact
+        ExactMannWhitneyUTest(stats0...)
     else
-        ApproximateMannWhitneyUTest(U, ranks, tieadj, nx, ny, median)
+        ApproximateMannWhitneyUTest(stats0...)
     end
 end
 
+mwu_decision_stats(nx, ny, tie_adjustment) =
+    (nx = nx, ny = ny, ties = tie_adjustment != 0, tie_adjustment = tie_adjustment)
```
andreasnoack: This function doesn't seem to add much value. It's trivial and used once.

src/mann_whitney.jl
```diff
@@ -41,23 +41,42 @@ The Mann-Whitney U test is sometimes known as the Wilcoxon rank-sum test.
...
+
+auto_mwu_method(s) =
```
andreasnoack: Same with this one

src/mann_whitney.jl
```diff
@@ -69,18 +88,22 @@ function mwustats(x::AbstractVector{<:Real}, y::AbstractVector{<:Real})
...
+struct ExactMannWhitneyUTest{T<:Real,S<:Real} <: HypothesisTest
     U::Float64              # test statistic: Mann-Whitney-U statistic
     ranks::Vector{Float64}  # ranks
     tie_adjustment::Float64 # adjustment for ties
     nx::Int                 # number of observations
     ny::Int
-    median::T               # sample median
+    median::T               # difference of sample medians
+    x::Vector{S}            # original values, first sample
+    y::Vector{S}            # original values, second sample
```
andreasnoack: It seems a bit inconsistent to have some of these quantities hard coded for `Float64` while others are not. Maybe safer to hard code everything for `Float64` and delegate work towards making this parametric (with appropriate testing) to another PR for another day.

yoninazarathy:
Three things since the summary above, all from the last hour.

**The ledger has moved by one.** The aliasing fix has since been argued back out and reverted in 1b42821, with the reasoning in [that thread](https://github.com/JuliaStats/HypothesisTests.jl/pull/367#discussion_r3894100410). So it is now eight of your comments fixed in code and three answered by argument rather than by a commit: the zeros in `hodgeslehmann`, the one-sided convention, and the aliasing, whose policy question belongs to #375 rather than to this PR. Everything else in that summary stands.

**CI is green again, and the two commits at the tip are why.** It had been red since this morning, for two reasons, neither of them the rank tests:

- `docs/Project.toml` still pinned `HypothesisTests = "0.11"`, so the 0.12.0 bump made the Documentation job fail at the `Pkg.develop` that `julia-docdeploy` runs, with `empty intersection between HypothesisTests@0.12.0 and project compatibility 0.11`, before it built anything. That one is mine, and is 29d5739. The doc build now runs clean, including the `@docs` blocks this branch adds, which `checkdocs = :exports` had therefore never seen.
- `test/show.jl` compares the one-sample t-test summary against a literal string carrying `t-statistic: 2.100420126042015`. Statistics 1.11.4 returns `std(-5:10) == 4.760952285695234` where 1.11.1 returned `4.760952285695233`, so that value and the empirical standard error each move by one ulp. That one is `master`'s rather than this branch's: it is why Julia 1 and Julia pre went red while min-patch, which resolves the older Statistics, stayed green, and `master` fails the same assertion with a fresh resolve. 5b48276 interpolates the two floats instead of spelling them out, so their presence, their full precision and their position are all still asserted and only their last bits are not. Happy to lift it into its own PR against `master` if you would rather have it separately.

The second had a consequence worth knowing about. That failure aborts `runtests.jl` at `show.jl`, which is included before `test/wilcoxon.jl`, so for those runs **none of the rank test suites were actually running in CI**. They are now: Wilcoxon 246 and Mann-Whitney 268, green on every job.

**Three corrections to the PR body**, found while checking it against the code: the test counts were stale, "Bounds on the two costs" is now three and names `ComputationTooLarge`, and the opening "additive" claim now carries the exception it always had, that `MAX_EXACT_CI_ESTIMATES` refuses a signed rank interval from `n = 224` where `master` answers slowly up to about 300. That was already stated plainly in [the bound thread](https://github.com/JuliaStats/HypothesisTests.jl/pull/367#discussion_r3893460396), and should not have been missing from the body.

PR 2 is rebased on this tip, and its suite is green. Its red mark on the fork is the codecov step, which cannot upload without a token on a fork; `julia-runtest` passes on all nine test jobs there.

---
<sub>Written by Claude Code (Opus 5).</sub>

src/HypothesisTests.jl
```diff
@@ -39,7 +39,10 @@ check_same_length(x::AbstractVector, y::AbstractVector) = if length(x) != length
     throw(DimensionMismatch("Vectors must be the same length"))
 end
 
-function check_level(level::Float64)
+# `Real` rather than `Float64`: most `confint` methods here declare `level::Float64` and
+# get the conversion for free at the call boundary, but the rank tests accept any real, so
+# this has to as well. The comparisons hold for any of them.
+function check_level(level::Real)
```
andreasnoack: As mentioned elsewhere, I think we should probably just stick with `Float64` for now. Loosening that will require more dedicated work and testing than what we are after right now.

src/mann_whitney.jl
```diff
@@ -69,18 +88,22 @@ function mwustats(x::AbstractVector{<:Real}, y::AbstractVector{<:Real})
         # U = (nx + ny)*(nx + ny + 1)/2 - sum(ranks_y) - nx*(nx + 1)/2
         U = ny*(2 * nx + ny + 1)/2 - sum(@view(ranks[(begin + nx):end]))
     end
-    return (U, ranks, tieadj, nx, ny, median(x)-median(y))
+    R = promote_type(eltype(x), eltype(y))
```
andreasnoack: I think it's often better to restrict element types to be the same in the signature and then have separate promotion methods.

src/wilcoxon.jl
```diff
@@ -163,19 +186,38 @@ function StatsAPI.pvalue(x::ExactSignedRankTest; tail=:both)
...
-struct ApproximateSignedRankTest{T<:Real} <: HypothesisTest
+struct ApproximateSignedRankTest{T<:Real,S<:Real} <: HypothesisTest
     vals::Vector{T} # original values
     W::Float64              # test statistic: Wilcoxon rank-sum statistic
     ranks::Vector{Float64} # ranks without ties (zero values)
     signs::BitArray{1}      # signs of input of ranks
     tie_adjustment::Float64 # adjustment for ties
     n::Int                  # number of observations
-    median::Float64         # sample median
+    median::S               # sample median
```
andreasnoack: Same as above

src/mann_whitney.jl
yoninazarathy: Done in `e2e1bf4`. All four structs lose every type parameter, and `confint` now infers as `Tuple{Float64,Float64}`.

One correction to what your `Project.toml` comment listed, since this changes it rather than settling it: the parameterised spelling is broken differently, not restored. On master `ExactMannWhitneyUTest{Float64}` is concrete; on the branch as you reviewed it, it was a `UnionAll` that still matched instances; now it raises a `TypeError`, and the plain name is concrete instead. A downstream annotation carrying the parameter has to drop it.

Two things were tangled here and only one was worth keeping.

**Propagation, dropped.** `mwustats` no longer promotes: it returns the samples as received and the fields convert. Ranking and `median` still run at the input's precision, which matters, since converting first would let two distinct `Rational`s collide in `Float64` and be counted as tied. The cost is `.median` no longer tracking the sample type, the one thing master does today that this gives up.

**Reachability, kept.** On master `SignedRankTest` accepts only `Int` and `Float64`; `Float32`, `Float16`, `Rational` and `BigFloat` all raise a MethodError, because `signedrankstats` returns `median(x)` in the sample's own type against a constructor declaring `median::Float64`. Leaving that argument `::Real` keeps every one of them constructible at no type-parameter cost.

`check_level` is `Float64`-only again, converting in `ci_alpha` instead. It has to be `Float64(level)` rather than `float(level)`, since `float(0.9f0)` is still `Float32`. Either way that MethodError predates this branch: master's `calculate_ci` already declared `level::Real`.

<sub>Written by Claude Code (Opus 5).</sub>

src/mann_whitney.jl
yoninazarathy: Both gone in `154ed17`, along with the two signed rank siblings of the same shape. `resolve_rank_method` stays, being the one with two callers and real logic.

Inlining made visible something that had been a claim. The automatic rules now read as the literal expressions they replaced: `nx + ny <= 10 || (nx + ny <= 50 && tieadj == 0)` is character for character master's line, and the signed rank one restores master's `n_nonzero = length(ranks)` local along with its threshold. You had to check that equivalence by hand in your first pass; it is on the page now. `auto_mwu_method` also read its rule off `s.ties`, one hop from the tie adjustment it came from, and inlined it reads the adjustment directly.

Two smaller consequences. The `mwu` abbreviation is master's, from `mwustats` and `mwuenumerate`, and was borrowed for these two helpers; this branch no longer extends it. And I renamed `stats0` to `fields`, since it holds the eight values that go on to the struct and sat two lines from `stats`, which holds the four the `method` callable sees.

<sub>Written by Claude Code (Opus 5).</sub>

src/HypothesisTests.jl
yoninazarathy: Done in `e2e1bf4`: `check_level` is `Float64`-only again, and the conversion moved to `ci_alpha`, the one place the rank tests reach it from a `confint` declaring `level::Real`.

Two details from doing it. It has to be `Float64(level)` rather than the `float(level)` your earlier comment suggested, since `float(0.9f0)` is still `Float32`. And the MethodError that comment reported predates this branch either way: master's `calculate_ci` already declares `level::Real` and already calls `check_level`, so `confint(SignedRankTest(x); level = 0.9f0)` fails there today.

<sub>Written by Claude Code (Opus 5).</sub>

src/wilcoxon.jl
yoninazarathy: Done in `e2e1bf4`, and it is the mirror image of the [Mann-Whitney thread](https://github.com/JuliaStats/HypothesisTests.jl/pull/367#discussion_r3894215596): there master parametrised the median and this branch added the samples, here master parametrised `vals::Vector{T}` and this branch had added `median::S`. Both are `Float64` now and neither struct carries a parameter.

The one piece kept is that the seven-argument constructor still declares `median::Real`. That widening, rather than the field's type, is what makes `SignedRankTest(Float32.(x))` construct at all: `signedrankstats` returns `median(x)` in the sample's own type, against master's `median::Float64` argument, so `Float32`, `Float16`, `Rational` and `BigFloat` all raise a MethodError there today. The conversion happens at the field instead.

<sub>Written by Claude Code (Opus 5).</sub>

src/mann_whitney.jl
yoninazarathy: Gone too, in `154ed17`, with the reasoning in [the thread above](https://github.com/JuliaStats/HypothesisTests.jl/pull/367#discussion_r3894196626).

The gain specific to this one: it read its rule off `s.ties`, one hop from the tie adjustment that was derived from. Inlined it reads the adjustment directly, and is then character for character master's line, `nx + ny <= 10 || (nx + ny <= 50 && tieadj == 0)`.

<sub>Written by Claude Code (Opus 5).</sub>

src/mann_whitney.jl
yoninazarathy: Answered by deletion rather than by restructuring, in `e2e1bf4`. Going to `Float64` throughout removed the promotion, so there is no promotion method left to separate: `mwustats` now takes the two samples as they arrive and returns them unchanged, and the struct fields convert.

One nuance worth recording, since it is why the conversion sits at the field rather than at the signature. Ranking and `median` still run at the input's own precision. Converting first would let two distinct `Rational`s collide in `Float64` and be counted as tied, which moves the p-value.

<sub>Written by Claude Code (Opus 5).</sub>

src/rank_common.jl
```diff
@@ -0,0 +1,309 @@
...
+Below that it can still be narrower by one order statistic, on strongly unbalanced
+designs: at `level = 0.90` that happens for 131 of the 1444 Mann-Whitney shapes with
+`nx, ny` in `3:40`, concentrated at `nx = 3`. It is the exact rule that is conservative
+there rather than this one that is loose, and coverage stays at nominal rather than
+falling below it: at `(3, 12)`, `(3, 15)` and `(3, 21)` the normal route covers `0.902`,
+`0.901` and `0.900` against the exact route's `0.932`, `0.927` and `0.919`.
+"""
+function normal_ci_index(m::Integer, mu::Real, sigma::Real, alpha::Real)
+    z = quantile(Normal(), 1 - alpha / 2)
+    return clamp(ceil(Int, mu - z * sigma - 0.5) - 1, 0, div(m, 2))
```
andreasnoack:
```suggestion
    q = quantile(Normal(mu, sigma), 1 - alpha / 2)
    return clamp(ceil(Int, q - one(q) / 2) - 1, 0, div(m, 2))
```

andreasnoack (commented):
I think this is almost ready now

src/rank_common.jl
yoninazarathy: Taken in `1a0d013`, with the probability changed to `alpha / 2`.

`quantile(Normal(mu, sigma), 1 - alpha/2)` is `mu + z*sigma`, the upper endpoint, where this rule needs the lower one.

The rest stands, and `z` is gone from the code.

andreasnoack (approved):

Issue #372, opened 2026-08-19
SignTest one-sided confint returns the wrong endpoint, bounded by the null value


yoninazarathy:
`confint(::SignTest; tail = :left)` returns a lower bound, and `tail = :right` an upper one. Every other test in this package that takes a `tail` returns the opposite, and so does R.

```julia
julia> x = randn(MersenneTwister(4), 20) .+ 0.5;

julia> confint(SignTest(x); tail = :left)
(-0.114, 0.0)

julia> confint(OneSampleTTest(x); tail = :left)
(-Inf, 0.896)
```

This is the same defect as JuliaStats/HypothesisTests.jl#368, which is about the four rank tests. `SignTest` was not covered there. Of the nine `confint` methods here that take a `tail`, eight agree: the three t-tests, the three z-tests, `BinomialTest` and `FisherExactTest`. `SignTest` is the ninth.

## Why the other convention is the right one

The alternative that `tail = :left` selects is location *below* the null. The values compatible with that alternative form an upper bound, so that is what inverting the left-tailed test gives.

Read the other way round, `pvalue` and `confint` on the same object with the same `tail` currently describe opposite alternatives. `pvalue(t; tail = :left)` tests location below the null; `confint(t; tail = :left)` returns the acceptance region of the right-tailed test. A caller who picks a tail once and uses it for both gets a mismatched pair.

## The second half

`SignTest` also returns the null value as the other endpoint rather than an infinite one:

```julia
if tail == :left
    q = Int(quantile(Binomial(x.n, 0.5), 1-level))
    (x.data[q+1], x.median)
elseif tail == :right
    q = Int(quantile(Binomial(x.n, 0.5), 1-level))
    (x.median, x.data[end-q])
```

`x.median` there is the hypothesised median, not an estimate from the data, so a one-sided interval is reported as bounded by the very value being tested. Every other one-sided interval here is unbounded on the side it does not constrain.

## Fixing it

Both halves are breaking, since they change what an existing call returns. The rank test half is being fixed in the pull request stacked on JuliaStats/HypothesisTests.jl#367; this one is independent of that work and could go with it or separately.

Issue #371, opened 2026-08-19
Rank test intervals and estimates materialise the whole pairwise set to read three order statistics


yoninazarathy:
`confint` and `hodgeslehmann` on the rank tests build the whole pairwise set in order to read three numbers out of it. Selecting those three directly is exact, is `O(n)` in memory rather than `O(n^2)`, and changes no returned value. This is the concrete form of #7, and it would close #97.

Not to be confused with #361, which is a different cost in the same function and is fixed in the pull request stack below.

## What is needed, and what is built

Both procedures are read off a set of pairwise estimates: the `n(n+1)/2` Walsh averages `(xᵢ + xⱼ)/2` for the signed rank tests, the `nx·ny` cross-group differences `xᵢ - yⱼ` for the Mann-Whitney tests.

`confint` needs two order statistics of that set, `vals[k + 1]` and `vals[m - k]`, and `k` comes from the null distribution rather than from the data. `hodgeslehmann` needs the median, so one or two more. Three numbers.

To get them, the set is materialised and sorted. At the bound that is 100 million elements built and sorted to read three of them.

## Where it bites

Construction and `pvalue` never touch the set, so this is entirely a cost of asking for an estimate or an interval.

One-sample signed rank:

| n | `hodgeslehmann` | `confint` |
|---|---|---|
| 1000 | 9.8 MB | 5.8 MB |
| 4000 | 201 MB, 0.21 s | 138 MB, 0.09 s |
| 20000 | refuses | refuses |

Two-sample Mann-Whitney:

| sample | `hodgeslehmann` | `confint` |
|---|---|---|
| 2000 v 2000 | 77 MB | 61 MB |
| 5000 v 5000 | 635 MB, 0.43 s | 440 MB, 0.37 s |
| 10001 v 10001 | refuses | refuses |

`show` pays both, and pays them separately, because the estimate and the interval each build their own copy and neither knows about the other. At 5000 v 5000 that is 1076 MB to print one test; at n = 4000 signed rank, 353 MB.

`MAX_PAIRWISE_ESTIMATES` keeps this from exhausting the machine, and it works, but it is a bound on a cost that need not be paid at all.

## The fix

[...]

Issue #369, opened 2026-08-17
confint(::SignTest) one-sided intervals use the hypothesised median as an endpoint, and can be reversed


yoninazarathy:
`confint(::SignTest)` builds its one-sided intervals from the hypothesised median rather than from an infinite endpoint, so the interval moves when the hypothesis moves, and either endpoint can end up on the wrong side.

On v0.11.8:

```julia
julia> x = [1.2, 2.3, 3.1, 4.4, 2.8, 3.9, 5.1, 2.2, 3.3, 4.0];

julia> confint(SignTest(x, 0.0); tail = :left)     # reversed: lower endpoint above upper
(2.3, 0.0)

julia> confint(SignTest(x, 10.0); tail = :right)   # reversed the other way
(10.0, 4.0)

julia> confint(SignTest(x, 3.0); tail = :left)     # same data, different null, different interval
(2.3, 3.0)

julia> confint(SignTest(x))                        # two-sided, invariant here because nothing ties the null
(2.3, 4.0)
```

A confidence interval is a function of the data and the level, not of the value being tested, so all three one-sided results are wrong, and two of them are not intervals at all.

The cause is in `src/binomial.jl`:

```julia
if tail == :left
    q = Int(quantile(Binomial(x.n, 0.5), 1-level))
    (x.data[q+1], x.median)
elseif tail == :right
    q = Int(quantile(Binomial(x.n, 0.5), 1-level))
    (x.median, x.data[end-q])
```

`x.median` is the hypothesised median, defaulting to `0`.

## The two-sided interval is hypothesis-dependent too

The two-sided branch escapes the endpoint problem, but not the underlying one, because `n` is `count(x .!= median)`: observations equal to the hypothesised median are dropped from the count while the endpoints are still read off all `N` sorted values. So changing the hypothesis changes `q`, and with it the interval, even though the data has not moved:

```julia
julia> x = [0.0, 0.0, 1.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0, 10.0, 11.0, 12.0];

julia> confint(SignTest(x, 0.0)), confint(SignTest(x, 2.5))
((1.0, 9.0), (2.0, 8.0))
```

Two observations equal `0.0`, so the first call has `n = 13` and the second `n = 15`, giving `q = 3` against `q = 4`.

That also means neither classical construction is implemented. On the sample above, at `level = 0.95`:

[...]

Issue #368, opened 2026-08-17
One-sided confint uses opposite conventions for the rank tests and everything else


yoninazarathy:
Across the tests that accept a `tail` argument, `confint(test; tail = :left)` means two opposite things.

| tests | `tail = :left` returns | `tail = :right` returns |
|---|---|---|
| `OneSampleTTest`, `EqualVarianceTTest`, `UnequalVarianceTTest`, `OneSampleZTest`, `EqualVarianceZTest`, `UnequalVarianceZTest`, `BinomialTest`, `FisherExactTest` | upper bound | lower bound |
| `ExactSignedRankTest`, `ApproximateSignedRankTest` | **lower bound** | **upper bound** |

R returns an upper bound for `alternative = "less"`, so the rank tests here disagree with R as well as with the rest of this package. On v0.11.8, run against R on the same data, the numbers match and the labels do not:

```r
> x <- c(1.2, 2.3, 3.1, 4.4, 2.8, 3.9, 5.1, 2.2, 3.3, 4.0)
> t.test(x, mu = 3, alternative = "less")$conf.int
[1]     -Inf 3.903834
> t.test(x, mu = 3, alternative = "greater")$conf.int
[1] 2.556166      Inf
> wilcox.test(x - 3, alternative = "less", conf.int = TRUE)$conf.int
[1] -Inf 0.95
> wilcox.test(x - 3, alternative = "greater", conf.int = TRUE)$conf.int
[1] -0.45   Inf
> wilcox.test(x - 3, conf.int = TRUE)$conf.int
[1] -0.7  1.1
```

```julia
julia> x = [1.2, 2.3, 3.1, 4.4, 2.8, 3.9, 5.1, 2.2, 3.3, 4.0];

julia> confint(OneSampleTTest(x, 3.0); tail = :left)
(-Inf, 3.903833519443472)

julia> confint(OneSampleTTest(x, 3.0); tail = :right)
(2.556166480556527, Inf)

julia> confint(ExactSignedRankTest(x .- 3); tail = :left)
(-0.45000000000000007, Inf)

julia> confint(ExactSignedRankTest(x .- 3); tail = :right)
(-Inf, 0.9499999999999997)

julia> confint(ExactSignedRankTest(x .- 3))
(-0.7000000000000002, 1.0999999999999999)
```

[...]

PR #376, opened 2026-09-02
Signed rank and Mann-Whitney fixes: intervals, the two-sided p-value, and the documented nulls

### LINKED ISSUE #361: `calculate_ci` uses the exact null distribution for `Approximate*` tests, and selects the wrong order statistic from n = 73

yoninazarathy:
**Breaking.** The second of the three replacing #366, on top of #367, which is merged. [The docs PR](https://github.com/yoninazarathy/HypothesisTests.jl/pull/3) is stacked on this one and is documentation only; it will be opened here once this lands.

Closes #361, #362, #363 and #368. Further defects found while writing the specification appendix, and in review passes against R and against the code, are fixed here too. Every change here alters numbers this package returns today, or what the documentation claims about them. The one exception is the commit at the tip, which sets `version = "0.12.0"`: the bump was deferred out of #367 in `fa7f790`, so that nothing between the two could be registered by accident, and this is where the break completes, so this is where it lands.

## The interval now depends on which test you selected (#361)

`StatsAPI.confint` was defined identically on both signed rank types and called `signrankcdf` either way, so selecting `ApproximateSignedRankTest` did not produce an approximate interval: the package had no normal-approximation signed rank interval at all, and the tie correction the approximate type carries in `sigma` never reached its interval.

`Exact*` now inverts the exact null distribution; `Approximate*` inverts the normal approximation, with a continuity correction and the tie-corrected variance.

## Level selection is conservative rather than nearest (#361)

`argmin(abs.(l .- c))` could settle on an attainable coverage *below* nominal. Taking instead the largest index that still attains the level reproduces R's `qsignrank` rule on all 67 sample sizes `n = 5:71`, at 0.80, 0.90, 0.95 and 0.99. The old rule agreed with R on 30, 35, 32 and 27 of those respectively.

It is also a binary search rather than a scan over every index, so it runs `O(log n)` lattice recursions where the old rule ran `O(n^2)`.

## Zero differences are dropped for the interval too (#362)

`signedrankstats` drops them before ranking, so the statistic and the p-value describe the non-zero differences, but the interval was built from the Walsh averages of *all* of them: 210 against 120, on a 20-point sample with five zeros. The p-value and the interval described different samples.

## The reported estimate is the Hodges-Lehmann estimate (#363)

On all four types. The label said "pseudomedian"; the value was `median(x)`, or `median(x) - median(y)`. The Hodges-Lehmann estimate, the median of the pairwise estimates the interval is read off, is a different quantity from either: `0.62` against `0.56` on the tied nine-against-nine Mann-Whitney sample, and `10.1` against `9.675` on the fifteen-point signed rank one.

The Mann-Whitney half is a separate commit here rather than travelling with the interval that made it visible, so that #367 could stay strictly non-breaking.

`calculate_ci` goes with them; every caller now routes through the shared index rules. Across the whole PR the source grows by 55 lines net, over 230 added and 175 removed, most of the addition being comments and docstrings.

## One-sided intervals used the opposite convention to the rest of the package (#368)

`confint(test; tail = :left)` returned a *lower* bound for all four rank tests, and an upper bound for eight of the nine other `confint` methods here that take a `tail`: the three t-tests, the three z-tests, `BinomialTest` and `FisherExactTest`. R returns an upper bound under `alternative = "less"`, so the rank tests disagreed with R as well.

`SignTest` is the ninth, and is wrong in two separate ways, both out of scope here. Its `tail = :left` returns a lower bound as these four did, which is this same defect and is filed as JuliaStats/HypothesisTests.jl#372. Independently of orientation, it bounds the other side by the hypothesised median rather than leaving it infinite, so the interval moves when the null moves and either endpoint can end up on the wrong side, which is JuliaStats/HypothesisTests.jl#369.

The naming is the smaller half of it. On the same object, `pvalue` and `confint` given the same `tail` described opposite alternatives: `pvalue(t; tail = :left)` tests location below the null, whose acceptance region is an upper bound, while `confint(t; tail = :left)` returned the acceptance region of the *right*-tailed test. A caller who picked a tail once and used it for both got a mismatched pair from the rank tests and a matched one from everything else.

Each tail now keeps the endpoint that inverts the test of the same name. The endpoints keep their values; only which one is returned changes. On the sample from the issue:

| | before | after | R |
|:---|:---|:---|:---|
| `tail = :left` | `(-0.45, Inf)` | `(-Inf, 0.95)` | `(-Inf, 0.95)` |
| `tail = :right` | `(-Inf, 0.95)` | `(-0.45, Inf)` | `(-0.45, Inf)` |

The tests now assert the pairing directly rather than the orientation alone, shifting the null across each endpoint and checking which one-sided p-value crosses 0.05 there.

## The exact two-sided p-value could exceed 1

`pvalue(ExactSignedRankTest([1.0, 2.0, -3.0]))` returned `1.25`.

The untied branch doubled the tail on whichever side of the null mean `W` fell, without clipping. Where `n(n+1)/2` is even the null mean `n(n+1)/4` is itself attainable, and a statistic sitting on it has both tails equal to `(1 + P(W == mean))/2`, so the doubled tail comes to `1 + P(W == mean)`.

Sweeping every sign pattern of every untied sample up to `n = 12`, 220 of 8190 returned a p-value above 1, the largest `1.25` at `n = 3`. The affected sizes are `n ≡ 0, 3 (mod 4)`.

The three neighbouring exact routes already clip: the tied branch immediately below does `min(1, 2 * minimum(...))`, and both branches of the exact Mann-Whitney p-value do `min(2 * p, 1.0)`. This brings the fourth into line. Nothing away from the null mean moves, and no existing test changed.

## The documented null was the wrong one

All three signed rank docstrings gave the null as "the distribution of `x` has zero median". The test does not control that null. Its exact null distribution comes from the signs being independent of the magnitudes and each ± with probability 1/2, which symmetry about zero gives and a zero median alone does not.

At a nominal 0.05, over 20000 samples of `n = 20` drawn from median-zero but asymmetric distributions, the two-sided test rejects at 0.101 for `Exponential(1) - ln 2` and 0.110 for `LogNormal(0,1) - 1`, against 0.0498 for a symmetric `Normal(0,1)`. The docstrings now say symmetry, and name the pseudomedian as the location estimated, which is what `hodgeslehmann` and `population_param_of_interest` already report. The alternative is stated as the negation of the null rather than as symmetry about some other location, which excluded the asymmetric alternatives the test also has power against, and the automatic selection rule is given as one threshold per tie pattern rather than as a single sentence carrying both.

## The Mann-Whitney null was stated as the wrong hypothesis

All three Mann-Whitney docstrings gave the null as `P(x > y) = P(y > x)`. The test does not control that hypothesis. Its exact null distribution comes from the pooled observations being exchangeable, which `F_x = F_y` gives and that weaker equality does not: it leaves the two distributions free to differ in spread.

Simulated at a nominal 0.05, `X ~ Normal(0, 1)` against `Y ~ Normal(0, 3)`, three times the spread, so that `P(X > Y) = 1/2` holds exactly by symmetry, 20000 replicates each:

| `(nx, ny)` | exact | approximate |
|---|---|---|
| `(30, 10)` | 0.1295 | 0.1236 |
| `(20, 20)` | 0.0706 | 0.0706 |
| `(10, 30)` | 0.0162 | 0.0144 |
| `(5, 45)` | 0.0013 | 0.0013 |

Against 0.0493 and 0.0466 for `(10, 30)` and `(5, 45)` when `F_x = F_y`. The error runs in both directions, set by which sample is larger, so it cannot be waved through as conservative.

Both `Approximate*` docstrings also labelled their **variance** as σ; both now say σ². And both presented μ = n(n+1)/4 (or n_x n_y/2) while `show` prints `normal approximation (μ, σ)` with the statistic *centred* at that mean, -37.5 where the docstring's μ would be 60 on the worked two-sample example; the docstrings now call the null mean μ₀ and say the reported pair is (W − μ₀, σ), which is also how the specification writes it.

## `show` reported a count nothing was computed from

For a 20-point sample with five zero differences it printed `number of observations: 20`, while the statistic, p-value, estimate and interval were all computed from the other 15. Both counts are now printed:

```
number of observations:         20
non-zero observations:          15
```

## An unattainable confidence level passed silently

The widest interval this construction admits is `(V_(1), V_(m))`, whose coverage is `1 - 2 P(W <= 0)`, or `1 - 2^(1-n)` for an untied signed rank sample: `0.75`, `0.875` and `0.9375` at `n = 3, 4, 5`. So `level = 0.95` is unattainable below `n = 6` and `level = 0.99` below `n = 8`, and `confint` returned the widest interval it had without saying it fell short. `show` printed a "95% confidence interval" that was not one.

Both routes still return that interval, since it is the best available, but now say so. The exact route names the coverage that is attainable. (R does not warn here: at `n = 5` and `level = 0.95` it returns the same interval silently, its `conf.level` attribute still claiming `0.95`; it renames the attribute and warns only once the shortfall exceeds `alpha/2`, as at 2 v 2, where the attribute becomes 2/3.) The approximate route cannot name a coverage: its rule asks for an order statistic outside the pairwise estimates on small samples, and that is not the same thing, since at `n = 8` and `level = 0.99` it wants `k = -1` while the exact route reaches `0.9922`. So it reports what happened and points at `method = :exact`.

Two refinements to the same warnings, from review. The warning no longer fires when `P(W <= 0)` equals `alpha/2` exactly, where the widest interval attains the request (at `n = 5` that is `level = 0.9375`, and R agrees with the returned interval there). And a one-sided call is warned about in its own terms, the level actually asked for and the one-sided attainable coverage `1 - P(W <= 0)`; it previously reported the two-sided `2*level - 1` as the request and a two-sided coverage as what was attainable.

## Two Int64 overflows, found in review

The approximate signed rank variance formed `nz * (nz + 1) * (2 * nz + 1)` in `Int64`, which wraps from `nz = 2^21`; at exactly `2^21` non-zero observations the product wraps to a *positive* wrong value, so `sigma` came out three orders of magnitude too small, over-rejecting massively with no error raised. `tiedrank_adj` had the same problem in `t^3 - t` for a single tie group of `2^21` or more equal values. Both products now form in `Int128`: below the old wrap point the results are bit-identical, and a test pins the `2^21` case against `BigInt` arithmetic. Both overflows predate this stack.

## The signed rank constructors rejected everything but `Vector`

They declared `AbstractVector` but passed the input straight to struct fields typed `Vector`, so `SignedRankTest(1:5)`, ranges, and views threw a `MethodError`. They convert on the way in now, as the Mann-Whitney constructors always have. Also pre-existing on `master`.

## Two `show` labels said the wrong thing

The signed rank tests labelled `W⁺` as `Wilcoxon rank-sum statistic`, which is the two-sample test's name; the line now reads `Wilcoxon signed rank statistic`. And the Mann-Whitney tests reported their parameter of interest as `Location parameter (pseudomedian)`, the one-sample estimand; under the shift model the two-sample parameter is the shift, estimated by the median of the cross-group differences, so they now report `Location shift`, as R's `wilcox.test` calls it. Printed output only, and both labels predate this stack.

---

## Migration

*Written to be pasted into the release notes of 0.12.0; there is no `CHANGELOG.md`.*

> Confidence intervals and point estimates from `SignedRankTest`, `ExactSignedRankTest` and `ApproximateSignedRankTest` change value, and one-sided intervals from all four rank tests change which endpoint they return.
>
> - The interval returned by `ApproximateSignedRankTest` is now the normal-approximation interval rather than the exact one. `method = :exact`, or `ExactSignedRankTest`, recovers the previous construction.
> - Intervals are now conservative: the attained coverage is the smallest available that is at or above the requested level, where previously it was the closest available, which could be below it. Intervals are therefore never narrower than before, and are wider on roughly half of sample sizes.
> - Samples containing zero differences now yield an interval and an estimate computed from the non-zero differences only, matching the p-value, which already ignored them.
> - The reported point estimate, for the Mann-Whitney tests as well as the signed rank ones, is the Hodges-Lehmann estimate, available directly as `hodgeslehmann`. It was the sample median, or the difference of sample medians. Exact symmetry makes the two agree, but they can also agree by coincidence, as on `[1, 3, 3, 8]`, where both are `3`.
> - `confint(test; tail = :left)` on any of the four rank tests now returns an upper bound, `(-Inf, hi)`, and `tail = :right` a lower bound, `(lo, Inf)`. They were the other way round. This matches the other eight tests in this package that accept a `tail`, and R's `alternative = "less"` and `"greater"`. Code that passed a tail to a rank `confint` and used the finite endpoint needs the other tail; code that read the endpoint positionally needs to swap indices.
>
> - Two-sided p-values from `ExactSignedRankTest` on untied samples are now capped at 1. They could previously exceed it, by up to 0.25, when the statistic landed exactly on its null mean.
>
> - The signed rank tests print an extra `non-zero observations` line, and the documented null hypotheses change wording without any change in what is computed.
>
> - `confint` on a sample too small for the requested level now warns, naming the attainable coverage. The widest-interval return there comes from the conservative-rule change above; the warning itself changes no value.
>
> - The signed rank `show` statistic line reads `Wilcoxon signed rank statistic`, and the Mann-Whitney parameter of interest reads `Location shift`. Printed output only.
>
> - `SignedRankTest` and both concrete signed rank constructors accept any `AbstractVector`, and the approximate signed rank `sigma` is correct past two million non-zero observations, where it silently wrapped `Int64` before.
>
> There is no opt-out for the corrections from the p-value cap down; they are corrections rather than choices.

## The interval bound, re-measured against the bisection

`MAX_EXACT_CI_ESTIMATES = 25_000` was sized in #367 against the `m/2`-recursion endpoint scan that #361 deletes here, about 9 s at the bound, a signed rank sample of 223. Re-measured against the bisection, the cost turns out to have moved routes: the expensive lattice recursion is now the two-sample `wilcoxcdf`, not the one-sample `signrankcdf`. So the recalibration splits by route rather than moving one number.

The **signed rank** exact interval sheds the time bound entirely: running to the memory bound, a sample of 1413, costs about 6 s in total, under the old ceiling, and `n = 300` answers in under a tenth of a second, where the scan this replaced took 39 s and `master` as it stands, since #367, refuses. The refusal band #367 documents, samples of 224 to 1413, closes here.

The **Mann-Whitney** exact interval keeps the bound at the same wall-clock ceiling re-measured: `90_000`, which admits a 300-against-300 design at about 11 s and refuses 400 against 400 before the differences are formed (35 s if allowed). `method = :approximate` stays the unbounded escape. The constant's docstring, the manual's bounds prose, the boundary assertions, and the routing tables in the docs PR move in step.

## What this deliberately leaves open

JuliaStats/HypothesisTests.jl#371. The estimate and the interval are read off a set of pairwise estimates, and that set is materialised in full to take three order statistics out of it. This branch does not change that, and in one respect makes it more visible: #363 makes the reported estimate the Hodges-Lehmann one, so `show` now builds the set twice, once for the estimate and once for the interval, where before it built it once.

The direction of travel is still strongly downwards. The scan #361 replaces dominated everything else: `confint(SignedRankTest(randn(400)))` took 164 s before this branch and takes 0.0012 s after it. What is left is memory rather than time, and it is bounded rather than unbounded: `MAX_PAIRWISE_ESTIMATES` refuses instead of thrashing, at `n = 1413`, where the set is about 8 MB and sorting it costs as much again. Selecting the three order statistics rather than sorting for them would need neither.

Selecting the three order statistics directly is exact and changes no returned value, so it is a self-contained follow-up rather than part of a breaking change, and it is worth more after this branch than before it: until #361 lands, the signed rank interval is dominated by that scan and not by the allocation. #371 carries the measurements and a checked proof of concept.

## Blast radius

Indicative only, since GitHub code search does not index everything and misses private code, but across the public corpus there are **no** hits for `confint` used with `SignedRankTest` or `MannWhitneyUTest`, against 12 and 9 repos referencing the `Exact*` and `Approximate*` types. Which is about what you would expect for a function that returned the exact interval regardless of which test you asked it for.

## Tests

Wilcoxon 412 (was 249 on `master` after #367, 62 before it); Mann-Whitney 308 (was 271 after #367, 77 before it).

The p-value sweep is one assertion over all 8190 sign patterns up to `n = 12`, reporting the offending `(n, bits)` pairs rather than a count of failures.

The six `PENDING` assertions #367 left behind become their fixed counterparts, which is the whole of the behaviour change as seen from the test suite: the approximate interval becomes `(3.05, 15.5)` against R's `(3.05004, 15.50001)`, the zeros sample `(-0.25, 0.75)` against R's `(-0.24999, 0.75005)`, the reported estimate becomes the Hodges-Lehmann one, and `method` reaches the interval rather than only the returned type.

New invariants across sizes and both routes: the estimate lies inside its own interval, the interval cannot narrow as the level rises, and each one-sided bound is the corresponding two-sided endpoint with the far side open and on the side its own tail's alternative allows.

A review pass added the pins the suite could not previously distinguish, each against R: a 3 v 6 sample on which the exact and approximate Mann-Whitney interval rules pick different order statistics, so a silent fallback from one rule to the other now fails; the continuity-corrected one-sided approximate Mann-Whitney p-values; the exact tied one-sided intervals of both tests and the tied approximate signed rank interval, the one place the tie-corrected `sigma` moves an endpoint; a 12 v 8 untied sample through the `nx > ny` swap on the StatsFuns route; the automatic rule's boundaries, `n = 50/51` pinned against R on both sides and the tied 15/16 and 10/11 asserted by type; levels at the edges, where 0.5 and 1.0 throw and 0.51 and a one-sided 0.99 match R; and all-equal samples returning `p = 1` on both routes.

Every route this PR touches is additionally pinned to a reference value rather than to a structural property alone: exact and approximate one-sided p-values and intervals against `wilcox.test` with `alternative` set, the tied one-sided p-values of both tests against `exactRankTests::wilcox.exact` to nine digits, and the degenerate small-sample intervals against what R returns there, `(-1.9, 2.3)` at `n = 5` and `(-3, -1)` for 2 v 2, where R's achieved `conf.level` attribute, 2/3, is the same attainable coverage the warning here names. One convention difference is recorded in a test comment rather than papered over: `wilcox.exact` sums the far tail for its two-sided value where this package doubles the smaller one, so the two-sided values agree one-sample, where the tied conditional distribution is symmetric, and differ two-sample, where it is not (`0.0120034764` against `0.0118890398` on the tied 10 v 12 sample); the one-sided values, which no convention touches, agree exactly.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,10 +1,6 @@
 name = "HypothesisTests"
 uuid = "09f84164-cd44-5f33-b23f-e6b0d136a0d5"
-# Deliberately not 0.12.0 yet. This branch is breaking and 0.12.0 is what it will
-# be, but the bump lands with the stacked PR, so nothing between the two can be
-# registered by accident: 0.11.8 is already in General against a different hash,
-# so the registrator refuses it.
-version = "0.11.8"
+version = "0.12.0"
 
 [deps]
 Combinatorics = "861a8166-3701-5b0c-9a16-15d98fcdc6aa"
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -6,7 +6,7 @@ HypothesisTests = "09f84164-cd44-5f33-b23f-e6b0d136a0d5"
 [compat]
 Documenter = "1"
 DocumenterCitations = "1"
-HypothesisTests = "0.11"
+HypothesisTests = "0.12"
 
 [sources]
 HypothesisTests = {path = ".."}
--- a/docs/src/nonparametric.md
+++ b/docs/src/nonparametric.md
@@ -86,7 +86,7 @@ hodgeslehmann
 The Hodges-Lehmann estimate is the point estimate the interval brackets, and is what all
 four types report as their parameter of interest. It is generally not the sample median:
 exact symmetry makes the two agree, but they can also agree by coincidence, as on
-`[0, 2, 2, 7]`, where both are `2`. Agreement is therefore no evidence of symmetry.
+`[1, 3, 3, 8]`, where both are `3`. Agreement is therefore no evidence of symmetry.
 
 Which order statistics `confint` returns depends on the test: the `Exact*` types invert
 the exact null distribution, conservatively, and the `Approximate*` types invert the
@@ -94,6 +94,12 @@ normal approximation with a continuity correction and a tie-corrected variance. 
 `Exact*` interval agrees with R's `wilcox.test` at `exact = TRUE` digit for digit; the
 `Approximate*` interval agrees with `exact = FALSE` to within one order statistic, since
 R solves for its endpoints numerically rather than reading them off the pairwise estimates.
+
+A one-sided interval keeps the endpoint that inverts the test of the same name:
+`tail = :left`, whose alternative is location below the null, gives an upper bound, and
+`tail = :right` a lower one. That is the convention of every other test here that takes a
+`tail`, and of R's `alternative = "less"` and `"greater"`. These four tests returned the
+other endpoint before this change (#368).
 
 Three named bounds apply, and past any of them the tests raise rather than run unbounded.
 The first bounds the tied-data enumeration a p-value runs, and `method = :approximate` is
--- a/docs/src/nonparametric.md
+++ b/docs/src/nonparametric.md
@@ -93,7 +93,7 @@ the exact null distribution, conservatively, and the `Approximate*` types invert
 normal approximation with a continuity correction and a tie-corrected variance. The
 `Exact*` interval agrees with R's `wilcox.test` at `exact = TRUE` digit for digit; the
 `Approximate*` interval agrees with `exact = FALSE` to within one order statistic, since
-R solves for its endpoints numerically rather than reading them off the contrast set.
+R solves for its endpoints numerically rather than reading them off the pairwise estimates.
 
 Three named bounds apply, and past any of them the tests raise rather than run unbounded.
 The first bounds the tied-data enumeration a p-value runs, and `method = :approximate` is
--- a/docs/src/nonparametric.md
+++ b/docs/src/nonparametric.md
@@ -90,8 +90,9 @@ exact symmetry makes the two agree, but they can also agree by coincidence, as o
 
 Which order statistics `confint` returns depends on the test: the `Exact*` types invert
 the exact null distribution, conservatively, and the `Approximate*` types invert the
-normal approximation with a continuity correction and a tie-corrected variance. The
-`Exact*` interval agrees with R's `wilcox.test` at `exact = TRUE` digit for digit; the
+normal approximation with a continuity correction and a tie-corrected variance. On untied
+data the `Exact*` interval agrees with R's `wilcox.test` at `exact = TRUE` digit for digit,
+and under ties R declines an exact interval even when asked for one; the
 `Approximate*` interval agrees with `exact = FALSE` to within one order statistic, since
 R solves for its endpoints numerically rather than reading them off the pairwise estimates.
 
--- a/docs/src/nonparametric.md
+++ b/docs/src/nonparametric.md
@@ -107,8 +107,9 @@ The first bounds the tied-data enumeration a p-value runs, and `method = :approx
 the way past it. The second bounds the set of pairwise estimates, which `confint` and
 `hodgeslehmann` form whichever route the p-value took, so `method` is no help there; it is a
 bound on memory, which is all the approximate interval spends. The third bounds the exact
-interval, which spends time as well, a lattice recursion for every candidate endpoint it
-considers, and so is bounded far below the second.
+Mann-Whitney interval, which spends time as well, a two-sample lattice recursion per
+bisection step, and so is bounded below the second; the exact signed rank interval bisects
+a cheaper recursion and meets only the second.
 
 All three raise `ComputationTooLarge`, which is its own type rather than an `ArgumentError`
 so that `show` can drop a refused interval line without also hiding real errors.
--- a/docs/src/nonparametric.md
+++ b/docs/src/nonparametric.md
@@ -83,10 +83,17 @@ and `hodgeslehmann` returns its median.
 hodgeslehmann
 ```
 
-The Hodges-Lehmann estimate is the point estimate the interval brackets. It is generally
-not the sample median: exact symmetry makes the two agree, but they can also agree by
-coincidence, as on `[0, 2, 2, 7]`, where both are `2`. Agreement is therefore no evidence
-of symmetry.
+The Hodges-Lehmann estimate is the point estimate the interval brackets, and is what all
+four types report as their parameter of interest. It is generally not the sample median:
+exact symmetry makes the two agree, but they can also agree by coincidence, as on
+`[0, 2, 2, 7]`, where both are `2`. Agreement is therefore no evidence of symmetry.
+
+Which order statistics `confint` returns depends on the test: the `Exact*` types invert
+the exact null distribution, conservatively, and the `Approximate*` types invert the
+normal approximation with a continuity correction and a tie-corrected variance. The
+`Exact*` interval agrees with R's `wilcox.test` at `exact = TRUE` digit for digit; the
+`Approximate*` interval agrees with `exact = FALSE` to within one order statistic, since
+R solves for its endpoints numerically rather than reading them off the contrast set.
 
 Three named bounds apply, and past any of them the tests raise rather than run unbounded.
 The first bounds the tied-data enumeration a p-value runs, and `method = :approximate` is
--- a/src/HypothesisTests.jl
+++ b/src/HypothesisTests.jl
@@ -102,7 +102,7 @@ function Base.show(_io::IO, test::T) where T<:HypothesisTest
     params = show_or_nothing(population_param_of_interest, test)
     println(io, "Population details:")
     if params === nothing
-        println(io, "    parameter of interest:   not computed for a sample this size")
+        println(io, "    parameter of interest:   not computed (over MAX_PAIRWISE_ESTIMATES)")
     else
         (param_name, param_under_h0, param_estimate) = params
         println(io, "    parameter of interest:   $param_name")
--- a/src/HypothesisTests.jl
+++ b/src/HypothesisTests.jl
@@ -69,26 +69,27 @@ end
 
 Base.showerror(io::IO, err::ComputationTooLarge) = print(io, "ComputationTooLarge: ", err.msg)
 
-# The interval `show` prints beside the point estimate, or `nothing` where there is none
-# to print. A test whose `confint` refuses the sample it was given still has a name, a
-# statistic and a p-value worth displaying, so the offending line is dropped rather than
-# the whole display: the rank tests bound the pairwise estimates their intervals are read
-# off (see `MAX_PAIRWISE_ESTIMATES`), and past that bound a large `MannWhitneyUTest` would
-# otherwise be impossible to look at.
+# `show` asks a test for two things it may not be able to give for the sample it was
+# handed: its point estimate and its interval. For the rank tests both are read off a
+# pairwise set bounded by `MAX_PAIRWISE_ESTIMATES`, so past that bound each raises. A
+# test that cannot afford one still has a name, a statistic and a p-value worth seeing,
+# so what cannot be computed is dropped rather than the whole display.
 #
-# Only `ComputationTooLarge` is caught. `confint` is generic over `HypothesisTest`, so
-# catching `ArgumentError` here would also swallow a genuine argument bug in any
-# downstream package's `confint` method and print a test with no interval line instead of
-# raising, which is a hard thing to debug.
-function show_confint(test::HypothesisTest)
-    applicable(confint, test) || return nothing
+# Only `ComputationTooLarge` is caught. These functions are generic over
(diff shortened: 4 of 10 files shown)


yoninazarathy:
Split into standalone PRs as discussed, so each fix can be reviewed and merged alone, in any order:

- #377 clips the exact signed rank two-sided p-value at 1
- #378 fixes the two Int64 overflows (sigma, and tiedrank_adj, which also serves Kruskal-Wallis)
- #379 accepts any AbstractVector in the signed rank constructors
- #380 fixes the two show labels and adds the non-zero observation count
- #381 states the nulls the tests actually control (docstrings only)
- #382 flips the one-sided intervals to the package's own convention (#368)
- #383 reports the Hodges-Lehmann estimate on all four types (#363)
- #384 warns on unattainable confidence levels and hardens the index rules
- #385 is test-only: pins of existing behaviour against R

The one thing that would not disentangle is #361 + #362: they are mutually exclusive edits to the same function (one fixes zero handling inside `calculate_ci`, the other deletes `calculate_ci`), so they stay one PR, opened once #382 and #384 land, which its tests build on. The bound recalibration and the 0.12.0 bump follow it, as before.

Completeness was checked mechanically: the union of the split diffs equals this PR's diff line for line, less the recalibration and the bump, which trail. Each branch passes the full suite on its own against `master`. Closing in favour of the split.

PR #378, opened 2026-09-02
Fix two Int64 overflows behind the rank test variances

yoninazarathy:
One of the standalone pieces of #376. Independent of the others; merge in any order.

The approximate signed rank variance formed `nz*(nz+1)*(2nz+1)` in `Int64`, which wraps from `nz = 2^21`, making `sigma` three orders of magnitude too small with no error raised. `tiedrank_adj` had the same problem in `t^3 - t` for a single tie group of `2^21` equal values, and that helper also serves Kruskal-Wallis. Both products now form in `Int128`: bit-identical below the old wrap point, and the `2^21` case is pinned against `BigInt`. Both predate the rank test work.

### DIFF
--- a/src/common.jl
+++ b/src/common.jl
@@ -45,8 +45,8 @@ function tiedrank_adj!(ord::AbstractVector, v::AbstractArray)
         if j > i
             t = j - i + 1
             m = sum(i:j) / t
-            # in Int128: t^3 wraps Int64 for a tie group of 2^21 or more equal values
-            tieadj += Int128(t)^3 - t
+            # avoid integer overflow
+            tieadj += (float(t) - 1)^2 * t
             for k = i:j
                 ord[place[k]] = m
             end
--- a/src/common.jl
+++ b/src/common.jl
@@ -46,7 +46,7 @@ function tiedrank_adj!(ord::AbstractVector, v::AbstractArray)
             t = j - i + 1
             m = sum(i:j) / t
             # avoid integer overflow
-            tieadj += (float(t) - 1)^2 * t
+            tieadj += (float(t)^2 - 1) * t
             for k = i:j
                 ord[place[k]] = m
             end
--- a/src/common.jl
+++ b/src/common.jl
@@ -45,7 +45,8 @@ function tiedrank_adj!(ord::AbstractVector, v::AbstractArray)
         if j > i
             t = j - i + 1
             m = sum(i:j) / t
-            tieadj += t^3 - t
+            # in Int128: t^3 wraps Int64 for a tie group of 2^21 or more equal values
+            tieadj += Int128(t)^3 - t
             for k = i:j
                 ord[place[k]] = m
             end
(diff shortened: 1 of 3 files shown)


src/common.jl
```diff
@@ -45,7 +45,8 @@ function tiedrank_adj!(ord::AbstractVector, v::AbstractArray)
         if j > i
             t = j - i + 1
             m = sum(i:j) / t
-            tieadj += t^3 - t
+            # in Int128: t^3 wraps Int64 for a tie group of 2^21 or more equal values
+            tieadj += Int128(t)^3 - t
```
andreasnoack:
```suggestion
            # avoid integer overflow
            tieadj += (float(t) - 1)^2 * t
```

src/wilcoxon.jl
```diff
@@ -246,7 +246,8 @@ Implements: [`pvalue`](@ref), [`confint`](@ref), [`hodgeslehmann`](@ref)
 function ApproximateSignedRankTest(x::Vector, W::Float64, ranks::Vector{T}, signs::BitArray{1}, tie_adjustment::Float64, n::Int, median::Real) where T<:Real
     nz = length(ranks) # num non-zeros
     mu = W - nz * (nz + 1)/4
-    std = sqrt(nz * (nz + 1) * (2 * nz + 1) / 24 - tie_adjustment / 48)
+    # the triple product in Int128: in Int64 it wraps from nz = 2^21, silently
+    std = sqrt(Int128(nz) * (nz + 1) * (2 * nz + 1) / 24 - tie_adjustment / 48)
```
andreasnoack:
```suggestion
    # avoid integer overflow
    std = sqrt(float(nz) * (nz + 1) * (2 * nz + 1) / 24 - tie_adjustment / 48)
```

src/common.jl
```diff
@@ -45,7 +45,8 @@ function tiedrank_adj!(ord::AbstractVector, v::AbstractArray)
         if j > i
             t = j - i + 1
             m = sum(i:j) / t
-            tieadj += t^3 - t
+            # avoid integer overflow
+            tieadj += (float(t) - 1)^2 * t
```
andreasnoack:
```suggestion
            tieadj += (float(t)^2 - 1) * t
```

andreasnoack:
Also a conflict here now

yoninazarathy:
Rebased onto master, so the conflict is gone, and your rewrite is in.

I checked it against exact arithmetic rather than taking the algebra on trust, and it is better than a wash: `(float(t)^2 - 1) * t` is **bit-identical** to `big(t)^3 - t` at `t = 2^20`, `2^21` and `2^21 + 1`, with a worst relative error of `6.4e-17` over `t` up to `10^7`; the sigma product is within `4.9e-18` over `nz` up to five million. And `tieadj` was already initialised to `0.0`, so the `Int128` intermediate was being converted to `Float64` at every step anyway. Nothing was bought by the wider integer type. I have put those numbers in the test comment so the next reader does not have to re-derive them.

CI had never run on your three commits, since the branch was conflicting; it runs on this head.

andreasnoack (approved):

PR #380, opened 2026-09-02
Report both observation counts, and name the signed rank statistic in show

yoninazarathy:
One of the standalone pieces of #376. Independent of the others; merge in any order. Printed output only.

For a 20-point sample with five zero differences, `show` printed `number of observations: 20` while the statistic, p-value and ranks were computed from the other 15; both counts are now printed. And the statistic line labelled `W⁺` as `Wilcoxon rank-sum statistic`, the two-sample test's name; it now reads `Wilcoxon signed rank statistic`. Both predate the rank test work.

### DIFF
--- a/src/wilcoxon.jl
+++ b/src/wilcoxon.jl
@@ -94,8 +94,8 @@ end
 
 struct ExactSignedRankTest <: HypothesisTest
     vals::Vector{Float64}   # original values
-    W::Float64              # test statistic: Wilcoxon rank-sum statistic
-    ranks::Vector{Float64}           # ranks without ties (zero values)
+    W::Float64              # test statistic: the signed rank statistic W+
+    ranks::Vector{Float64}           # midranks of |d| over the non-zero observations
     signs::BitArray{1}      # signs of input of ranks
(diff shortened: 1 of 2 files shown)


andreasnoack (commented):
Needs a test. Ideally against a full output string with `repr(MIME"text/plain", ...)`

yoninazarathy:
Added, in `repr(MIME"text/plain"(), ...)` form as you suggested, matching the snapshot convention the Mann-Whitney tests already use. Both types are pinned; the sample carries zeros so the two counts differ, which is the distinction the fix is about (`n = 10` against the seven non-zero differences the statistic, the p-value and the ranks actually describe). The approximate snapshot covers the one line the exact one does not, the centred statistic beside sigma, and its column alignment.

I checked the snapshot has teeth rather than assuming it: with the fix reverted, the two lines it catches are

```
    number of observations:      10          # no companion line
    Wilcoxon rank-sum statistic: 20.0        # the two-sample test's name
```

This also replaces the bare `show(IOBuffer(), ...)` in the basic testset, which asserted nothing. Rebased onto master.

andreasnoack (approved):

PR #381, opened 2026-09-02
State the nulls the rank tests control, and label the variances as variances

yoninazarathy:
One of the standalone pieces of #376. Independent of the others; merge in any order. Docstrings only; nothing computed changes.

The signed rank docstrings gave the null as a zero median, which the test does not control: over 20000 samples of `n = 20` from median-zero but asymmetric laws, a nominal 0.05 test rejects at 0.101 (`Exponential(1) - ln 2`) and 0.110 (`LogNormal(0,1) - 1`). They now say symmetry. The Mann-Whitney docstrings gave the null as `P(x > y) = P(y > x)`, which leaves the spreads free: for `Normal(0,1)` against `Normal(0,3)`, where that equality holds exactly, rejection is about 13% at `(30, 10)` and 1.7% at `(10, 30)`, in both directions, so it cannot be excused as conservative. They now say equality of distributions. Both `Approximate*` docstrings labelled their variance σ (now σ²) and presented a μ that `show` never prints; they now state the reported pair `(statistic − μ₀, σ)`.

### DIFF
--- a/src/mann_whitney.jl
+++ b/src/mann_whitney.jl
@@ -30,18 +30,26 @@ export MannWhitneyUTest, ExactMannWhitneyUTest, ApproximateMannWhitneyUTest
 """
     MannWhitneyUTest(x::AbstractVector{<:Real}, y::AbstractVector{<:Real}; method = :auto)
 
-Perform a Mann-Whitney U test of the null hypothesis that the probability that an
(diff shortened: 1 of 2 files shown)


andreasnoack (approved):

andreasnoack:
~Also a conflict here now~

PR #384, opened 2026-09-02
Warn when a requested confidence level is out of reach for the sample

yoninazarathy:
One of the standalone pieces of #376. Independent of the others; merge in any order.

An unattainable confidence level passed silently: at two observations against two, a "95% confidence interval" with coverage 2/3 was printed as though it met the request. Both index rules now warn when they fall short, naming the request and, on the exact rule, the attainable coverage; nothing fires where `P(W <= 0)` equals `alpha/2` exactly, where the widest interval attains the request. R warns only once the shortfall exceeds `alpha/2` (its `conf.level` attribute claims the requested level until then), which the tests record. The approximate rule warns for its own weaker reason, an endpoint outside the pairwise estimates, and does not claim unattainability, which does not follow from it.

A `tail` keyword shapes the warning for one-sided callers, and both rules cap the index at `div(m - 1, 2)`, so an inverted interval at even `m` is unrepresentable rather than argued away. Only the Mann-Whitney intervals reach these rules today; the signed rank route gains the same warnings when #361 routes it through them.

### DIFF
--- a/test/mann_whitney.jl
+++ b/test/mann_whitney.jl
@@ -425,6 +425,7 @@ end
     # the normal index rule cannot tell that mean from the centred statistic, and the
     # reflection holds either way; on a clearly shifted sample it does not, which is
     # what makes this testset detect the confusion its `confint` comment warns about.
+    warned = 0
     for (x, y) in (([1.0, 4, 6, 9, 11, 14], [2.0, 5, 7, 8, 12, 13]),   # no ties
                    ([1.0, 2, 2, 5, 5, 5, 8], [2.0, 3, 5, 5, 9, 9]),    # tied within and across
                    ([1.0, 2, 3], [10.0, 11, 12, 13]),                  # disjoint, unequal sizes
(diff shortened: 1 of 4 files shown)


test/mann_whitney.jl
```diff
@@ -400,8 +400,13 @@ end
             @test pvalue(a; tail=:left) == pvalue(b; tail=:right)
             # the two-sided interval reflects about zero, ends exchanged
             for level in (0.90, 0.95, 0.99)
-                lo_a, hi_a = confint(a; level=level)
-                lo_b, hi_b = confint(b; level=level)
+                # the smaller designs here cannot attain the higher levels, so both
+                # routes warn. That behaviour is pinned in "Unattainable levels warn"
+                # in test/wilcoxon.jl; here it is noise over the reflection identity.
+                (lo_a, hi_a), (lo_b, hi_b) =
+                    Base.CoreLogging.with_logger(Base.CoreLogging.NullLogger()) do
```
andreasnoack: Isn't it possible to test the logged message?

yoninazarathy:
It is, and you were right that silencing them was the wrong answer. Both places now assert.

The reflection sweep captures instead of discarding: every record has to be a warning naming the level actually asked for, and the count is pinned at eight, which is the 3 v 4 design on both routes at 0.95 and 0.99, twice over since each iteration confints the design and its swap. So an unexpected message fails there now rather than scrolling past, and if the warnings stopped firing altogether the count would catch it.

The warning testset checks the payload rather than a regex on the text:

- the two-sided case names `requested = 0.95` and `attainable = 1 - 2/6`, which is exactly R's achieved `conf.level` attribute of 2/3 on that 2 v 2 design;
- the one-sided case names `requested = 0.95` and `attainable = 1 - 1/6`, since a one-sided caller asked for `level` rather than the two-sided `2 level - 1` the index rule works in, and only one tail can overshoot;
- the approximate rule's warning names the request and deliberately carries **no** `attainable` key, since its trigger, an endpoint outside the pairwise estimates, says nothing about coverage.

One wrinkle worth knowing: `attainable` compares with `≈` rather than `==`, because `1 - 2 * p0` and `1 - 2/6` differ in the last ulp.

Rebased onto master.

andreasnoack (approved):

PR #385, opened 2026-09-02
Pin the rank test routes the suite could not tell apart, against R

yoninazarathy:
One of the standalone pieces of #376. Test-only: every pin records behaviour the package already has, against R's `wilcox.test` or `exactRankTests::wilcox.exact`, so routes the suite could not tell apart each get a value of their own. Merge in any order.

Tied one-sided p-values of both tests against `wilcox.exact` to nine digits, with the doubling-versus-far-tail convention difference recorded where two-sided values part ways; a 3 v 6 sample on which the exact and approximate Mann-Whitney interval rules pick different order statistics; the continuity-corrected one-sided approximate p-values; a 12 v 8 untied sample through the `nx > ny` swap; the automatic rule's boundaries (`n = 50/51` against R, tied 15/16 and 10/11 by type); all-equal samples returning `p = 1` on both routes. One loose tolerance is tightened to the exact enumeration value it was hiding.

### DIFF
--- a/test/mann_whitney.jl
+++ b/test/mann_whitney.jl
@@ -266,6 +266,16 @@ end
     @test all(isapprox.(confint(ta; tail=:left), (-Inf, 1.01); atol=1e-3))
     @test all(isapprox.(confint(ta; tail=:right), (-0.098, Inf); atol=1e-3))
 
+    # tied one-sided p-values against exactRankTests::wilcox.exact, which computes the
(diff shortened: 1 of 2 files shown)


andreasnoack (approved):

andreasnoack:
There is now a conflict here

andreasnoack:
There is now a new conflict here

andreasnoack (approved):

PR #386, opened 2026-09-03
Give the signed rank tests intervals that match the test, dropping zeros

### LINKED ISSUE #361: `calculate_ci` uses the exact null distribution for `Approximate*` tests, and selects the wrong order statistic from n = 73

### LINKED ISSUE #362: Zero differences are dropped for the signed-rank statistic but retained for its confidence interval

yoninazarathy:
Closes #361, closes #362. The last of the nine-way split; everything else from it is merged.

These two cannot be separated: #362 fixes zero handling *inside* `calculate_ci`, and #361 deletes `calculate_ci`, so whichever landed second would be void or a trivial conflict.

**#361.** `StatsAPI.confint` was defined identically on both signed rank types and called `signrankcdf` either way, so selecting `ApproximateSignedRankTest` did not produce an approximate interval: there was no normal-approximation signed rank interval in the package at all, and the tie correction the approximate type carries in `sigma` never reached its interval. `Exact*` now inverts the exact null distribution through the shared conservative index rule; `Approximate*` inverts the normal approximation with the continuity correction and the tie-corrected variance. On the fifteen-point sample in the tests the two now differ, `(3.3, 15.5)` against `(3.05, 15.5)`, matching R's `exact = TRUE` and `exact = FALSE`.

Deleting `calculate_ci` also makes level selection conservative rather than nearest. The old `argmin` could settle on an attainable coverage *below* nominal; taking the largest index that still attains the level reproduces R's `qsignrank` rule on all 67 sample sizes `n = 5:71` at 0.80, 0.90, 0.95 and 0.99, where the old rule agreed with R on 30, 35, 32 and 27. It is a binary search rather than a scan, so `confint(SignedRankTest(randn(400)))` goes from 164 s to under a millisecond.

**#362.** `signedrankstats` drops zeros before ranking, so the statistic and the p-value describe the non-zero differences, but the interval was built from the Walsh averages of all of them: 210 against 120 on a 20-point sample with five zeros. The p-value and the interval described different samples, and the estimate could fall outside the interval printed beside it, which a test pins on a sample where it did.

Routing through the shared rules also brings these tests the out-of-reach warnings those rules already carry for Mann-Whitney, so the named coverage is pinned for both procedures now, including the one-sided form and the boundary where `P(W <= 0)` equals `alpha/2` exactly and nothing fires.

Every warning the suite emits from these routes is asserted rather than silenced, both reflection sweeps included, which was your ask on #384.

One PR follows this: the `MAX_EXACT_CI_ESTIMATES` recalibration and the 0.12.0 bump.

### DIFF
--- a/docs/src/nonparametric.md
+++ b/docs/src/nonparametric.md
@@ -83,10 +83,18 @@ and `hodgeslehmann` returns its median.
 hodgeslehmann
 ```
 
-The Hodges-Lehmann estimate is the point estimate the interval brackets. It is generally
-not the sample median: exact symmetry makes the two agree, but they can also agree by
-coincidence, as on `[0, 2, 2, 7]`, where both are `2`. Agreement is therefore no evidence
-of symmetry.
+The Hodges-Lehmann estimate is the point estimate the interval brackets, and is what all
+four types report as their parameter of interest. It is generally not the sample median:
+exact symmetry makes the two agree, but they can also agree by coincidence, as on
+`[1, 3, 3, 8]`, where both are `3`. Agreement is therefore no evidence of symmetry.
+
+Which order statistics `confint` returns depends on the test: the `Exact*` types invert
+the exact null distribution, conservatively, and the `Approximate*` types invert the
+normal approximation with a continuity correction and a tie-corrected variance. On untied
+data the `Exact*` interval agrees with R's `wilcox.test` at `exact = TRUE` digit for digit,
+and under ties R declines an exact interval even when asked for one; the
+`Approximate*` interval agrees with `exact = FALSE` to within one order statistic, since
+R solves for its endpoints numerically rather than reading them off the pairwise estimates.
 
 A one-sided interval keeps the endpoint that inverts the test of the same name:
 `tail = :left`, whose alternative is location below the null, gives an upper bound, and
--- a/src/wilcoxon.jl
+++ b/src/wilcoxon.jl
@@ -196,15 +196,11 @@ function StatsAPI.pvalue(x::ExactSignedRankTest; tail=:both)
     end
 end
 
-# The Walsh averages that the point estimate is read off.
+# The Walsh averages that the interval and the point estimate are read off.
 #
 # `signedrankstats` drops zero differences before ranking, so the statistic and the
-# p-value describe the non-zero differences, and the estimate must describe the same
-# sample. R's `wilcox.test` drops them likewise.
-#
-# `confint` does not yet use this: it still forms its Walsh averages from every
-# difference, zeros included, so the interval and the estimate can describe
-# different samples. See #362.
+# p-value describe the non-zero differences; the interval and the estimate must
+# describe the same sample. R's `wilcox.test` drops them likewise.
 function signedrank_pairwise_estimates(vals::AbstractVector{<:Real})
     nonzero = filter(!iszero, vals)
     # every difference is zero: nothing to drop, and the pairwise estimates degenerate to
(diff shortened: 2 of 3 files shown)


andreasnoack (approved):

yoninazarathy:
Rebased onto master, and it carries one extra commit that is not about the rank tests: `docs/Project.toml`'s compat bound follows the 0.12.0 bump from `96ec50b`. Without it the Documentation job fails at `Pkg.develop` with `empty intersection between HypothesisTests@0.12.0 and project compatibility 0.11`, which is why **master is currently red on its own run** and every open PR inherits it. Happy to split that commit out if you would rather have it separately; it is here because this PR cannot be green without it. Verified locally: full suite green, and the doc build completes clean.

One thing worth flagging before you register anything: master is stamped 0.12.0 but #361 and #362 are still unfixed there, so a release cut now would ship the approximate signed rank interval still returning the exact one, and zeros still in the interval. This PR is what closes that gap.

yoninazarathy:
Ignore my comment above, it crossed with your merge and is out of date: you had already fixed the docs compat in #387 and rebased this branch onto it, so the extra commit it describes was never needed. I briefly force-pushed my own version over your rebase; the branch is back to `434d482`, the commit you actually merged, and `master` is unaffected.

Thanks for #387, and for taking the version bump. What remains from the original plan is just the `MAX_EXACT_CI_ESTIMATES` recalibration, which I will put up next.

Issue #43, opened 2016-01-04
Integer overflow in PowerDivergenceTest()


annalam:
Running a power divergence test on a matrix with large-ish numbers (in the thousands) currently results in integer overflow on line 223:

```
V = Float64[ colsums[j]*rowsums[i]*(n-rowsums[i])*(n-colsums[j])/n^3 for i in 1:nrows, j in 1:ncols]
```

This ultimately causes sqrt(V) to fail on line 253.

The integer overflow is resolved by changing line 223 to:

```
V = Float64[Float64(colsums[j])*rowsums[i]*(n-rowsums[i])*(n-colsums[j])/Float64(n)^3 for i in 1:nrows, j in 1:ncols]
```

PR #392, opened 2026-09-09
Add 32-bit Linux CI job and fix 32-bit failures

### LINKED ISSUE #244: KSampleADTest fails on 32-bit with longer data arrays
On 32-bit systems (tested on 32-bit Linux), `KSampleADTest` starts to fail when the data arrays grow larger  (not really all that large):

```
julia> using HypothesisTests

julia> pvalue(KSampleADTest(randn(100), randn(100)))
0.4684450603589013

julia> pvalue(KSampleADTest(randn(1000), randn(1000)))
ERROR: DomainError with -0.008755839260454446:
sqrt will only return a complex result if called with a complex argument. Try sqrt(Complex(x)).
[...]

julia> pvalue(KSampleADTest(randn(50000), randn(50000)))
NaN

julia> versioninfo()
Julia Version 1.6.2
Commit 1b93d53fc4 (2021-07-14 15:36 UTC)
Platform Info:
  OS: Linux (i686-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
  WORD_SIZE: 32
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, skylake)
```

On 64-bit, the above runs fine.

andreasnoack:
Adds one matrix entry that runs the latest stable Julia 1.x as an i686 build on the x64 `ubuntu-latest` runner, via `arch: x86` in `julia-actions/setup-julia`. GitHub has no native 32-bit runners, but 32-bit Julia runs fine on the 64-bit Linux and Windows images, and this is the cheapest way to catch `Int32` sizing and overflow bugs.

The existing nine jobs are unchanged: they get `arch: default`, which is the action's own default and resolves to the runner's native architecture. The job name now includes the arch so the new entry is distinguishable in the checks list.

The new job caught several 32-bit bugs, fixed here:

- **`KSampleADTest` (#244).** In `a2_ksample` the variance polynomial evaluates `N^3` in `Int32` before the `Float64` coefficient enters, so for a pooled sample above 1290 it wraps negative and `sqrt(σ²)` throws a `DomainError`. The polynomial is now evaluated in Horner form so every product starts from a `Float64` coefficient.
- **`PowerDivergenceTest` dispatch.** Struct fields and the `bootstrap_iters` keyword were annotated `Int64`, so on 32-bit the `Int32` default and the `Int32` values from `length` and `count` failed to dispatch. They are now `Int`.
- **`PowerDivergenceTest` overflow.** Sison-Glaz raised the integer cell count to the fourth power before the `Float64` factor entered, wrapping for counts above 215; the count is now converted to `Float64` first. The constructor formed the integer outer product `rowsums * colsums` before dividing by `n`, wrapping for the #43 table; it is now the broadcast `rowsums .* (colsums ./ n)`.

Bumps the version to 0.12.2.

Fixes #244

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/power_divergence.jl
+++ b/src/power_divergence.jl
@@ -10,9 +10,9 @@ struct PowerDivergenceTest <: HypothesisTest
     lambda::Float64
     theta0::Vector{Float64}
     stat::Float64
-    df::Int64
-    observed::Matrix{Int64}
-    n::Int64
+    df::Int
+    observed::Matrix{Int}
+    n::Int
     thetahat::Vector{Float64}
 
     expected::Matrix{Float64}
@@ -67,7 +67,7 @@ one of the following methods. Possible values for `method` are:
 """
 function StatsAPI.confint(x::PowerDivergenceTest; level::Float64=0.95,
                           tail::Symbol=:both, method::Symbol=:auto, correct::Bool=true,
-                          bootstrap_iters::Int64=10000, GC::Bool=true)
+                          bootstrap_iters::Int=10000, GC::Bool=true)
     check_level(level)
 
     m  = length(x.thetahat)
@@ -99,7 +99,7 @@ function StatsAPI.confint(x::PowerDivergenceTest; level::Float64=0.95,
 end
 
 # Bootstrap
-function ci_bootstrap(x::PowerDivergenceTest,alpha::Float64, iters::Int64)
+function ci_bootstrap(x::PowerDivergenceTest,alpha::Float64, iters::Int)
     m = mapslices(x -> quantile(x, [alpha / 2, 1 - alpha / 2]), rand(Multinomial(x.n, convert(Vector{Float64}, x.thetahat)),iters) / x.n, dims=2)
     Tuple{Float64,Float64}[(boundproportion(m[i,1]), boundproportion(m[i,2])) for i in 1:length(x.thetahat)]
 end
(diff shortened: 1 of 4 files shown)


src/power_divergence.jl
```diff
@@ -297,7 +297,7 @@ function PowerDivergenceTest(x::AbstractMatrix{T}; lambda::U=1.0, theta0::Vector
         colsums = sum(x, dims=1)
         df = (nrows - 1) * (ncols - 1)
         thetahat = x ./ n
-        xhat = rowsums * colsums / n
+        xhat = float(rowsums) * colsums / n
```
devmotion: This `float` allocates a vector, doesn't it? I wonder, if we should instead of `rowsums` work with `z = sum(thetahat; dims=2)` (maybe also for `colsums` but it seems the formulas are not completely symmetric) as `rowsums` seems to appear only as `rowsums / n`, here and in the definition of `V` below.

andreasnoack: Claude had sneaked that one in. I'd asked it to make it `rowsums * (colsums / n)` but that broke the tests because of the show methods printing all the digits. I've updated the PR.

devmotion (approved):

PR #393, opened 2026-09-09
Print scalar test details through the compact IO context

andreasnoack:
`Base.show` for a `HypothesisTest` wraps the stream in an `IOContext` with `:compact => true`, which is why vectors such as residuals already print with six significant digits. Several `show_params` methods interpolated their scalars into the string instead, as in `"statistic: $(x.stat)"`. Interpolation formats the number before it reaches the stream, so the context is bypassed and statistics, standard errors and non-integer degrees of freedom printed with all seventeen digits, while the methods that pass the value as a `println` argument printed compactly.

This PR makes every `show_params` method pass its scalars as `println` arguments. Before and after for the power divergence test:

```
statistic:          30.070149095754687
statistic:          30.0701
```

Affected methods: Anderson-Darling (both), binomial, circular (both), Clark-West, Diebold-Mariano, F, Kolmogorov-Smirnov (all three), power divergence, t (both), Wald-Wolfowitz and z (both). Bracketed pairs of integer counts such as `[10,8]` are left as they were, since compact printing does not change integers.

The show tests and the t-test doctests are updated to the compact values. A side effect is that the show tests no longer encode the last bits of the statistic, so they stop breaking on harmless changes in floating-point evaluation order, which is what surfaced this in #392.

The full test suite and the doctests pass locally.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/src/specs/t_test.md
+++ b/docs/src/specs/t_test.md
@@ -210,9 +210,9 @@ Test summary:
 
 Details:
     number of observations:   10
-    t-statistic:              3.2836227276929635
+    t-statistic:              3.28362
     degrees of freedom:       9
-    empirical standard error: 0.1674979270186815
+    empirical standard error: 0.167498
 
 
 julia> pvalue(t)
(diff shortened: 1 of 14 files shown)


devmotion (approved):
I completely agree with this PR (see also https://docs.julialang.org/en/v1/manual/performance-tips/#Avoid-string-interpolation-for-I/O)

